PR #1, opened 2022-01-28
docs: Link to qem-dashboard from README.md

kalikiana:
Adding the link to the project on GitHub.


okurz:
So far this project is mirrored from gitlab to github with no way back. So your commit might be overwritten again. But let's try it out, shall we? :)

PR #3, opened 2022-02-22
Change log level from error to warning wheen read metadata

mimi1vx:
This isn't a fatal error and can confuse users when checking logs

https://progress.opensuse.org/issues/107227


Martchus:
`s/wheen/when/` in commit message

PR #5, opened 2022-03-09
Skip kernel job without product repo

mimi1vx:


Martchus (commented):
It would be good to state the intention somewhere in the commit message (and possibly reference a ticket or state some more context information).

PR #13, opened 2022-04-08
Add CI job and SMELT/DASHBOARD url to job settings

mimi1vx:
See https://progress.opensuse.org/issues/109512

### DIFF
--- a/openqabot/openqabot.py
+++ b/openqabot/openqabot.py
@@ -1,5 +1,6 @@
 from argparse import Namespace
 from logging import getLogger
+from os import environ
 
 import requests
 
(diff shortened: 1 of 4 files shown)


okurz (approved):
Very nice

Martchus (approved):
Looks generally good but we should wait until https://github.com/os-autoinst/openQA/pull/4598 is deployed. This also shows that you haven't fully tested it, so how did you test it?

mimi1vx:
@Martchus local run and check results, not tested against openQA .. what openQA do with settings isn't something which bot +- cares

Martchus:
@mimi1vx Ok, but this means that no products will be scheduled until the next OSD deployment (which will contain my PR).

PR #14, opened 2022-04-08
Remove obsollete GitlabCI config

mimi1vx:


baierjan (approved):
Yes, this is definitely useless here.

PR #15, opened 2022-04-08
Add CI job URL to scheduled product settings

Martchus:
See https://progress.opensuse.org/issues/109512

### DIFF
--- a/openqabot/openqa.py
+++ b/openqabot/openqa.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from pprint import pformat
 from urllib.parse import ParseResult
 from openqa_client.client import OpenQA_Client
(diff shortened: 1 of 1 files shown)


Martchus:
There was no good starting point for tests so I've just tested it manually so far by invoking `CI_JOB_URL=https://martchus.no-ip.biz ./test.py`:

```
#!/usr/bin/python3

import openqabot.openqa
import urllib

i = openqabot.openqa.openQAInterface(instance=urllib.parse.urlparse('http://localhost:9526'))
i.post_job({"ISO": "test", "DISTRI": "foo", "VERSION": "42", "FLAVOR": "any", "ARCH": "x86_64"})
```

openqabot/openqa.py
```diff
@@ -18,6 +19,8 @@ def __bool__(self) -> bool:
         return self.url.netloc == "openqa.suse.de"
 
     def post_job(self, settings) -> None:
+        if 'CI_JOB_URL' in os.environ:
+            settings['__TRIGGERED_BY'] = os.environ['CI_JOB_URL']
```
mimi1vx, line 23: this will cause exception outside gitlab CI ,  but nice location to add

Martchus: Why will it cause an exception outside of the CI? As I've just mentioned, I've tested it using a simple script locally. I've also invoked it without having the env variable set and it worked (although of course then there's no URL added).

Martchus:
Closing in favor of https://github.com/openSUSE/qem-bot/pull/13 which seems more comprehensive.

openqabot/openqa.py
mimi1vx: `os.environ['CI_JOB_URL']` without defined `CI_JOB_URL` throws `KeyError` exception as is in reality python dict.

mimi1vx: ah, I,m blind ... you have it in condition :D I was focused on line 23 :(

PR #16, opened 2022-04-08
Fix aggregates variables.

mimi1vx:
I was too fast and used same vars as in Incidets, which is incorrect
because Aggregate shedule jobs with multiple incidents contained

So we need array of URL's


Martchus:
This PR should fix displaying it within openQA: https://github.com/os-autoinst/openQA/pull/4599

PR #17, opened 2022-04-11
Move SMELT and Dashboard info before PC handling

mimi1vx:
https://progress.opensuse.org/issues/109512#note-8

### DIFF
--- a/openqabot/types/incidents.py
+++ b/openqabot/types/incidents.py
@@ -239,6 +239,13 @@ class Incidents(BaseConf):
                     if "params_expand" in data:
                         full_post["openqa"].update(data["params_expand"])
 
+                    full_post["openqa"][
+                        "__SMELT_INCIDENT_URL"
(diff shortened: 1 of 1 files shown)


Martchus (commented):
I'd remove the dots at the end of the commit message. Otherwise I suppose this should work (by moving the code before the "public cloud continues". There are actually more cases in which the loop continues before but I suppose they're not relevant here.

Martchus:
Oh, and it would also great if you could reference the related ticket, or in this case even the related comment in the commit message: https://progress.opensuse.org/issues/109512#note-8

okurz (approved):

Martchus (approved):

okurz:
I would have expected that a recently scheduled incident product like https://openqa.suse.de/admin/productlog?id=895372 would have the necessary URLs already but I don't see them there. Am I too early?

PR #19, opened 2022-04-19
Improve logging of bot-ng

mimi1vx:
https://progress.opensuse.org/issues/109488

### DIFF
--- a/openqabot/loader/config.py
+++ b/openqabot/loader/config.py
@@ -37,7 +37,7 @@ def load_metadata(
             continue
 
         if "product" not in data:
-            logger.warning("Missing product in %s" % p)
+            logger.debug("Skipping invalid config %s" % p)
             continue
 
         if settings:
@@ -66,16 +66,16 @@ def read_products(path: Path) -> List[Data]:
         data = loader.load(p)
 
         if not data:
-            logger.warning("something wrong with %s" % str(p))
+            logger.info("Skipping invalid config %s - empty config" % str(p))
             continue
         if not isinstance(data, dict):
-            logger.warning("something wrong with %s" % str(p))
+            logger.info("Skipping invalid config %s - invalid format" % str(p))
             continue
 
         try:
             flavor = data["aggregate"]["FLAVOR"]
         except KeyError:
-            logger.info("file %s dont have aggregate" % str(p))
+            logger.info("Config %s does not have aggregate" % str(p))
             continue
 
         try:
--- a/openqabot/openqa.py
+++ b/openqabot/openqa.py
@@ -4,6 +4,7 @@ import logging
 from pprint import pformat
 from urllib.parse import ParseResult
 from openqa_client.client import OpenQA_Client
+from openqa_client.exceptions import RequestError
 
 from .types import Data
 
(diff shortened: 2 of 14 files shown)


openqabot/loader/config.py
```diff
@@ -64,16 +64,16 @@ def read_products(path: Path) -> List[Data]:
         data = loader.load(p)
 
         if not data:
-            logger.warning("something wrong with %s" % str(p))
+            logger.debug("Skiping invalid config %s" % str(p))
```
okurz, line 67: I suggest to not simply reuse the same text as in before but use specific, differing log messages

openqabot/types/incidents.py
```diff
@@ -135,10 +137,11 @@ def __call__(
                             ArchVer(arch, self.settings["VERSION"])
                         ]
                     except KeyError:
-                        logger.debug(
-                            "Incident %s does not have %s arch in %s"
-                            % (inc.id, arch, self.settings["VERSION"])
-                        )
+                        if REALLY_LONG_LOG:
```
okurz, line 140: What is this about?

okurz (changes requested):

okurz:
@Martchus

openqabot/loader/config.py
kalikiana:
```suggestion
            logger.debug("Skipping invalid config %s" % str(p))
```

Martchus (commented):
I only haven't added a review because I don't have anything to say in addition to @okurz 's review but all his comments should be addressed.

openqabot/types/incidents.py
mimi1vx: trace, usually this info has no value , only when is needed detailed info

kalikiana: That's how I would describe *debug* 🤔

okurz: I suggest to either just use debug or introduce a trace level like in https://www.programcreek.com/python/example/56843/logging.TRACE

openqabot/openqa.py
```diff
@@ -20,15 +21,24 @@ def __bool__(self) -> bool:
...
+        except RequestError as e:
+            logger.error("openQA returned %s" % e[-1])
+            logger.error("Post failed with {}".format(pformat(settings)))
         except Exception as e:
             logger.exception(e)
             logger.error("Post failed with {}".format(pformat(settings)))
 
     def get_jobs(self, data: Data):
-        logger.debug("getting openqa for %s" % pformat(data))
+        logger.info("getting openqa for %s" % pformat(data))
```
baierjan, line 41: Cosmetic suggestion (all other messages beings with a capital letter)?
```suggestion
        logger.info("Getting openQA for %s" % pformat(data))
```

baierjan (approved):

openqabot/openqa.py
```diff
@@ -20,15 +21,24 @@ def __bool__(self) -> bool:
...
+        except RequestError as e:
+            logger.error("openQA returned %s" % e[-1])
+            logger.error("Post failed with {}".format(pformat(settings)))
         except Exception as e:
             logger.exception(e)
             logger.error("Post failed with {}".format(pformat(settings)))
 
     def get_jobs(self, data: Data):
-        logger.debug("getting openqa for %s" % pformat(data))
+        logger.info("Getting openqa for %s" % pformat(data))
```
Martchus, line 41:
```suggestion
        logger.info("Getting openQA for %s" % pformat(data))
```
Same counts for other occurrences. About this specific one: Should it actually be "openQA job"?
```suggestion
        logger.info("Getting openQA job for %s" % pformat(data))
```

Martchus (approved):

PR #21, opened 2022-04-21
Add tests+CI

okurz:
* Add github CI instructions
* Add simple, initial test with Makefile

After
* #20

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,15 @@
+---
+name: ci
+# yamllint disable-line rule:truthy
+on: [push, pull_request]
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v3
+        with:
+          python-version: '3.6'
+          cache: 'pip'
+      - run: pip install pytest
+      - run: make test
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,6 @@
+.PHONY: all
+all:
+
+.PHONY: test
+test:
+	py.test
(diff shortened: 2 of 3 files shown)


kraih (approved):
It's a start.

Makefile
```diff
@@ -0,0 +1,6 @@
+.PHONY: all
+all:
+
+.PHONY: test
+test:
+	py.test
```
baierjan, line 6: This is a typo or a some Makefile auto-magic?

```suggestion
	pytest
```

baierjan:
Looking at https://github.com/okurz/qem-bot/runs/6111349548?check_suite_focus=true#step:4:25, it would definitely need some other modules. We might try to find out the minimal set or extract them from the current image

Makefile
baierjan: Ah, that's pip magic.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,15 @@
+---
+name: ci
+# yamllint disable-line rule:truthy
+on: [push, pull_request]
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
mimi1vx, line 9: What about using v3 plus setup-python action ?

mimi1vx (changes requested):

.github/workflows/ci.yml
baierjan: Agree. Something like this would probably work? (Not sure about the requirements though.)
```
steps:
- uses: actions/checkout@v3
- uses: actions/setup-python@v3
  with:
    python-version: '3.6'
    cache: 'pip'
- run: pip install -r requirements.txt
```

okurz: Haven't used `pip install -r requirements.txt` yet but the other suggestions.

Martchus (approved):

mimi1vx:
superseded by #22

PR #25, opened 2022-04-26
Make pytest work without adding path

perlpunk:


kalikiana (approved):

Martchus (approved):
Ok, although I suppose the Makefile is quite useless when one can just invoke `pytest`. (Will be fixed by https://github.com/openSUSE/qem-bot/pull/23/commits/8bf810c63f9a9e43b71da18f555c6fd75ed02b38.)

okurz:
> Ok, although I suppose the Makefile is quite useless when one can just invoke `pytest`. (Will be fixed by [8bf810c](https://github.com/openSUSE/qem-bot/commit/8bf810c63f9a9e43b71da18f555c6fd75ed02b38).)

It is providing the simple convenient entry path because many people know how to call "make" or "make test" regardless of the programming language specific ecosystem. Also we can easily extend the Makefile but the instructions stay the same :)

PR #26, opened 2022-04-29
Fix status badge

okurz:
Sorry about that. I previously hand-crafted the badge
URL when instead I should have used
the "Create status badge" button within github actions.


perlpunk (approved):

okurz:
It still looks a bit weird in the preview, let's see.

PR #31, opened 2022-05-30
Print only single line for job post in result sync

mimi1vx:
https://progress.opensuse.org/issues/111407

### DIFF
--- a/openqabot/aggrsync.py
+++ b/openqabot/aggrsync.py
@@ -75,7 +75,11 @@ class AggregateResultsSync:
                 results.append(r)
 
         for r in results:
-            logger.info("Posting aggregate job result to dashboard: %s" % pformat(r))
+            logger.info(
+                "Posting results of aggregate job %s with status %"
+                % (r["job_id"], r["status"])
+            )
+            logger.debug("Full aggregate job result to dashboard: %s" % pformat(r))
 
             if not self.dry and self.client:
                 post_job(self.token, r)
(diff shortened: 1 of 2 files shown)


openqabot/aggrsync.py
```diff
@@ -75,7 +75,10 @@ def __call__(self) -> int:
                 results.append(r)
 
         for r in results:
-            logger.info("Posting aggregate job result to dashboard: %s" % pformat(r))
+            logger.info(
+                "Posting results of %s with status %" % (r["job_id"], r["status"])
```
okurz, line 79: Now the string is the same as in incsyncres. How about

```suggestion
                "Posting aggregate results of %s with status %" % (r["job_id"], r["status"])
```

openqabot/incsyncres.py
```diff
@@ -68,7 +68,11 @@ def __call__(self) -> int:
                 results.append(r)
 
         for r in results:
-            logger.info("Posting job results to dashboard: %s" % pformat(r))
+            logger.info(
+                "Posting results of incident job %s with status %"
+                % (r["job_id"], r["status"])
+            )
+            logger.debug("Full post data: %s" % pformat(r))
```
baierjan, line 75: Maybe a small change to make it more similar to aggregates messages?

```suggestion
            logger.debug("Full incident job result to dashboard: %s" % pformat(r))
```

baierjan (approved):

openqabot/incsyncres.py
mimi1vx: this info is in logger.info above ... so it will be only duplicate

PR #32, opened 2022-05-30
Fix missing format specification in result sync

mimi1vx:


baierjan (approved):
Nice typo, maybe it is time for a new keyboard? 😄

mimi1vx:
> Nice typo, maybe it is time for a new keyboard? 😄

keyboard is good ( it will outlive me, its cherry brown) , my ADHD is to blame ...

PR #34, opened 2022-05-31
Refactor result sync code

mimi1vx:
https://progress.opensuse.org/issues/111506

### DIFF
--- a/openqabot/openqa.py
+++ b/openqabot/openqa.py
@@ -1,8 +1,10 @@
 # Copyright SUSE LLC
 # SPDX-License-Identifier: MIT
+from functools import lru_cache
 import logging
 from pprint import pformat
 from urllib.parse import ParseResult
+
 from openqa_client.client import OpenQA_Client
 from openqa_client.exceptions import RequestError
 
@@ -56,3 +58,16 @@ class openQAInterface:
             logger.exception(e)
             raise e
         return ret
+
+    @lru_cache(maxsize=256)
+    def is_devel_group(self, groupid: int) -> bool:
+        ret = None
+
+        try:
+            ret = self.openqa.openqa_request("GET", f"job_groups/{groupid}")
+        except Exception as e:
+            logger.exception(e)
+            raise e
+
+        # return True as safe option if ret = None
+        return ret[0]["name"] == "Development" if ret else True
--- a/openqabot/syncres.py
+++ b/openqabot/syncres.py
@@ -1,0 +1,81 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: MIT
+
+from argparse import Namespace
+from logging import getLogger
+from pprint import pformat
+from typing import Dict
+
+from .loader.qem import post_job
+from .openqa import openQAInterface
+from .types import Data
+from .utils import normalize_results
+
+logger = getLogger("bot.syncres")
+
+
+class SyncRes:
+    operation = "null"
+
+    def __init__(self, args: Namespace) -> None:
+        self.dry: bool = args.dry
+        self.token: Dict[str, str] = {"Authorization": f"Token {args.token}"}
+        self.client = openQAInterface(args.openqa_instance)
+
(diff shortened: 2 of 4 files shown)


openqabot/openqa.py
```diff
@@ -56,3 +58,16 @@ def get_jobs(self, data: Data):
...
             raise e
         return ret
+
+    @lru_cache(maxsize=256)
+    def is_devel_group(self, groupid: int) -> bool:
+        ret = None
+
+        try:
+            ret = self.openqa.openqa_request("GET", f"job_groups/{groupid}")
+        # TODO: correct handling
```
okurz, line 68: As you stated, TODO :)

openqabot/syncres.py
```diff
@@ -70,6 +70,10 @@ def filter_jobs(data) -> bool:
             )
             return False
 
+        if self.client.is_devel_group(data["group_id"]):
+            logger.info("Devel job %s in group %s" % (data["id"], data["group"]))
```
okurz, line 74: Better make that debug

openqabot/syncres.py
```diff
@@ -56,6 +55,7 @@ def filter_jobs(data) -> bool:
             return False
 
         if "Devel" in data["group"] or "Test" in data["group"]:
+
```
okurz, line 58: I like it better without :)

openqabot/openqa.py
```diff
@@ -56,3 +58,16 @@ def get_jobs(self, data: Data):
...
+        except Exception as e:
+            logger.exception(e)
+            raise e
+
+        return ret[0]["parent_id"] == 9  # magic number of parent "Development" group
```
okurz, line 73: We can do better :)

okurz (changes requested):

openqabot/syncres.py
mimi1vx: this isn't bug , but one of many correct states

openqabot/openqa.py
mimi1vx: How ? `ret[0]["name"] == ""Development"` ... same thing

openqabot/syncres.py
okurz: Yes, but "info" should be used for something that is noteworthy and shouldn't appear that often. I assume this line would be logged quite often.

I would prefer if you could not call a conversation "resolved" in github when we haven't found a resolution yet.

openqabot/openqa.py
okurz: Well, using "Development" as part of the name is certainly a convention. But we shouldn't hardcode a "magic number" in the source code. I think reading out the parent job group names is feasible but if you would prefer to prevent an API call every time then I suggest to move this ID to a configuration variable that has "9" as default

openqabot/openqa.py
```diff
@@ -56,3 +58,19 @@ def get_jobs(self, data: Data):
...
+            ret = self.openqa.openqa_request("GET", f"job_groups/{groupid}")
+        except Exception as e:
+            logger.exception(e)
+            raise e
+
+        if ret:
+            return ret[0]["name"] == "Development"
+
+        # return True as safe option if ret = None
+        return True
```
okurz, line 76:
```suggestion
        return ret[0]["name"] == "Development" if ret else True
```

openqabot/syncres.py
okurz: oh, wait. You are just moving the code. So, right, keep as is for now. We can still change this in another change but this is unrelated to a refactoring.

openqabot/openqa.py
okurz: Well, in perl it would be

```perl
return defined ret ? ret[0]{name} eq 'Development' : 1;
```

right? ;)

okurz (approved):

kalikiana (changes requested):
I like the deduplication. However please add at least some test coverage.

PR #36, opened 2022-06-01
Correctly work with 404 error from OBS in approver

mimi1vx:


openqabot/approver.py
```diff
@@ -123,11 +123,17 @@ def osc_approve(inc: IncReq) -> bool:
             )
         except HTTPError as e:
             if e.code == 403:
-                logger.debug(
+                logger.error(
```
okurz, line 126: If qem-bot already takes the action of ignoring, why log as error?

okurz (changes requested):

perlpunk (commented):
I think I would find it helpful to also log the url.

mimi1vx:
> I think I would find it helpful to also log the url.

URL will be pretty awful (obs api) , what will be helpful is Request number

openqabot/approver.py
```diff
@@ -123,13 +123,22 @@ def osc_approve(inc: IncReq) -> bool:
...
         except HTTPError as e:
             if e.code == 403:
-                logger.debug(
+                logger.info(
                     "Received '%s'. Request likely already approved, ignoring"
                     % e.reason
                 )
                 return True
+            elif e.code == 404:
+                logger.error(
```
okurz, line 132: still here I would use `logger.debug` (or `logger.info`) as there is already the action mentioned that is conducted when the situation is hit. We should ask ourselves what a reader of an error log message should do. If it's nothing that needs to be done then we also don't need to log with that high level

okurz (changes requested):

okurz (approved):

PR #37, opened 2022-06-01
Stop filtering groups by special keywords

kalikiana:
We should rely on the development parent group.

Related: https://progress.opensuse.org/issues/111506

### DIFF
--- a/openqabot/syncres.py
+++ b/openqabot/syncres.py
@@ -54,14 +54,6 @@ class SyncRes:
             logger.info("Clone job %s" % data["clone_id"])
             return False
 
-        if "Devel" in data["group"] or "Test" in data["group"]:
-            logger.info("Devel job %s in group %s" % (data["id"], data["group"]))
-            return False
-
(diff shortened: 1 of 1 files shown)


mimi1vx:
on parent group filtering needs call to OBS so its expensive :(

okurz (changes requested):
We should not *only* rely on the parent group. We should still look into the group names.

kalikiana:
> We should not _only_ rely on the parent group. We should still look into the group names.

I guess I'm not clear on what [the point of the ticket](https://progress.opensuse.org/issues/111506#note-1) is. Maybe we should instead document that groups with _Timo_, _Test_ or _Devel_ get skipped. But then, we go the expensive parent group route at all?

okurz:
> But then why go the expensive parent group route at all?

Because some job groups might be defined as an ignorable group by being a member of a *parent* Development, some are defined by the job group name having "Devel" included itself.

PR #40, opened 2022-06-09
Fix development group check

mimi1vx:
Add performace optimization when fetching result from openqa
Add test for IncSyncRes

### DIFF
--- a/openqabot/openqa.py
+++ b/openqabot/openqa.py
@@ -8,7 +8,7 @@ from urllib.parse import ParseResult
 from openqa_client.client import OpenQA_Client
 from openqa_client.exceptions import RequestError
 
-from . import OPENQA_URL
+from . import DEVELOPMENT_PARENT_GROUP_ID, OPENQA_URL
 from .types import Data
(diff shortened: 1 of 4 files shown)


openqabot/openqa.py
```diff
@@ -12,6 +12,8 @@
 
 logger = logging.getLogger("bot.openqa")
 
+_DEVELOPMENT_PARENT_GROUP_ID = 9
```
okurz, line 15: I suggest to move that to __init__.py where we have other configuration settings already

okurz (changes requested):

okurz (changes requested):
@mimi1vx you asked for review but seems you missed the actual suggested changes

mimi1vx:
> @mimi1vx you asked for review but seems you missed the actual suggested changes

oh, .... I forget push

okurz (dismissed):
Code looks good but please still fix the typo in the git commit message

baierjan (approved):

PR #41, opened 2022-06-10
Reduce verbose line "Post results …" to debug level

okurz:
A "sync aggregates" job like
https://gitlab.suse.de/qa-maintenance/bot-ng/-/jobs/1009892
has multiple thousands of lines with content
"Post results of aggregate job …" so I consider the "debug" log level
more fitting.


Martchus:
The commit itself looks good but you've also pushed a WIP commit.

okurz:
oops, fixed

PR #44, opened 2022-06-14
Handle correctly new product 'openSUSE-Leap-Micro'

mimi1vx:
This product has same incident structure as 'openSUSE-SLE' incidents

New nested generator is more universal --> in future changes affects
only loader/qem.py code

https://progress.opensuse.org/issues/112430

### DIFF
--- a/openqabot/loader/repohash.py
+++ b/openqabot/loader/repohash.py
@@ -24,9 +24,8 @@ def get_max_revision(
     url_base = f"http://download.suse.de/ibs/{project.replace(':',':/')}"
 
     for repo in repos:
-
-        # don't use openSUSE-SLE
-        if repo[0] == "openSUSE-SLE":
+        # openSUSE and SLE incidents have different handling of architecture
+        if repo[0].startswith("openSUSE"):
             url = f"{url_base}/SUSE_Updates_{repo[0]}_{repo[1]}/repodata/repomd.xml"
         else:
             url = f"{url_base}/SUSE_Updates_{repo[0]}_{repo[1]}_{arch}/repodata/repomd.xml"
--- a/openqabot/types/incident.py
+++ b/openqabot/types/incident.py
@@ -23,9 +23,13 @@ class Incident:
         self.channels = [
             Repos(p, v, a)
             for p, v, a in (
-                r.split(":")[2:]
-                for r in incident["channels"]
-                if r.startswith("SUSE:Updates") and "openSUSE-SLE" not in r
+                val
+                for val in (
+                    r.split(":")[2:]
+                    for r in incident["channels"]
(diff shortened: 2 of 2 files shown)


openqabot/loader/repohash.py
```diff
@@ -25,8 +25,7 @@ def get_max_revision(
 
     for repo in repos:
 
-        # don't use openSUSE-SLE
-        if repo[0] == "openSUSE-SLE":
+        if repo[0] in ("openSUSE-SLE", "openSUSE-Leap-Micro"):
```
okurz, line 28: well, was not clear in before, still not clear why these should be handled differently, better to use a function with good name describing that

openqabot/types/incident.py
```diff
@@ -35,12 +39,14 @@ def __init__(self, incident):
...
-                    for i in incident["channels"]
-                    if i.startswith("SUSE:Updates") and "openSUSE-SLE" in i
+                val
+                for val in (
+                    r.split(":")[2:]
+                    for r in (
+                        i for i in incident["channels"] if i.startswith("SUSE:Updates")
+                    )
                 )
+                if len(val) == 2
```
okurz, line 49: a prime example of a magic number I would way. This should be avoided. Maybe at least you can extract that into a function with a good name explaining this?

okurz (changes requested):

openqabot/types/incident.py
mimi1vx: This isnt magic number:) it simply length of tuple 'Val'

openqabot/loader/repohash.py
mimi1vx: because SLE incidents repo's have different structure than openSUSE.
And we handle only these two ... Right it can be simpler

openqabot/types/incident.py
```diff
@@ -23,9 +23,13 @@ def __init__(self, incident):
...
-                r.split(":")[2:]
-                for r in incident["channels"]
-                if r.startswith("SUSE:Updates") and "openSUSE-SLE" not in r
+                val
+                for val in (
+                    r.split(":")[2:]
+                    for r in incident["channels"]
+                    if r.startswith("SUSE:Updates")
+                )
+                if len(val) == 3
```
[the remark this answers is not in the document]
mimi1vx: ` Repos(p, v, a) for p,v,a in ....` so you excepting tuple with 3 values .. and this filter out results with different length tuples

some light text about comprehensions and generators https://djangostars.com/blog/list-comprehensions-and-generator-expressions/

openqabot/loader/repohash.py
okurz: much better, thank you

PR #46, opened 2022-06-15
Fixes in type annotations suggested by mypy

mimi1vx:


Martchus (approved):

okurz:
Interesting, didn't know about mypy, something learned :)

PR #47, opened 2022-07-01
Ensure to check proper results of all request.get calls

okurz:
https://progress.opensuse.org/issues/113087

### DIFF
--- a/openqabot/openqabot.py
+++ b/openqabot/openqabot.py
@@ -3,14 +3,15 @@
 from argparse import Namespace
 from logging import getLogger
 from os import environ
+import requests as req
 
-import requests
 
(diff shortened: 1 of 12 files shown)


Martchus (approved):

openqabot/openqabot.py
```diff
@@ -3,14 +3,15 @@
 from argparse import Namespace
 from logging import getLogger
 from os import environ
+import requests as req
```
mimi1vx, line 6: why ?

mimi1vx:
probably caused schedule error ..

openqabot/openqabot.py
okurz: To distinguish from the internal requests I added

PR #50, opened 2022-07-05
Revert obsolete "Skip approving incidents with 15-SP4 jobs"

okurz:
This reverts commit 624200907d3daa50d9bc5f4168b689336c3d940a as we now
have proper 15-SP4 tests in place.

Related progress issue: https://progress.opensuse.org/issues/112124


okurz:
I removed the approval in the branch protection so that we can rely on mergify with the label "merge-fast".
As an alternative we can consider still requiring at least one approval but let mergify provide an automatic approval, see https://docs.mergify.com/actions/review/

@Mergify refresh

PR #51, opened 2022-07-19
Explain openSUSE requirements in readme file

pdostal:
When I wanted to use the bot I missed some packages.
This is the documentation. Thanks @baierjan for help.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -116,6 +116,14 @@ openqa-cli --host https://openqa.suse.de api -X post isos ARCH=s390x BASE_TEST_I
 
 Apply caution to keep all the other parameters in place.
 
+## Requirements
+
+In openSUSE you can install the requirements by running:
+
+```
+sudo zypper in python3-beautifulsoup4 python3-openqa_client python3-requests python3-ruamel.yaml
+```
+
 ## Contribute
 
 This project lives in https://github.com/openSUSE/qem-bot


Readme.md
```diff
@@ -116,6 +116,14 @@ openqa-cli --host https://openqa.suse.de api -X post isos ARCH=s390x BASE_TEST_I
 
 Apply caution to keep all the other parameters in place.
 
+## Requirements
+
+In openSUSE you can install the requirements by running:
+
+```
+sudo zypper in python3-beautifulsoup4 python3-openqa_client python3-requests python3-ruamel.yaml
```
okurz, line 124: Thank you for the attempt to improve the documentation. But I think it's better if we do not manually spell out package dependencies in documentation because whenever dependencies changes this becomes obsolete without anyone knowing. We have the list of requirements in requirements.txt although it seems this isn't used anywhere, only requirements-dev.txt is. Maybe reference only those files?

okurz (changes requested):

Readme.md
pdostal: Hello Oliver, as we discussed yesterday the recommended (and working) way is to use system packages.

I tried to use `requirements.txt` yesterday and failed.

baierjan: I think the general problem is more like mapping the content of the requirements file (package names as wanted by pip) to the system package name (as wanted by zypper). You want to run the bot locally the "proper way", you need to guess package names. Maybe there is some space for a yet to be created tool like `zypper install --python-requirements`?

I agree with Oliver @okurz here: It is difficult to keep up with the packaging; although I understand the need to for the package names. But as I said, that's more like general (not only) python problem.

baierjan: Maybe a solution would be to provide spec file and build it in OBS?

okurz: By the way, this is how requirements.txt looks like in openqa-review in https://github.com/os-autoinst/openqa_review/blob/master/requirements.txt#L2

```
# see https://caremad.io/2013/07/setup-vs-requirement/
-e .
```

with requirements in setup.py in https://github.com/os-autoinst/openqa_review/blob/master/setup.py#L27

there is a spec file in https://build.opensuse.org/package/view_file/devel:languages:python/python-openqa_review/python-openqa_review.spec?expand=1 but the requirement list so far is also only manually maintained. Within the scope of os-autoinst we use https://github.com/os-autoinst/os-autoinst-common/blob/master/tools/update-deps to keep a generic package requirements document, a cpanfile, a Dockerfile and a spec file in sync.

mimi1vx: > I think the general problem is more like mapping the content of the requirements file (package names as wanted by pip) to the system package name (as wanted by zypper). You want to run the bot locally the "proper way", you need to guess package names. Maybe there is some space for a yet to be created tool like `zypper install --python-requirements`?
> 
> I agree with Oliver @okurz here: It is difficult to keep up with the packaging; although I understand the need to for the package names. But as I said, that's more like general (not only) python problem.

I don't understand of need for this mapping for zypper ... on Tumbleweed we have: `python3dist(openqa-client)` etc.

so `zypper in  python3dist(openqa-client)` works

okurz: Sure, but we can't just call `zypper setup.py` or on a generic requirements file :)

mimi1vx: @okurz  on other side pavel problem with requirements file was missing c/c++ devel and build stack for some deeper dependencies ( as for example osc deps needs openssl-devel)

PR #52, opened 2022-07-25
Catch also HTTPError in public cloud code

mimi1vx:
See: https://progress.opensuse.org/issues/114607

### DIFF
--- a/openqabot/pc_helper.py
+++ b/openqabot/pc_helper.py
@@ -1,10 +1,10 @@
 # Copyright SUSE LLC
 # SPDX-License-Identifier: MIT
-import bs4
+from functools import lru_cache
 import logging
 import re
-from functools import lru_cache
-from requests.exceptions import ConnectionError, Timeout
+
+import bs4
 
 from .requests import requests
 
@@ -20,11 +20,7 @@ def fetch_matching_link(url, regex):
         # So, the first link matching the regex is the most recent one
         req = requests.get(url + "/?C=M;O=A")
         text = req.text
-    except (
-        ConnectionError,
-        Timeout,
-        ValueError,
-    ) as err:
+    except BaseException as err:
         logger.error("error fetching '%s': %s" % (url, err))
         return None
     getpage_soup = bs4.BeautifulSoup(text, "html.parser")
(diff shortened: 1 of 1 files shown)


jlausuch (approved):

openqabot/pc_helper.py
```diff
@@ -20,11 +21,7 @@ def fetch_matching_link(url, regex):
         # So, the first link matching the regex is the most recent one
         req = requests.get(url + "/?C=M;O=A")
         text = req.text
-    except (
-        ConnectionError,
-        Timeout,
-        ValueError,
-    ) as err:
+    except (ConnectionError, Timeout, ValueError, HTTPError) as err:
```
asmorodskyi, line 24: I would actually go further and just catch parent error class to mute ANY errors . Enumerating errors make sense only when situation is "Ok I know that errors A, B and C may happen let's catch them but anything else happens I would like throw it further " and it is NOT the case here . Here we saying "OK whatever error happened I want to mute it " . Because some day in future we will be here again adding yet-another-error-class into the list

openqabot/pc_helper.py
```diff
@@ -62,12 +59,7 @@ def apply_publiccloud_regex(settings):
             settings["PUBLIC_CLOUD_IMAGE_REGEX"]
         )
         return settings
-    except (
-        ConnectionError,
-        Timeout,
-        ValueError,
-        re.error,
-    ) as e:
+    except (ConnectionError, Timeout, ValueError, re.error, HTTPError) as e:
```
pdostal: Is the necessary? The `get_latest_pc_image()` calls `fetch_matching_link()` where we are already checking for those exceptions.

openqabot/pc_helper.py
mimi1vx: hmm, also possibility `BaseException` should work

openqabot/pc_helper.py
```diff
@@ -133,12 +120,7 @@ def apply_publiccloud_pint_image(settings):
         if "PUBLIC_CLOUD_PINT_FIELD" in settings:
             del settings["PUBLIC_CLOUD_PINT_FIELD"]
         return settings
-    except (
-        ConnectionError,
-        Timeout,
-        ValueError,
-        re.error,
-    ) as e:
+    except (ConnectionError, Timeout, ValueError, re.error, HTTPError) as e:
```
pdostal: Can we move this to `pint_query()`?

pdostal (approved):
Thank you @mimi1vx!

I have two suggestions, but we may address them separately.

asmorodskyi (approved):
LGTM

asmorodskyi:
agree with questions from @pdostal  but they looks to me as future improvements which we can address later . this can be merged

openqabot/pc_helper.py
mimi1vx: IDK .. its code written by PubCloud team :D and really bigger changes aren't target of this PR :D

Martchus (approved):

PR #53, opened 2022-07-26
Switch kernel-azure QEMU VMs to virtio GPU

mdoucha:
HyperV (virtio) GPU is the video device that will be used by kernel-azure in production environment. Switch to it in QEMU tests.

### DIFF
--- a/openqabot/types/incidents.py
+++ b/openqabot/types/incidents.py
@@ -124,6 +124,8 @@ class Incidents(BaseConf):
 
                     if inc.azure:
                         full_post["openqa"]["AZURE"] = "1"
+                        if flavor == "Server-DVD-Incidents-Kernel":
+                            full_post["openqa"]["QEMUVGA"] = "virtio"
 
                     full_post["openqa"]["BUILD"] = f":{inc.id}:{inc.packages[0]}"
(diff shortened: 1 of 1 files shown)


Martchus (approved):
It would be nice to include a bugref in the commit message for additional context. Otherwise it looks good as far as I can tell.

okurz (commented):
I don't think these settings should go into the source code of qem-bot. Can you not put those into medium settings?

mdoucha:
We use the same medium for kernel-azure, kernel-default and kernel-livepatch. Everything except kernel-azrue should use the default Bochs GPU instead. Ideally, these settings (including `AZURE=1` and `KGRAFT=1` should go into the bot config YAML but it seems that the bot doesn't support such complex logic.

okurz:
I think a "machine" in openQA would work better for such setting. How about defining this in a new machine and use that in job templates?

mdoucha:
> I think a "machine" in openQA would work better for such setting. How about defining this in a new machine and use that in job templates?

I don't see how to change machine for kernel-azure tests without either introducing a completely new flavor or breaking aarch64 kernel-default tests.

PR #57, opened 2022-07-29
Add status_forcelist to Retry + simple test

mimi1vx:
urlib3 by default don't force retry (not even on RETRY_AFTER_STATUS_CODES)
so we need explicitly declare on which status codes we need retry.

Plus we must bump minimal version of responses to 0.21.0 (only from this
has proper Retry mechanic support), efectively disabling default Leap python packages.

### DIFF
--- a/openqabot/utils.py
+++ b/openqabot/utils.py
@@ -2,7 +2,6 @@
 # SPDX-License-Identifier: MIT
 from copy import deepcopy
 
-import requests
 from requests import Session
 from requests.adapters import HTTPAdapter
 from urllib3.util.retry import Retry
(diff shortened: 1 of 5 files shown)


Martchus (approved):

openqabot/utils.py
```diff
@@ -51,8 +50,14 @@ def normalize_results(result: str) -> str:
 
 def __retry(retries: Optional[int]) -> Session:
 
-    adapter = HTTPAdapter(max_retries=Retry(retries, backoff_factor=5))
-    http = requests.Session()
+    adapter = HTTPAdapter(
+        max_retries=Retry(
+            retries,
+            backoff_factor=5,
+            status_forcelist=frozenset({404, 403, 413, 429, 503}),
```
okurz: I found this while debugging why my local `qem-bot` call takes ages to return with a useful result which turned out to be a missing token. Why should qem-bot retry on 403?

PR #66, opened 2022-08-23
Fix setting the arch when using TEST_ISSUES_ARCH

grisu48:
Don't set the `ARCH` setting to the same as the `TEST_ISSUES_ARCH`.

* Follow-up of https://github.com/openSUSE/qem-bot/pull/65
* Related ticket: https://progress.opensuse.org/issues/115652

### DIFF
--- a/openqabot/types/aggregate.py
+++ b/openqabot/types/aggregate.py
@@ -84,8 +84,7 @@ class Aggregate(BaseConf):
             test_repos = defaultdict(list)
 
             # Temporary workaround for applying the correct architecture on jobs, which use a helper VM
-            if "TEST_ISSUES_ARCH" in self.settings:
-                arch = self.settings["TEST_ISSUES_ARCH"]
+            issues_arch = self.settings.get("TEST_ISSUES_ARCH", arch)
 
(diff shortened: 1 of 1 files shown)


grisu48:
A local verification run shows that `arch` is set to `x86_64` while the `_TEST_REPOS` is using `aarch64` now: http://pasta.qam.suse.de/wELbwzuV (trimmed to relevant parts)

openqabot/types/aggregate.py
```diff
@@ -84,16 +84,20 @@ def __call__(
             test_repos = defaultdict(list)
 
             # Temporary workaround for applying the correct architecture on jobs, which use a helper VM
+            issues_arch = arch
```
mimi1vx, line 87: maybe better:

`issues_arch' = self.settings.get("TEST_ISSUES_ARCH", arch)`

mimi1vx (changes requested):
I would

openqabot/types/aggregate.py
grisu48: Amended, VR is in progress.

mimi1vx (approved):

PR #67, opened 2022-08-25
Add simple schema validation for SMELT responses

baierjan:
Use simple JSON schema to validate responses from SMELT. For now, only
the attributes used in iterations are required. Any error during
validation will be explained.

Reference: https://progress.opensuse.org/issues/115733

### DIFF
--- a/openqabot/loader/smelt.py
+++ b/openqabot/loader/smelt.py
@@ -3,6 +3,7 @@
 import concurrent.futures as CT
 from logging import getLogger
 from typing import Any, List, Set
+from jsonschema import validate, ValidationError
 
 import urllib3
 import urllib3.exceptions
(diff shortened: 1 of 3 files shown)


perlpunk (commented):
Maybe it would also be a good idea to do:
```
res.raise_for_status()
```
and only after that try to decode the JSON and read from the data.

Then we automatically get the HTTP status and URL in the exception.

baierjan:
> Maybe it would also be a good idea to do:
> 
> ```
> res.raise_for_status()
> ```
> 
> and only after that try to decode the JSON and read from the data.
> 
> Then we automatically get the HTTP status and URL in the exception.
> 
> I would assume in most of the cases the cause is a 503 or something like that and not a combination of a 200 plus invalid data.

We are using `Retry` from `urllib3.util.retry` under the hood: https://github.com/openSUSE/qem-bot/blob/master/openqabot/utils.py#L54. So if the [documentation](https://urllib3.readthedocs.io/en/stable/reference/urllib3.exceptions.html#urllib3.exceptions.MaxRetryError) is correct the code should raise `MaxRetryError` exception on non 200 status)

perlpunk:
Ah ok. In that case it be interesting to know what data we got in our case and what status code.

perlpunk (approved):
Just a note: I personally would find it more readable if we had the Schema in YAML format

baierjan:
> Ah ok. In that case it be interesting to know what data we got in our case and what status code.
According to the stack trace, it gets past the `ndata["data"]["incidents"]`, so there needs to be at least something.

> Just a note: I personally would find it more readable if we had the Schema in YAML format
I agree, should that schema be more verbose, it surely needs to be at least in a separate file. And yes, it was a little hell to pair those curly braces properly.

pdostal:
This is probably a beginners question, but will this fail if a new item will be added to the JSON schema?

baierjan:
> This is probably a beginners question, but will this fail if a new item will be added to the JSON schema?

Not sure, what do you mean, the schema is "static". The validation will fail, if the response is not containing a required attribute or if the attribute has wrong type. It will not fail, if the result contains an attribute which is not defined by the schema (that could be changed by setting `"additionalProperties": false`).

Additionally in our case, we defined the attributes in the results directly in the query itself (see https://github.com/openSUSE/qem-bot/blob/master/openqabot/loader/smelt.py#L25).

mimi1vx:
@pdostal GraphQL ... it isn't excepted to change schema , and if changes it returns error directly

PR #72, opened 2022-10-12
Don't check product in Kernel-Azure flavor's

mimi1vx:


Martchus (commented):
It would be good if the commit message would state the motivation and had a ticket reference for additional context.

mimi1vx:
> It would be good if the commit message would state the motivation and had a ticket reference for additional context.

it restores old behavior ? And fixes scheduling kernel-azure package

baierjan (approved):

PR #75, opened 2022-10-14
Ensure approval tests actually run despite using decorators

Martchus:


baierjan (approved):

tests/test_approve.py
```diff
@@ -470,9 +470,11 @@ def test_one_aggr_failed(fake_qem, caplog):
 @responses.activate
 @pytest.mark.parametrize("fake_qem", [("NoResultsError isn't raised")], indirect=True)
 @fake_responses_for_unblocking_incidents_via_openqa_comments(incident=2)
+@fake_openqa_comment_api
```
mimi1vx, line 473: this should not be decorator but proper pytest fixture ...

tests/test_approve.py
```diff
@@ -481,9 +483,11 @@ def test_approval_unblocked_via_openqa_comment(fake_qem, caplog):
 @responses.activate
 @pytest.mark.parametrize("fake_qem", [("NoResultsError isn't raised")], indirect=True)
 @fake_responses_for_unblocking_incidents_via_openqa_comments(incident=22)
+@fake_openqa_comment_api
```
mimi1vx, line 486: same as above ....

mimi1vx (changes requested):

Martchus:
Maybe it would still good to merge this for now to get rid of the pipeline failures? Just because I don't know how long it'll take me to refactor this using a pytest fixute. I was already glad to make it work using decorators in general.

mimi1vx (approved):
for fix of the pipeline ..., but it still needs fix formatting @Martchus

PR #80, opened 2022-10-19
Avoid logging 404 errors when getting comments during approval

Martchus:
It seems quite common that the job IDs don't exist anymore and there's no need to clutter logs with according errors.

See https://progress.opensuse.org/issues/107923#note-22

### DIFF
--- a/openqabot/openqa.py
+++ b/openqabot/openqa.py
@@ -74,7 +74,9 @@ class openQAInterface:
             )
             ret = list(map(lambda c: {"text": c.get("text", "")}, ret))
         except Exception as e:
-            logger.exception(e)
+            (method, url, status_code) = e.args
+            if status_code != 404:
+                logger.exception(e)
         return ret
 
     @lru_cache(maxsize=256)
--- a/tests/test_approve.py
+++ b/tests/test_approve.py
@@ -481,7 +481,7 @@ def test_one_incident_failed(fake_qem, fake_openqa_comment_api, caplog):
 
     assert approver() == 0
 
-    assert len(caplog.records) == 8
+    assert len(caplog.records) == 7
     messages = [x[-1] for x in caplog.record_tuples]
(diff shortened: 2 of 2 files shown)


openqabot/openqa.py
```diff
@@ -74,7 +74,9 @@ def get_job_comments(self, job_id: int):
             )
             ret = list(map(lambda c: {"text": c.get("text", "")}, ret))
         except Exception as e:
-            logger.exception(e)
+            (method, url, status_code) = e.args
+            if status_code != 404:
+                logger.exception(e)
```
perlpunk, line 79: Maybe it could still be helpful to at least log that a job was not found.
The exception was too noisy, but a one-line info message?

Martchus: I don't know. Let's see what the rest thinks.

perlpunk: Ok, agreed

perlpunk (approved):

baierjan (approved):

okurz:
Is it that jobs or comments don't exist anymore? Comments not existing would be ok, maybe people removed them to correct. But jobs not existing is more noteworthy

Martchus:
Yes, it is gone. However, it seems to be normal here that job IDs that are referenced aren't existing anymore. Likely there's just no mechanism to catch up with openQA's cleanup of jobs in this workflow.

okurz:
That sounds scary. Are we talking about failed jobs? Can it happen that maintenance updates have bugs, tests fail, then tests eventually just vanish and we approve because failed jobs have been deleted?

Martchus:
No, I suppose we wouldn't just approve (because the failure is still within the qem dashboard, that's also why we hit the case of requesting comments for jobs that don't exist). However, we'd lose the ability to ignore the blocker via an openQA comment. Not sure when old jobs are wiped out from the qem dashboard.

PR #82, opened 2022-10-25
Avoid requesting further comments if there's already one failure

Martchus:
If there's already one failing openQA job blocking the approval we don't need to check further openQA jobs as the incident is blocked anyways.

Related ticket: https://progress.opensuse.org/issues/107923


perlpunk (approved):

okurz (approved):
A test would have been nice :)

PR #83, opened 2022-11-18
Improve approver logging output

okurz:
* approver: Extract _handle_http_error function
* Squeeze some whitespace in code with shorter 'log'
* approver: Extract approvable method
* Add and change log messages in case of approval-blocking tests


baierjan (approved):
This is a subset of #84 right?

okurz:
> This is a subset of #84 right?

correct.

Rebased

PR #85, opened 2022-11-24
Log failed jobs

perlpunk:
Previously we were looking into pipelines and were wondering if we could improve the "Inc x has failed jobs" message by adding the list of failed jobs.

Note that if the list is empty, then there were jobs which were simply not found anymore in openQA.

Related issue: https://progress.opensuse.org/issues/114415

This was a low hanging fruit as I would have touched that code anyway to save the list of failed jobs one level up.

### DIFF
--- a/openqabot/approver.py
+++ b/openqabot/approver.py
@@ -64,13 +64,25 @@ class Approver:
 
                 u_jobs = []
 
-            if not self.get_incident_result(i_jobs, "api/jobs/incident/", inc.inc):
-                logger.info("Inc %s has failed job in incidents" % str(inc.inc))
+            incident_results = self.get_incident_result(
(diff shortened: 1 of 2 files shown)


Martchus (approved):

okurz (changes requested):
Seems you overlooked https://github.com/openSUSE/qem-bot/pull/83 where I already added a log message? My version so far looks simpler. I don't mind if we choose your version but we should be sure we are aware of the both solutions to make a conscious decision

perlpunk:
I hadn't expected someone doing a large refactoring on code while I am assigned to this ticket :(
For example, I also wouldn't do a refactoring of isotovideo now because there is a ticket open that would conflict.

PR #84, opened 2022-11-20
Simplify, cleanup, update code

okurz:
* Simplify loader/qem
* Simplify smeltsync
* Extract 'handle_arch' method in types/aggregates
* Simplify counter assignment in aggregates
* tests: Simplify test_repohash
* tests: Simplify remaining approver generators
* tests: Extract 'approver' helper in test_approve
* tests: Extract "two_passed_responses" helper
* Simplify tests
* tests: Skip 'responses' test if not available on older python
* Use updated codecov uploader v3
* approver: Extract _handle_http_error function
* Squeeze some whitespace in code with shorter 'log'
* approver: Extract approvable method
* Add and change log messages in case of approval-blocking tests

PR #88, opened 2022-11-25
Fix boolean return values

perlpunk:
Followup to #84 which caused a regression

Issue: https://progress.opensuse.org/issues/120973

When doing something like

    return array and x == y

then it will return array and not a boolean if array is empty


baierjan (approved):

kraih (approved):

okurz:
Thank you!

PR #92, opened 2022-11-30
Proof of Concept: Post a comment

perlpunk:
Issue: https://progress.opensuse.org/issues/114415

Usage:
```
./bot-ng.py -c /etc/openqabot --token abc --dry -c tests/fixtures/config/ inc-approve --incident 25842 --post-comment
```
To post an actual comment, remove the `--dry`, but beware that you might also approve the incident.

I think using the `Commenter` interface doesn't make much sense, as it currently expects a namespace argument for its constructor.
But if we are never going to use the comment subcommand, then we can rewrite the Commenter interface, of course.


perlpunk:
rebased and added a small (still commented) test.

PR #99, opened 2022-12-07
Squeeze some whitespace in code with shorter 'log'

okurz:


baierjan (approved):

mimi1vx (approved):
most of PR is unnecessary.. but why not

PR #101, opened 2022-12-20
Slightly simplify

okurz:
* Simplify counter assignment in aggregates
* approver: Extract _handle_http_error function

### DIFF
--- a/openqabot/types/aggregate.py
+++ b/openqabot/types/aggregate.py
@@ -47,18 +47,14 @@ class Aggregate(BaseConf):
         return f"<Aggregate product: {self.product}>"
 
     @staticmethod
-    def get_buildnr(repohash: str, old_repohash: str, cur_build: str) -> str:
+    def get_buildnr(repohash: str, old_repohash: str, build: str) -> str:
         today = date.today().strftime("%Y%m%d")
         build = today
 
-        if cur_build.startswith(today) and repohash == old_repohash:
+        if build.startswith(today) and repohash == old_repohash:
             raise SameBuildExists
 
-        if cur_build.startswith(today):
-            counter = int(cur_build.split("-")[-1]) + 1
-        else:
-            counter = 1
-
+        counter = int(build.split("-")[-1]) + 1 if build.startswith(today) else 1
         return f"{build}-{counter}"
 
     def __call__(
--- a/openqabot/approver.py
+++ b/openqabot/approver.py
@@ -28,6 +28,27 @@ from .utils import retry3 as requests
 log = getLogger("bot.approver")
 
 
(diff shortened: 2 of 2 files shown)


mimi1vx (commented):
Small stupid question ? Why ..

This isnt perl to have everything on one line ...

amd why extract code to separatred function to use in one place ?

okurz:
> Small stupid question ? Why ..
> 
> This isnt perl to have everything on one line ...

This does not have to do with Perl or Python but is applying the generic concept of conditional value assignment. Less procedural code when the only thing we want is actually a value "counter" applied, just with values depending on conditions

> amd why extract code to separatred function to use in one place ?

to reduce complexity on each level with abstraction. A side-effect is less indendation. Same reason why the Linux kernel coding style only allows two levels of indendation.

mimi1vx:
> > Small stupid question ? Why ..
> > This isnt perl to have everything on one line ...
> 
> This does not have to do with Perl or Python but is applying the generic concept of conditional value assignment. Less procedural code when the only thing we want is actually a value "counter" applied, just with values depending on conditions
> 
> > amd why extract code to separatred function to use in one place ?
> 
> to reduce complexity on each level with abstraction. A side-effect is less indendation. Same reason why the Linux kernel coding style only allows two levels of indendation.

so .. lower readability for better code .., btw kernel has millions of lines so some restrictive rules are really needed

mimi1vx (approved):

grisu48:
I'm late for the party but also like to add that it's in general not an improvement to make code more compact and less readable. See [the rule of six](https://davidamos.dev/the-rule-of-six/).

Less code that does more typically means increased mental effort to understand what's going on.

openqabot/types/aggregate.py
```diff
@@ -47,18 +47,14 @@ def __repr__(self):
         return f"<Aggregate product: {self.product}>"
 
     @staticmethod
-    def get_buildnr(repohash: str, old_repohash: str, cur_build: str) -> str:
+    def get_buildnr(repohash: str, old_repohash: str, build: str) -> str:
         today = date.today().strftime("%Y%m%d")
         build = today
```
baierjan, line 52: Variable name clash here, see #104

openqabot/types/aggregate.py
```diff
@@ -47,18 +47,14 @@ def __repr__(self):
...
+        if build.startswith(today) and repohash == old_repohash:
             raise SameBuildExists
 
-        if cur_build.startswith(today):
-            counter = int(cur_build.split("-")[-1]) + 1
-        else:
-            counter = 1
-
+        counter = int(build.split("-")[-1]) + 1 if build.startswith(today) else 1
         return f"{build}-{counter}"
```
baierjan, line 58: and here

PR #103, opened 2022-12-21
Extract 'handle_arch' method in types/aggregates

okurz:
https://progress.opensuse.org/issues/107923

### DIFF
--- a/openqabot/types/aggregate.py
+++ b/openqabot/types/aggregate.py
@@ -57,6 +57,151 @@ class Aggregate(BaseConf):
         counter = int(build.split("-")[-1]) + 1 if build.startswith(today) else 1
         return f"{build}-{counter}"
 
+    def handle_arch(
+        self,
+        arch,
+        incidents: List[Incident],
+        token: Dict[str, str],
+        ci_url: Optional[str],
+        ignore_onetime: bool = False,
+    ) -> Dict[str, Any]:
+        full_post: Dict["str", Any] = {}
+        full_post["openqa"] = {}
+        full_post["qem"] = {}
+        full_post["qem"]["incidents"] = []
+        full_post["qem"]["settings"] = {}
+        full_post["api"] = "api/update_settings"
+        if ci_url:
+            full_post["openqa"]["__CI_JOB_URL"] = ci_url
+
+        test_incidents = defaultdict(list)
+        test_repos = defaultdict(list)
+
+        # Temporary workaround for applying the correct architecture on jobs, which use a helper VM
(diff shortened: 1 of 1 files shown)


okurz:
I would like to keep this open until more people return after Christmas absence.

openqabot/types/aggregate.py
```diff
@@ -57,157 +57,158 @@ def get_buildnr(repohash: str, old_repohash: str, build: str) -> str:
...
+    def __call__(
+        self,
+        incidents: List[Incident],
+        token: Dict[str, str],
+        ci_url: Optional[str],
+        ignore_onetime: bool = False,
+    ) -> List[Dict[str, Any]]:
+        return [
+            self.handle_arch(incidents, token, ci_url, ignore_onetime)
+            for arch in self.archs
```
perlpunk, line 213: I think in the previous PR we saw that `arch` is not defined in `handle_arch` which resulted in a regression in production.
`arch` should be passed as an additional argument.

perlpunk (changes requested):
See comment about `handle_arch`

openqabot/types/aggregate.py
okurz: I thought the regression was fixed in https://github.com/openSUSE/qem-bot/pull/89

okurz: but, yes, the arch argument is still missing. I am more concerned if no CI checks fail and we don't see a report about missing coverage

perlpunk: The regression that was fixed was that it should be called as a method, but after we deployed that we still saw that `arch` was undefined.
And then we reverted the whole PR...

perlpunk: I am and was actually concerned about the fact why this compiled but the `arch` variable was undefined.
And I wasn't able to reproduce this locally with a simple function. In that case `arch` was defined.

openqabot/types/aggregate.py
```diff
@@ -57,157 +57,159 @@ def get_buildnr(repohash: str, old_repohash: str, build: str) -> str:
...
+            full_post["openqa"][template] = ",".join(issues)
+        for issues in test_incidents.values():
+            full_post["qem"]["incidents"] += issues
+
+        full_post["qem"]["incidents"] = [
+            str(inc) for inc in set(full_post["qem"]["incidents"])
+        ]
+
+        if not full_post["qem"]["incidents"]:
+            return {}
```
baierjan, line 186: This will ultimately create a different list than in the original code. Instead of skipping invalid items, this will add an empty one, e.g. you can end up with something like: `[{},{},{}]`

PR #105, opened 2022-12-23
Enhance JobAggr to also have openQA job id and use it

baierjan:
Use openQA job id instead of dashboard openqa_jobs.id where applicable. This will also fix unblocking incidents with openQA comment.

Reference: https://progress.opensuse.org/issues/122308

### DIFF
--- a/openqabot/approver.py
+++ b/openqabot/approver.py
@@ -119,7 +119,7 @@ class Approver:
     def is_job_marked_acceptable_for_incident(self, job: JobAggr, inc: int) -> bool:
         regex = re.compile(r"\@review\:acceptable_for\:incident_%s\:(.+)" % inc)
         try:
-            for comment in self.client.get_job_comments(job.job_id):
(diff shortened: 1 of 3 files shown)


Martchus (approved):
That clears up the confusion I had on Wednesday when reading the code (https://progress.opensuse.org/issues/122308#note-4). So the ID is in fact so far *not* an openQA job ID - and somehow everybody overlooked it in the pair programming session where we added the feature to read openQA comments.

okurz (changes requested):
I suggest you also rename the variable "job" to e.g. "job_aggr"

mimi1vx (approved):

baierjan:
> I suggest you also rename the variable "job" to e.g. "job_aggr"

I propose to solve that inside #106 together with some more (potentially dangerous) polishing/renaming.

okurz (approved):

PR #106, opened 2022-12-29
Refactor JobAggr to comply with dashboard database

baierjan:
The SQL schema for dashboard uses a slightly different terminology for some data. Rename some attributes to avoid confusion.

* rename job variable to job_aggr for clarity
* rename attribute job_id to just id
* adapt log messages to use setting id when referring to JobAggr.id

Reference: https://progress.opensuse.org/issues/122308

### DIFF
--- a/openqabot/approver.py
+++ b/openqabot/approver.py
@@ -116,10 +116,12 @@ class Approver:
         return True
 
     @lru_cache(maxsize=512)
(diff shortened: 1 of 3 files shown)


okurz (approved):

baierjan:
Let's try again, refactor first. The "unblock via comment" feature is not working as intended (the current code is using the wrong id)

okurz (approved):

kraih (approved):

PR #109, opened 2023-01-02
Fix unblocking with comments

baierjan:
Needs:
* #106

After the refactoring, it is quite clear how to fix the issue with the comment fetching.

### DIFF
--- a/openqabot/approver.py
+++ b/openqabot/approver.py
@@ -116,10 +116,12 @@ class Approver:
         return True
 
     @lru_cache(maxsize=512)
-    def is_job_marked_acceptable_for_incident(self, job: JobAggr, inc: int) -> bool:
+    def is_job_marked_acceptable_for_incident(
+        self, job_aggr: JobAggr, inc: int
+    ) -> bool:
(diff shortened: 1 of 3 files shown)


openqabot/approver.py
```diff
@@ -116,51 +116,51 @@ def _approvable(self, inc: IncReq) -> bool:
...
-                    % (job.job_id, inc)
+                    % (res["job_id"], inc)
                 )
                 res["status"] = "passed"
             else:
                 log.info(
-                    "Found failed, not-ignored job %s for incident %s"
-                    % (job.job_id, inc)
+                    "Found failed, not-ignored job setting %s for incident %s"
+                    % (job_aggr.id, inc)
```
okurz, line 149: Can we also use something like `res["job_id"]` here to refer to real openQA job IDs?

baierjan: Ah, yes... in this context we can. I will update the PR.

okurz (approved):

kraih (approved):

PR #118, opened 2023-06-02
Mark missing jobs as obsolete in QEM dashboard

Martchus:
Related ticket: https://progress.opensuse.org/issues/126551

### DIFF
--- a/openqabot/approver.py
+++ b/openqabot/approver.py
@@ -59,7 +59,7 @@ class Approver:
         self.single_incident = args.incident
         self.token = {"Authorization": "Token {}".format(args.token)}
         self.all_incidents = args.all_incidents
-        self.client = openQAInterface(args.openqa_instance)
+        self.client = openQAInterface(args)
 
     def __call__(self) -> int:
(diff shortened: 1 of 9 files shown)


Martchus:
Note that the CI check passes locally (with a slightly older requests version but otherwise the same versions). I don't know why it doesn't pass in the CI. The failing request is supposed to be mocked but this apparently doesn't work.

okurz:
I could reproduce the CI problem in an ubuntu container as well as a Tumbleweed container so the interesting question is really why it can even work for you :)

perlpunk:
Reproduced with requests==2.31.0, but passing with requests==2.30.0
So something changed in requests that make the responses mocking fail

Martchus:
I've just reproduced it locally on the master branch. So the failure is definitely unrelated to my changes. Maybe it would be useful to downgrade urllib3 for now within our CI.

okurz (approved):

kraih (approved):

perlpunk (approved):

PR #119, opened 2023-06-12
Run CI checks on Tumbleweed

Martchus:


baierjan (approved):
Seems like a nice solution. Just for the record, the production environment runs on Leap.

baierjan (changes requested):
And it seems that the python action needs to be replaced by native zypper call (or rather image with python included)?

Martchus:
> Just for the record, the production environment runs on Leap.

I know but I think it is ok. The Ubuntu-based CI also used a more recent Python version than we probably use in production. (In fact, it also Python 3.10 - the same as on Tumbleweed.)

> And it seems that the python action needs to be replaced by native zypper call (or rather image with python included)?

Yes, I'm currently preparing an image.

baierjan (approved):

PR #120, opened 2023-06-12
Fix CI checks by avoiding incompatible Python modules

Martchus:


requirements-dev.txt
```diff
@@ -9,3 +9,4 @@ ruamel.yaml
 beautifulsoup4
 black
 jsonschema
+urllib3<2
```
kalikiana, line 12: What problem are we solving with this?

kalikiana: I suggest a comment, ideally with a poo# reference, explaining why to avoid somebody being smart and undoing whatever you're fixing since we're already getting the dep elsewhere.

perlpunk: Is there an issue for the responses module maybe that we could link to?

Martchus: So the problem has already been resolved by upstream but it hasn't been released yet.

perlpunk (approved):

requirements-dev.txt
kalikiana: I would still mention that here.

kalikiana (approved):

PR #121, opened 2023-06-20
pc_helper: Add tests and cleanup unused code

asmorodskyi:
PR contains several commits :
-   Add ability allowing to check Public Cloud related logic outside of bot-ng
-   Drop function because it is not used anymore
-   Drop use of variable PUBLIC_CLOUD_PINT_QUERY    
-   Add tests for pc_helper
-   Switch to pythonic documentation format

PR #123, opened 2023-06-22
pc_helper: Add tests and cleanup unused code

asmorodskyi:
PR contains several commits :

    - Add ability allowing to check Public Cloud related logic outside of bot-ng
    - Drop function because it is not used anymore
    - Drop use of variable PUBLIC_CLOUD_PINT_QUERY
    - Add tests for pc_helper
    - Switch to pythonic documentation format
    - Fix logging in pc_helper 
    - Call python3 binary in officially recommended way
 

Related ticket : https://progress.opensuse.org/issues/115211

NOTE : this is second attempt of merging #121 which was reverted #122  due to exception thrown during run . Difference between #121  and this one is that in case of failure we not logging the problem in caller but only inside called function . Also I add extra commit which fixing logging within pc_helper

### DIFF
--- a/openqabot/main.py
+++ b/openqabot/main.py
@@ -1,25 +1,13 @@
 # Copyright SUSE LLC
 # SPDX-License-Identifier: MIT
-import logging
 import sys
 
+from .utils import create_logger
 from .args import get_parser
 
 
-def create_logger() -> logging.Logger:
-    log = logging.getLogger("bot")
-    handler = logging.StreamHandler()
-    formatter = logging.Formatter(
-        fmt="%(asctime)s %(levelname)-8s %(message)s", datefmt="%Y-%m-%d %H:%M:%S"
-    )
-    handler.setFormatter(formatter)
-    log.addHandler(handler)
-    log.setLevel(logging.INFO)
-    return log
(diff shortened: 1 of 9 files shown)


okurz (changes requested):
Please

* Reference a ticket for your work
* Fix CI
* Explain what you changed in relation to your previous attempt
* Definitely link the previous attempt PR!

bot-ng.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
```
baierjan, line 1: That's the exact opposite of what `env-script-interpreter` rpmlint check wants.

okurz:
Please look into the discussion in https://suse.slack.com/archives/C02CANHLANP/p1687438881079609?thread_ts=1687361176.085319&cid=C02CANHLANP where we discussed backward compatible changes. If you want to stay with the current implementation regardless I suggest we merge this and the mentioned merge request when you are available for longer after that, e.g. start of a business day

mimi1vx (approved):

okurz (approved):

PR #124, opened 2023-06-29
Fix debug option broken by #123


asmorodskyi:
Import of logging package was removed in https://github.com/asmorodskyi/qem-bot/commit/5d66d5e9c271ed24765425d64570ccf6d44311b5
but it is still needed by setting DEBUG log level.
Revert import removal solve the problem


okurz (approved):

okurz:
Deploying together with https://github.com/openSUSE/qem-bot/pull/126

PR #126, opened 2023-06-29
Teach pc_helper_online to process metadata

asmorodskyi:
Initial version was doing only two function calls with dummy data
This time we taking same folder with metadat as bot-ng. But instead
of going through all actions done by bot-ng we concentrating on two variables
which matter for Public Cloud. PUBLIC_CLOUD_TOOLS_IMAGE_QUERY and PUBLIC_CLOUD_PINT_QUERY

### DIFF
--- a/pc_helper_online.py
+++ b/pc_helper_online.py
@@ -1,6 +1,9 @@
 #!/usr/bin/env python3
 # Copyright SUSE LLC
 # SPDX-License-Identifier: MIT
+from argparse import ArgumentParser
+from pathlib import Path
+from ruamel.yaml import YAML
 from openqabot.pc_helper import apply_pc_tools_image, apply_publiccloud_pint_image
 from openqabot.utils import create_logger
 
@@ -9,35 +12,44 @@ def main():
     """
     This code is used only for testing purpose.
     Allowing to prove that Public Cloud related logic is actually working without executing
-    a lot of code which is unrelated to pc_helper
+    a lot of code which is unrelated to pc_helper.
+    As input it getting directory with openqabot configuration metadata (same folder as bot-ng )
+    but processing only variables related to openqabot.pc_helper module
     """
     log = create_logger("pc_helper_online")
-    settings = {
-        "PUBLIC_CLOUD_TOOLS_IMAGE_QUERY": "https://openqa.suse.de/group_overview/276.json"
-    }
-    log.info("Calling apply_pc_tools_image with :\n%s", settings)
-    apply_pc_tools_image(settings)
-    log.info("Setting after the call to apply_pc_tools_image:\n%s", settings)
-
-    settings = {
-        "PUBLIC_CLOUD_PINT_QUERY": "https://susepubliccloudinfo.suse.com/v1/amazon/images/",
-        "PUBLIC_CLOUD_PINT_NAME": "suse-sles-12-sp5-v[0-9]{8}-hvm-ssd-x86_64",
-        "PUBLIC_CLOUD_PINT_REGION": "us-east-1",
-        "PUBLIC_CLOUD_PINT_FIELD": "id",
-    }
-    log.info("Calling apply_publiccloud_pint_image with :\n%s", settings)
-    apply_publiccloud_pint_image(settings)
-    log.info("Setting after the call to apply_publiccloud_pint_image:\n%s", settings)
-
-    settings = {
-        "PUBLIC_CLOUD_PINT_QUERY": "https://susepubliccloudinfo.suse.com/v1/amazon/images/",
-        "PUBLIC_CLOUD_PINT_NAME": "suse-sles-15-sp2-chost-byos-v[0-9]{8}-hvm-ssd-x86_64",
-        "PUBLIC_CLOUD_PINT_REGION": "us-east-1",
-        "PUBLIC_CLOUD_PINT_FIELD": "id",
-    }
-    log.info("Calling apply_publiccloud_pint_image with :\n%s", settings)
-    apply_publiccloud_pint_image(settings)
-    log.info("Setting after the call to apply_publiccloud_pint_image:\n%s", settings)
+    parser = ArgumentParser(
+        prog="pc_helper_online",
+        description="Dummy code to test functionality related to pc_helper code",
+    )
+    parser.add_argument(
+        "-c",
+        "--configs",
+        type=Path,
+        default=Path("/etc/openqabot"),
+        help="Directory with openqabot configuration metadata",
+    )
+    args = parser.parse_args()
+    log.info(f"Parsing configuration files from {args.configs}")
+    loader = YAML(typ="safe")
+    for p in Path(args.configs).glob("*.yml"):
+        try:
+            data = loader.load(p)
+            log.info(f"Processing {p}")
+            if "settings" in data:
+                settings = data["settings"]
+                if "PUBLIC_CLOUD_TOOLS_IMAGE_QUERY" in settings:
+                    apply_pc_tools_image(settings)
+                    if "PUBLIC_CLOUD_TOOLS_IMAGE_BASE" not in settings:
+                        log.error(
+                            f"Failed to get PUBLIC_CLOUD_TOOLS_IMAGE_BASE from {data}"
+                        )
+                if "PUBLIC_CLOUD_PINT_QUERY" in settings:
+                    apply_publiccloud_pint_image(settings)
+                    if "PUBLIC_CLOUD_IMAGE_ID" not in settings:
+                        log.error(f"Failed to get PUBLIC_CLOUD_IMAGE_ID from {data}")
+        except Exception as e:
+            log.exception(e)
+            continue
 
 
 if __name__ == "__main__":


pc_helper_online.py
```diff
@@ -9,35 +12,43 @@ def main():
...
+                    apply_pc_tools_image(settings)
+                    if not settings.get("PUBLIC_CLOUD_TOOLS_IMAGE_BASE", False):
+                        log.error(
+                            f"Failed to get PUBLIC_CLOUD_TOOLS_IMAGE_BASE from {data}"
+                        )
+                if "PUBLIC_CLOUD_PINT_QUERY" in settings:
+                    apply_publiccloud_pint_image(settings)
+                    if not settings.get("PUBLIC_CLOUD_IMAGE_ID", False):
+                        log.error(f"Failed to get PUBLIC_CLOUD_IMAGE_ID from {data}")
+        except Exception as e:
```
ricardobranco777, line 49: Since we're using broad exception catching, we could as well avoid checking the keys in the `data` and `settings` dict's.

pc_helper_online.py
```diff
@@ -9,35 +12,43 @@ def main():
...
+                settings = data["settings"]
+                if "PUBLIC_CLOUD_TOOLS_IMAGE_QUERY" in settings:
+                    apply_pc_tools_image(settings)
+                    if not settings.get("PUBLIC_CLOUD_TOOLS_IMAGE_BASE", False):
+                        log.error(
+                            f"Failed to get PUBLIC_CLOUD_TOOLS_IMAGE_BASE from {data}"
+                        )
+                if "PUBLIC_CLOUD_PINT_QUERY" in settings:
+                    apply_publiccloud_pint_image(settings)
+                    if not settings.get("PUBLIC_CLOUD_IMAGE_ID", False):
```
ricardobranco777, line 47:
```suggestion
                    if "PUBLIC_CLOUD_IMAGE_ID" not in settings:
```

pc_helper_online.py
```diff
@@ -9,35 +12,43 @@ def main():
...
+    log.info(f"Parsing configuration files from {args.configs}")
+    for p in Path(args.configs).glob("*.yml"):
+        try:
+            data = loader.load(p)
+            log.info(f"Processing {p}")
+            if "settings" in data:
+                settings = data["settings"]
+                if "PUBLIC_CLOUD_TOOLS_IMAGE_QUERY" in settings:
+                    apply_pc_tools_image(settings)
+                    if not settings.get("PUBLIC_CLOUD_TOOLS_IMAGE_BASE", False):
```
ricardobranco777, line 41:
```suggestion
                    if "PUBLIC_CLOUD_TOOLS_IMAGE_BASE" not in settings:
```

pc_helper_online.py
```diff
@@ -9,35 +12,43 @@ def main():
...
+        default=Path("/etc/openqabot"),
+        help="Directory with openqabot configuration metadata",
+    )
+    args = parser.parse_args()
+    loader = YAML(typ="safe")
+    log.info(f"Parsing configuration files from {args.configs}")
+    for p in Path(args.configs).glob("*.yml"):
+        try:
+            data = loader.load(p)
+            log.info(f"Processing {p}")
```
ricardobranco777, line 36:
```suggestion
            log.info(f"Processing {p}")
            data = loader.load(p)
```
I would log first in case `loader.load()` throws an exception but it's a matter of choice anyway.

ricardobranco777 (changes requested):

pc_helper_online.py
asmorodskyi: done

pc_helper_online.py
asmorodskyi: done

pc_helper_online.py
asmorodskyi: done

Martchus (commented):
How is this tested?

ricardobranco777 (approved):

pc_helper_online.py
asmorodskyi: > It's an anti-pattern frowned upon by linters for good reason.

which is ?

> In the case of libraries they usually define a base class for the exceptions they throw

the problem is that you using some libraries which using other libraries . Also undefined combination of different OS errors so at the end you always need several iterations to define all possible error types you want to catch so your loop will keep going in case something will happen. By "iteration" I mean here that you running your program and it get interrupted by new flavor of exception **unexpectedly** ( because you NOT expecting your loop to be interrupted no matter what)  so you creating new commit which add yet-another-handler  . And here comes the fun part **which is doing exactly the same what all other handlers defined before were doing** - log the error and continue .
For me rule is different - **define special error handlers when you have different error handling based on exception type **  E.g. if you got network error switch to offline mode ( if you app has it ) or when you got parsing error of config file try to generate default config . But if your reaction to exception is log the error and continue it does not matter how many handler you have

asmorodskyi:
> How is this tested?

this code meant to test https://github.com/openSUSE/qem-bot/blob/master/openqabot/pc_helper.py function calls with real data against real API endpoints . What kind of testing do you want to have ? `pc_helper.py` code already covered by corresponding pytests . Code inside `pc_helper_online.py` does not hold any value , it is test code itself ( which is stated in python doc-string )

pc_helper_online.py
ricardobranco777: > > It's an anti-pattern frowned upon by linters for good reason.
> 
> which is ?

I should have used plural.  You just mentioned one.  I could add that code with general exception catching can't be reused as functions by code that for some reason needs to catch an exception.

And the problem, again, is not just the general exception catching but the big block within the try/except which makes it harder to find where exactly the exception happened.

This code prints 2:

```
try:
    d = {}
    d[1] = 3
    print(d[2])
except Exception as e:
    print("%s" % e)
```

> the problem is that you using some libraries which using other libraries .

Those libraries should define a subclass of Exception and re-raise most exceptions using it.  But general exception catching makes the problem worse, IMO.

> Also undefined combination of different OS errors so at the end you always need several iterations to define all possible error types you want to catch so your loop will keep going in case something will happen.

I believe that every OS error is covered by `OSError`.

asmorodskyi:
> > How is this tested?
> 
> this code meant to test https://github.com/openSUSE/qem-bot/blob/master/openqabot/pc_helper.py function calls with real data against real API endpoints . What kind of testing do you want to have ? `pc_helper.py` code already covered by corresponding pytests . Code inside `pc_helper_online.py` does not hold any value , it is test code itself ( which is stated in python doc-string )

1. obviously I run it locally and saw that it produce expected output 
2. Some background of this PR . I was trying investigate WARNING from https://gitlab.suse.de/qa-maintenance/bot-ng/-/jobs/1657326/raw (`2023-06-27 09:16:52 WARNING  PUBLIC_CLOUD_PINT_QUERY handling failed for sles-15-sp3-v[0-9]{8}: Cannot find matching image in pint`). But I can not reproduce the issue **because this part of code executed ONLY without --dry** . So after trying different things I came up with this PR and manage to reproduce the problem !

pc_helper_online.py
asmorodskyi:
```
try:
    d = {}
    d[1] = 3
    print(d[2])
except Exception as e:
    print("%s" % e)
```
I would say this is more about bad exception handling . I can fix it without breaking general exception handling . 

```
import traceback
try:
    d = {}
    d[1] = 3
    print(d[2])
except Exception as e:
    traceback.print_exception(e)
```

can you give me code snippet how this code can be improved by catching more special exception ?

asmorodskyi: > Those libraries should define a subclass of Exception and re-raise most exceptions using it 
we speaking about code which **using** this libraries not how everything should be written in ideal world :)

grisu48 (commented):
Overall OK, but I'm still puzzled about the motivation for this PR.

Also, if I got it right, wouldn't it be better to pass a list of yaml files instead of a directory and processing all of them? I mean, the purpose is to check if the internal handling gives the correct results, limiting the input then to 1-2 dedicated PC-related yaml files instead of shoving SAP and whatnot into the little helper script might be more aimed at the task, no?

pc_helper_online.py
asmorodskyi: And getting back to my main argument : 

How this is better 

```
for i in i_array:

    try:
        do_something(i)
        do_something2(i)
        do_something3(i)
    except OSError as e:
        log.error("Something happened" % )
    except KeyError as e:
        log.error("Something happened" % )
     except ValueError as e:
       log.error("Something happened" % )
```

than this : 

```
for i in i_array:

    try:
        do_something(i)
        do_something2(i)
        do_something3(i)
    except Exception as e:
        log.error("Something happened" % )
```

? 

or you speaking about "too long blocks in try/catch " so it getting even worth : 

```
```
for i in i_array:

    try:
        do_something(i)
    except OSError as e:
        log.error("Something happened" % )
    except KeyError as e:
        log.error("Something happened" % )
     except ValueError as e:
       log.error("Something happened" % )

    try:
        do_something2(i)
        do_something3(i)
    except OSError as e:
        log.error("Something happened" % )
    except KeyError as e:
        log.error("Something happened" % )
     except ValueError as e:
       log.error("Something happened" % )
```

**Comment by asmorodskyi** [comment]
> Overall OK, but I'm still puzzled about the motivation for this PR.

 I was trying investigate WARNING from https://gitlab.suse.de/qa-maintenance/bot-ng/-/jobs/1657326/raw (2023-06-27 09:16:52 WARNING  PUBLIC_CLOUD_PINT_QUERY handling failed for sles-15-sp3-v[0-9]{8}: Cannot find matching image in pint). But I can not reproduce the issue because this part of code executed ONLY without --dry . So after trying different things I came up with this PR and manage to reproduce the problem !

> 
> Also, if I got it right, wouldn't it be better to pass a list of yaml files instead of a directory and processing all of them? I mean, the purpose is to check if the internal handling gives the correct results, limiting the input then to 1-2 dedicated PC-related yaml files instead of shoving SAP and whatnot into the little helper script might be more aimed at the task, no?

1. I am trying to stay as much as possible close to the code executed by bot-ng 
2. What the problem of executing ALL of configuration files ? Time ? Currently overall execution takes 30 seconds 
3. Currently problem pop ups only for SINGLE file so if we limit execution to limited amount of items we most probably skip it

File: pc_helper_online.py (hunk 1 above)
Comment (ricardobranco777) [reply pc_helper_online.py:49]: I shouldn't have used that code snippet.

> How this is better
> 
> ```
> for i in i_array:
> 
>     try:
>         do_something(i)
>         do_something2(i)
>         do_something3(i)
>     except OSError as e:
>         log.error("Something happened" % )
>     except KeyError as e:
>         log.error("Something happened" % )
>      except ValueError as e:
>        log.error("Something happened" % )
> ```
> 
> than this :
> 
> ```
> for i in i_array:
> 
>     try:
>         do_something(i)
>         do_something2(i)
>         do_something3(i)
>     except Exception as e:
>         log.error("Something happened" % )
> ```
> 
> ?

If the only thing we do is log the error then `except OSError, KeyError, ValueError as e:` is better.

Comment (ricardobranco777) [reply pc_helper_online.py:49]: > > Those libraries should define a subclass of Exception and re-raise most exceptions using it
> 
> we speaking about code which **using** this libraries not how everything should be written in ideal world :)

But the YAML library the script is using is well known and you're calling it inside a big try/except block.

File: pc_helper_online.py [hunk 5]
```diff
@@ -9,35 +12,43 @@ def main():
     """
     This code is used only for testing purpose.
     Allowing to prove that Public Cloud related logic is actually working without executing
-    a lot of code which is unrelated to pc_helper
+    a lot of code which is unrelated to pc_helper.
+    As input it getting same folder as bot-ng but processing only two variables which are related to Public Cloud
```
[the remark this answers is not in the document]
asmorodskyi: Yeah it was not ideal but your version also raising some questions from my side . I tried to improve  mine . If you still don't like it please be more specific than "???" what actually you don't like there

pc_helper_online.py
asmorodskyi: > If the only thing we do is log the error then except OSError, KeyError, ValueError as e: is better.

I still didn't get how it is better . Only code example which you gave contained problem unrelated to the one which we discussing here . 

So you saying that : 

```
for i in i_array:

    try:
        do_something(i)
        do_something2(i)
        do_something3(i)
    except OSError, KeyError, ValueError as e:
          log.error("Something happened" % )
```
is better than : 

```
for i in i_array:
    try:
        do_something(i)
        do_something2(i)
        do_something3(i)
    except Exception as e:
          log.error("Something happened" % )
```

and I am saying that it is worse because when tomorrow do_something2 will change the code and start throwing also some JSONException or any other NeverSeenBEforeException I will need to come up with extra PR extending this list : 
```
for i in i_array:

    try:
        do_something(i)
        do_something2(i)
        do_something3(i)
    except OSError, KeyError, ValueError, NeverSeenBEforeException as e:
          log.error("Something happened" % )
```

and this is obvious price which you need to pay - when you have exact list you need to keep it up to date with your code . But it is unclear what exactly I will get for this "money" ? Just feeling that I am following some pattern written in some book ? Sorry it is not enough for me . I need something REAL

pc_helper_online.py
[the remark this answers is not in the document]
grisu48: `???` typically means that I have no idea what this means :sweat_smile: I wish I could, but that also means that I cannot be more specific.

I gave you a suggestion for rewording, which might reflect what you want to say, but since I couldn't understand what you're saying this is really the best I could come up with.

grisu48:
> > Overall OK, but I'm still puzzled about the motivation for this PR.
> 
> I was trying investigate WARNING from https://gitlab.suse.de/qa-maintenance/bot-ng/-/jobs/1657326/raw (2023-06-27 09:16:52 WARNING PUBLIC_CLOUD_PINT_QUERY handling failed for sles-15-sp3-v[0-9]{8}: Cannot find matching image in pint). But I can not reproduce the issue because this part of code executed ONLY without --dry . So after trying different things I came up with this PR and manage to reproduce the problem !
> 
> > Also, if I got it right, wouldn't it be better to pass a list of yaml files instead of a directory and processing all of them? I mean, the purpose is to check if the internal handling gives the correct results, limiting the input then to 1-2 dedicated PC-related yaml files instead of shoving SAP and whatnot into the little helper script might be more aimed at the task, no?
> 
>     1. I am trying to stay as much as possible close to the code executed by bot-ng
> 
>     2. What the problem of executing ALL of configuration files ? Time ? Currently overall execution takes 30 seconds
> 
>     3. Currently problem pop ups only for SINGLE file so if we limit execution to limited amount of items we most probably skip it

Yeah, but for debugging most of those problems you only need to execute it against one single yaml file and not all of them. This reduces runtime, verbosity and allows you to debug it much more specific.

In short: You then could still run it against all of them, but you could also run it against one carefully designed yaml file, which allows you to debug it more specific.

asmorodskyi:
> Yeah, but for debugging most of those problems you only need to execute it against one single yaml file and not all of them. This reduces runtime, verbosity and allows you to debug it much more specific.

1. It is not just about debugging some problem . In current version of code we checking pc_helper AND metadata by doing **exactly** the same what bot-ng is doing but minus all requests to smelt and qem-dashboard. If we fallback to execution of some synthetic yaml we would certainly miss this problem https://gitlab.suse.de/qa-maintenance/metadata/-/merge_requests/887 
2. "This reduces runtime, verbosity and allows you to debug it much more specific." - all this already addressed by cutting out all requests to smelt and qem-dashboard . Currently execution of pc_helper_online takes 30 seconds and output 130 lines which in comparison to bot-ng execution is huge improvement . Cutting it further will complicated **main** use case of this file which is "reproduce behavior of bot-ng without smelt and qem-dashboard "

> 
> In short: You then could still run it against all of them, but you could also run it against one carefully designed yaml file, which allows you to debug it more specific.

If you really want to run against single file you can easily do this with current codebase - just create dummy folder and put single file there and you are done 

> 
> This change would be rather cheap, and allowing your fellow users to have a more precise tool than a sledgehammer is a courtesy which IMHO is appreciated slightly_smiling_face

yeah programmatically it is cheap but it will complicated main use case where we need to do exactly the same what bot-ng is doing and bot-ng is taking whole metadata folder

pc_helper_online.py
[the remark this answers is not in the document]
asmorodskyi: I haven't blindly copy/paste your suggestion but I re-word my version . Is it better now ? Do you understand new version ?

pc_helper_online.py
asmorodskyi: I think we are getting to deep into rabbit hole :) let's move this discussion out of github PR ...

okurz (approved):
@asmorodskyi please advise when you are ready to monitor the effect after merge. Then we can remove the "not-ready" flag

asmorodskyi:
> @asmorodskyi please advise when you are ready to monitor the effect after merge. Then we can remove the "not-ready" flag

I would like first close discussion with @grisu48  ( get his LGTM )

grisu48:
> > @asmorodskyi please advise when you are ready to monitor the effect after merge. Then we can remove the "not-ready" flag
> 
> I would like first close discussion with @grisu48 ( get his LGTM )

I give up and LGTM the changes.

grisu48 (approved):

PR #127, opened 2023-07-03
Embargoed: implement skipping of embargoed updates

asmorodskyi:
Embargoed: implement skipping of embargoed updates

### DIFF
--- a/openqabot/loader/smelt.py
+++ b/openqabot/loader/smelt.py
@@ -25,7 +25,8 @@ after: "%(cursor)s" ) { pageInfo { hasNextPage endCursor} edges { node { inciden
 INCIDENT = '{incidents(incidentId: %(incident)s) { edges { node {emu project \
 repositories { edges { node { name } } } requestSet(kind: "RR") { edges { node \
 { requestId status { name } reviewSet { edges { node { assignedByGroup { name } \
-status { name } } } } } } } packages { edges { node { name } } } } } } }'
+status { name } } } } } } } packages { edges { node { name } } } } } \
+    edges{ node { crd } } } }'
 
 ACTIVE_INC_SCHEMA = {
     "type": "object",
@@ -101,6 +102,9 @@ INCIDENT_SCHEMA = {
                                         },
                                         "requestSet": {
                                             "type": "object",
+                                        },
+                                        "crd": {
+                                            "type": "string",
                                         },
                                     },
                                 },
--- a/openqabot/smeltsync.py
+++ b/openqabot/smeltsync.py
@@ -116,10 +116,7 @@ class SMELTSync:
         incident["approved"] = approved
         incident["rr_number"] = rr_number
         incident["inReviewQAM"] = inReviewQAM
-        # this is temporary solution. At first stage we will just push this field into dashboard
-        # next we will add logic to store it in dashboard and then finally we will replace
-        # here dummy value with real one from smelt
-        incident["embargoed"] = False
+        incident["embargoed"] = bool(inc["crd"])
 
         return incident
 
--- a/openqabot/types/baseconf.py
+++ b/openqabot/types/baseconf.py
@@ -24,3 +24,6 @@ class BaseConf(metaclass=ABCMeta):
     @abstractstaticmethod
     def normalize_repos(config):
         pass
+
+    def filter_embargoed(self) -> bool:
+        return any(k.startswith("PUBLIC") for k in self.settings.keys())
--- a/openqabot/types/incident.py
+++ b/openqabot/types/incident.py
@@ -19,6 +19,7 @@ class Incident:
         self.id = incident["number"]
         self.rrid = f"{self.project}:{self.rr}" if self.rr else None
         self.staging = not incident["inReview"]
+        self.embargoed = incident["embargoed"]
 
         self.channels = [
             Repos(p, v, a)
--- a/tests/test_smeltsync.py
+++ b/tests/test_smeltsync.py
@@ -2,7 +2,6 @@ from collections import namedtuple
 import logging
 import re
 
-import functools
 import pytest
 import responses
 
@@ -58,6 +57,7 @@ def fake_smelt_api(request):
                                     ]
                                 },
                                 "packages": {"edges": [{"node": {"name": "xrdp"}}]},
+                                "crd": request.param[3],
                             }
                         }
                     ]
@@ -90,7 +90,9 @@ def fake_dashboard_replyback():
 @responses.activate
 @pytest.mark.parametrize("fake_qem", [()], indirect=True)
 @pytest.mark.parametrize(
-    "fake_smelt_api", [(["qam-openqa", "new", "review"])], indirect=True
+    "fake_smelt_api",
+    [(["qam-openqa", "new", "review", "2023-01-01 04:31:12"])],
+    indirect=True,
 )
 def test_sync_qam_inreview(fake_qem, caplog, fake_smelt_api, fake_dashboard_replyback):
     caplog.set_level(logging.DEBUG, logger="bot.syncres")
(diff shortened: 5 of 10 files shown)


openqabot/types/incident.py
```diff
@@ -19,6 +19,7 @@ def __init__(self, incident):
         self.id = incident["number"]
         self.rrid = f"{self.project}:{self.rr}" if self.rr else None
         self.staging = not incident["inReview"]
+        self.embargoed = incident["crd"] is not None
```
mimi1vx, line 22: this is wrong :D It will have set CRD after embargo is lifted ...

So it should also compare dater time in CRD with current date time

asmorodskyi: I replied to you in private . resolving this here if I not convinced you please reply me in private too

mimi1vx (changes requested):

openqabot/types/baseconf.py
```diff
@@ -24,3 +24,6 @@ def __call__(
     @abstractstaticmethod
     def normalize_repos(config):
         pass
+
+    def filter_embargoed(self) -> bool:
+        return bool(self.settings.get("PUBLIC_CLOUD_RUN", False))
```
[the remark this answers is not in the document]
mimi1vx: +- here with this `return any( k.startswith("PUBLIC") for k in self.settings.keys())` we got the same effect without a new variable to use in the job call :D

asmorodskyi:
DISCLAIMER : I have 0 confidence in changes what I have done and without real run  ( not dry_run ) I can not prove anything . Tests are passing and I assume everything should correct but please be picky with this one to not solve disaster later

asmorodskyi:
beside variable name raised by @mimi1vx  , ** PR is ready for review **. 
( variable name can be changed within 1 minute after I get suggestion what should be the name ... )

openqabot/smeltsync.py
```diff
@@ -116,10 +116,7 @@ def _create_record(cls, inc):
         incident["approved"] = approved
         incident["rr_number"] = rr_number
         incident["inReviewQAM"] = inReviewQAM
-        # this is temporary solution. At first stage we will just push this field into dashboard
-        # next we will add logic to store it in dashboard and then finally we will replace
-        # here dummy value with real one from smelt
-        incident["embargoed"] = False
+        incident["embargoed"] = inc["crd"][0]["crd"]
```
mimi1vx, line 119: what type is `inc["crd"][0]["crd"]` and what happens if is "crd" missing ?

openqabot/types/baseconf.py
```diff
@@ -1,17 +1,17 @@
 # Copyright SUSE LLC
 # SPDX-License-Identifier: MIT
-from abc import ABCMeta, abstractmethod, abstractstaticmethod
```
mimi1vx, line 3: Why ?

openqabot/types/baseconf.py
```diff
@@ -1,17 +1,17 @@
...
 from typing import Any, Dict, List, Optional
 
 from .incident import Incident
 
 
-class BaseConf(metaclass=ABCMeta):
```
mimi1vx, line 9: dtto

openqabot/types/baseconf.py
```diff
@@ -21,6 +21,9 @@ def __call__(
     ) -> List[Dict[str, Any]]:
         pass
 
-    @abstractstaticmethod
+    @staticmethod
```
mimi1vx, line 24: why this change ? its abstract method which needs to be implemented ..

openqabot/types/baseconf.py
```diff
@@ -1,17 +1,17 @@
...
 # SPDX-License-Identifier: MIT
-from abc import ABCMeta, abstractmethod, abstractstaticmethod
 from typing import Any, Dict, List, Optional
 
 from .incident import Incident
 
 
-class BaseConf(metaclass=ABCMeta):
+class BaseConf:
+    OBEY_EMBARGOED_CONF = "PUBLIC_CLOUD_RUN"
```
mimi1vx, line 9: no , this is pretty awful ...

openqabot/types/baseconf.py
```diff
@@ -21,6 +21,9 @@ def __call__(
     ) -> List[Dict[str, Any]]:
         pass
 
-    @abstractstaticmethod
+    @staticmethod
     def normalize_repos(config):
         pass
+
+    def filter_embargoed(self) -> bool:
+        return bool(self.settings.get(BaseConf.OBEY_EMBARGOED_CONF, False))
```
mimi1vx, line 29: same problem as before, don use `settings` for bot configuration --> `settings` is mean to be passed into openQA.

mimi1vx (changes requested):

openqabot/smeltsync.py
asmorodskyi: 1. crd suppose to be boolean 
2. Regarding what will happen if field will be missing I think same question can be applied to ALL fields in `inc` , right ? Don't see why `crd` field should have special treatment here . All fields which we got here if I understand everything right are defined by `INCIDENT` query in smelt.py . So removal of any field from this query will cause exceptions in this function ( and in many other places ) .

openqabot/types/baseconf.py
asmorodskyi: Because I need to test `is_embargoed` method , for this I need to create instance of this class see test code

openqabot/types/baseconf.py
asmorodskyi: I am closing this comment , it is just duplication of same question as above , don't see the point to copy/pasting same answer two times

openqabot/types/baseconf.py
asmorodskyi: why ? please provide some practical reasoning . such comment is applicable to movie that you have watched or food which you tried . In code review it is not allow me to keep meaningful discussion . you saying "this is awful" and I am saying "no this is cool" and we are blocked ...

From my side I can say that such approach allow to avoid typos in variable name when it is used in several places and ease change of its name

openqabot/types/baseconf.py
asmorodskyi: ok will change that

openqabot/types/baseconf.py
mimi1vx: test can be on Any other class which inherits  from BaseConf ......

openqabot/smeltsync.py
mimi1vx: 1 ... `crd` according docs is `DateTime` and api of dashboard excepts boolean ...

mimi1vx: 1) `crd`  according to docs is `DateTime`  ...

openqabot/types/baseconf.py
asmorodskyi: ok will do that

openqabot/types/baseconf.py
asmorodskyi: resolved

openqabot/types/baseconf.py
asmorodskyi: done

openqabot/loader/smelt.py
```diff
@@ -25,7 +25,8 @@
 INCIDENT = '{incidents(incidentId: %(incident)s) { edges { node {emu project \
 repositories { edges { node { name } } } requestSet(kind: "RR") { edges { node \
 { requestId status { name } reviewSet { edges { node { assignedByGroup { name } \
-status { name } } } } } } } packages { edges { node { name } } } } } } }'
+status { name } } } } } } } packages { edges { node { name } } } } } \
+    edges{ node { crd } } } }'
```
mimi1vx, line 29: please update also `INCIDENT_SCHEMA`

asmorodskyi: can you help me with that ?

openqabot/smeltsync.py
asmorodskyi: during initial discussions I remember statement that we can not rely on this date , because actual agreement saying " until customer get the update first " which normally happening before date defined in crd but not always and in THAT case we should obey actual delivery not this date . 

So agreement which I remember was that **everything** what has this date should be blocked no matter if date there in the past or future . Just the fact of something defined in this field give us true

tests/test_smeltsync.py
```diff
@@ -90,7 +90,7 @@ def reply_callback(request):
 @responses.activate
 @pytest.mark.parametrize("fake_qem", [()], indirect=True)
 @pytest.mark.parametrize(
-    "fake_smelt_api", [(["qam-openqa", "new", "review"])], indirect=True
+    "fake_smelt_api", [(["qam-openqa", "new", "review", True])], indirect=True
```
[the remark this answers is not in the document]
asmorodskyi: fixed

openqabot/loader/smelt.py
asmorodskyi: so I added something , not sure if it , right

mimi1vx (approved):
LGTM

okurz (approved):

PR #128, opened 2023-07-04
Init process of using embargoed field for Incidents

asmorodskyi:
Because of usage of same data structures in two independent applications (qem-bot and qem-dashboard)
we need to do introduction of new field in several stages. At first stage we will start pushing
this field into qem-dashboard from qem-bot. So after deploying this we can than read propely
this field coming from bot

### DIFF
--- a/openqabot/smeltsync.py
+++ b/openqabot/smeltsync.py
@@ -116,6 +116,10 @@ class SMELTSync:
         incident["approved"] = approved
         incident["rr_number"] = rr_number
         incident["inReviewQAM"] = inReviewQAM
+        # this is temporary solution. At first stage we will just push this field into dashboard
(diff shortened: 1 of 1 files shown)


okurz (approved):

Martchus (commented):
Looks like tests still need to be adapted.

kraih:
Very useful test failures, confirm that it actually works as expected.

kraih (changes requested):
Looks like the wrong place. It should be incidents and not openQA jobs.

asmorodskyi:
> Looks like the wrong place. It should be incidents and not openQA jobs.

yes you are correct . Is it right place now ?

kraih (approved):
Does look like the right place.

PR #132, opened 2023-08-10
Rename 'TestBaseConf' to avoid conflict with pytest test discovery

mimi1vx:


perlpunk (approved):
Not sure how the conflict would look like, I guess you have seen a problem somewhere else, but ok

mimi1vx:
@perlpunk 

~~~
tests/test_baseconf.py:7
  /home/mimi/qem-bot/tests/test_baseconf.py:7: PytestCollectionWarning: cannot collect test class 'TestBaseConf' because it has a __init__ constructor (from: tests/test_baseconf.py)
    class TestBaseConf(BaseConf):

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
~~~

pytest test discovery excepts any file starting with `test` and any class starting `Test` as test file/class

Martchus (approved):

PR #133, opened 2023-08-14
Allow single full filename as -c args

mpagot:
`-c` now support both directory with multiple .yml and single .yml filename.


mimi1vx (changes requested):
`read_products` should;d have same functionality as `load_metadata`

mpagot:
> `read_products`

Seems `pc_helper_online.py` too

mimi1vx (approved):

asdil12 (approved):

Martchus (approved):

mpagot:
> Do you have a ticket reference?

No I do not.

PR #134, opened 2023-08-14
Use correct logging in pc_helper

mimi1vx:


asdil12 (approved):

Martchus (approved):

openqabot/pc_helper.py
```diff
@@ -1,12 +1,12 @@
...
 # SPDX-License-Identifier: MIT
 from functools import lru_cache
+from logging import getLogger
 import re
 
 from .utils import retry5 as requests
-from .utils import create_logger
 
-log = create_logger("pc_helper")
+log = getLogger("openqabot.pc_helper")
```
asmorodskyi, line 9: saying that something is "correct" and other thing is not does not helping understanding what kind of problem you actually see here . Can you please elaborate ? 

use of create_logger function everywhere in code give us confidence that every logger in app will be configured in same way ( same StreamHandler and Formatter , same default log level ) Now we falling back to defaults ....

mimi1vx: `create_logger` is used in the root of the app with the name `openqabot`, other use this as a root logger and inherit all settings from this logger.

asmorodskyi: ah so it is just lack of my python knowledge than so if you do `getLogger("openqabot.pc_helper")` you will get Streamer and default log level from "openqabot" logger ? Nice !

asmorodskyi: than probably we can drop `create_logger` function from utils ! ( if above message is true )

PR #135, opened 2023-09-01
Support both .yml and .yaml files

mpagot:
Support both .yml and .yaml files
Move get_yml_list to utils to be able to use it in pc_helper_online too.
Fix some typo and formatting in the documentation.

### DIFF
--- a/openqabot/utils.py
+++ b/openqabot/utils.py
@@ -2,7 +2,8 @@
 # SPDX-License-Identifier: MIT
 import logging
 from copy import deepcopy
-from typing import Optional
+from typing import Optional, List
+from pathlib import Path
 
 from requests import Session
 from requests.adapters import HTTPAdapter
@@ -19,6 +20,15 @@ def create_logger(name: str) -> logging.Logger:
     log.addHandler(handler)
     log.setLevel(logging.INFO)
     return log
+
+
+def get_yml_list(path: Path) -> List[Path]:
+    yml_list = []
+    if path.is_file() and path.match("*.yml"):
+        yml_list.append(path)
+    else:
+        yml_list = [p for p in path.glob("*.yml")]
+    return yml_list
 
 
 def walk(inc):
--- a/openqabot/utils.py
+++ b/openqabot/utils.py
@@ -23,11 +23,18 @@ def create_logger(name: str) -> logging.Logger:
 
 
 def get_yml_list(path: Path) -> List[Path]:
+    """
+    Create a list of YML filenames from a folder
+    or from a single file path.
+    """
     yml_list = []
-    if path.is_file() and path.match("*.yml"):
-        yml_list.append(path)
-    else:
-        yml_list = [p for p in path.glob("*.yml")]
+    extensions = ("yml", "yaml")
+    for ext in extensions:
+        if path.is_file():
+            if path.name.endswith("." + ext):
+                yml_list.append(path)
+        elif path.is_dir():
+            yml_list += list(path.glob("*." + ext))
     return yml_list
 
 
--- a/doc/config.md
+++ b/doc/config.md
@@ -4,14 +4,15 @@ Configuration, in another way also called metadata of qem-bot, is held in yaml f
 Configuration usually spans over multiple files. The default location accepted from qem-bot is `/etc/openqabot/`. All configuration files must have `.yml` extension.
 
 ```bash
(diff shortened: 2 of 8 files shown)


asmorodskyi (approved):
LGTM

openqabot/utils.py
```diff
@@ -21,6 +22,20 @@ def create_logger(name: str) -> logging.Logger:
     return log
 
 
+def get_yml_list(path: Path) -> List[Path]:
+    """
+    Create a list of YML filesnames from a folder
```
b10n1k, line 27: just a typo. Sorry s/filesnames/filenames/

openqabot/utils.py
```diff
@@ -21,6 +22,20 @@ def create_logger(name: str) -> logging.Logger:
...
+    or from a single file path.
+    """
+    yml_list = []
+    extension = "*.yml"
+    if path.is_file() and path.match(extension):
```
b10n1k, line 32: Maybe it is just fine but a yaml file can have either `yml` or `yaml`. The second is not gonna ever match.

openqabot/utils.py
```diff
@@ -21,6 +22,20 @@ def create_logger(name: str) -> logging.Logger:
...
+        yml_list.append(path)
+    else:
+        yml_list = [p for p in path.glob(extension)]
```
ricardobranco777, line 35:
```suggestion
        yml_list = list(path.glob(extension))
```

openqabot/utils.py
ricardobranco777, line 32:
```suggestion
    extensions = (".yml", ".yaml")
    if path.is_file() and path.name.endswith(extensions):
```

ricardobranco777 (approved):
Couple of suggestions.  Otherwise LGTM.

openqabot/utils.py
mpagot: seems by design https://github.com/openSUSE/qem-bot/blob/master/doc/config.md?plain=1#L4

```
All configuration files must have `.yml` extension.
```

I like the idea to extend it, eventually in a different PR.

mpagot: I elaborated it a little bit more. `.something` is ok for `.endswith` but it is not for `.glob` that expect a "search pattern" like `*.something` with the `*`

openqabot/utils.py
```diff
@@ -21,6 +22,22 @@ def create_logger(name: str) -> logging.Logger:
...
+    or from a single file path.
+    """
+    yml_list = []
+    extensions = ("yml", "yaml")
+    for ext in extensions:
+        if path.is_file():
+            if path.name.endswith("." + ext):
+                yml_list.append(path)
+        else:
+            yml_list += list(path.glob("*." + ext))
```
ricardobranco777, line 37:
```suggestion
    extensions = (".yml", ".yaml")
    if path.is_file() and path.suffix in extensions:
        yml_list.append(path)
    elif path.is_dir()
        yml_list += [f for f in p.iterdir() if f.suffix in extensions]
```

mpagot: I like to keep the `if` in two separate layers like

```
 if path.is_file():
            if path.name.endswith("." + ext):
```

and not doing 
```
 if path.is_file() and path.name.endswith("." + ext):
```

for a specific use case. It is the one covered by unit test `test_get_yml_list_single_file_not_yml` : th euser provide a single file path (not a folder) but with an invalid extension. Using the `if  ... AND ... :` pattern results in it to be considered a folder.

mpagot: I like to change `else:` to `elif path.is_dir():` 

I'm not sure I like to move away from `.glob` that was the original code also before https://github.com/openSUSE/qem-bot/pull/133

okurz:
@Mergifyio rebase

okurz (changes requested):
Given all the changes I think it's time you update the PR subject and potentially also git commit messages to something more realistically describing what change this will actually bring :)

okurz:
I see now that you have multiple changes in a single commit. If it's not too much trouble please separate the changes into individual commits

mpagot:
> I see now that you have multiple changes in a single commit. If it's not too much trouble please separate the changes into individual commits

Done: 3 commits

okurz:
@Mergifyio refresh

okurz:
@openSUSE/qa-tools-team second approval needed

kalikiana (approved):

PR #140, opened 2023-09-22
Catch KeyError exception on incidents

foursixnine:
Possibly missing from 0cc1b68777b2674fbef532315e6905bdc305d460

this is a naive HailMary attempt here https://progress.opensuse.org/issues/136331


okurz (approved):
Well, no clue but, sure, why not :)

PR #141, opened 2023-09-26
Check also unversioned SLE-12 for existing jobs

mimi1vx:
fixes https://progress.opensuse.org/issues/137006

### DIFF
--- a/openqabot/types/incidents.py
+++ b/openqabot/types/incidents.py
@@ -72,11 +72,21 @@ class Incidents(BaseConf):
                 ):
                     return True
             except KeyError:
-                log.debug(
-                    "Incident %s does not have %s arch in SLE-12 module version"
-                    % (inc.id, arch)
-                )
-                continue
+                try:
+                    if ver.startswith("12"):
+                        if (
+                            job["flavor"] == flavor
+                            and job["arch"] == arch
+                            and job["settings"]["REPOHASH"]
+                            == inc.revisions[ArchVer(arch, "12")]
+                        ):
+                            return True
+                except KeyError:
+                    log.debug(
+                        "Incident %s does not have %s arch in SLE-12 module version"
+                        % (inc.id, arch)
+                    )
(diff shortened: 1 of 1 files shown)


okurz (changes requested):
Please add according test coverage. See https://suse.slack.com/archives/C02CANHLANP/p1695648806182389?thread_ts=1695647246.909829&cid=C02CANHLANP for context

openqabot/types/incidents.py
```diff
@@ -72,11 +72,21 @@ def _is_scheduled_job(
...
+                        if (
+                            job["flavor"] == flavor
+                            and job["arch"] == arch
+                            and job["settings"]["REPOHASH"]
+                            == inc.revisions[ArchVer(arch, "12")]
+                        ):
+                            return True
+                except KeyError:
+                    log.debug(
+                        "Incident %s does not have %s arch in SLE-12 module version"
```
kalikiana, line 86: At least a single sentence in the description explaining what is being addressed would be helpful here.

Also please add a unit test to check this case, since we had two commits namely #140 and #138 that did not fix it.

kalikiana (changes requested):

czerw (approved):
Added logic seems to be correct, but I can't predict if it will not break something else. So from me LGTM.

perlpunk (commented):
Also I'm wondering if we can avoid this nested KeyError exception handling by testing proactively whether a key exists and then handle the conditions in if/elsif blocks.

openqabot/types/incidents.py
kalikiana: The error handling is actually wrong. We have two places where we're checking for an unversioned module and where we check that this is 12. I'm looking to see if we can deduplicate this first.

kalikiana: See #142

kalikiana:
Closeing in favor of #142

PR #142, opened 2023-09-26
Deduplicate handling of unversioned modules

kalikiana:
The logic should be the same to avoid different behavior in the same case.

See: https://progress.opensuse.org/issues/137006

### DIFF
--- a/openqabot/types/incident.py
+++ b/openqabot/types/incident.py
@@ -71,6 +71,18 @@ class Incident:
         self.revisions = self._rev(self.channels, self.project)
         self.livepatch: bool = self._is_livepatch(self.packages)
 
+    def revisions_with_fallback(self, arch: str, ver: str):
+        try:
+            arch_ver = ArchVer(arch, ver)
+            # An unversioned SLE12 module will have ArchVer version "12"
+            # but settings["VERSION"] can be any of "12","12-SP1" ... "12-SP5".
+            if arch_ver not in self.revisions and ver.startswith("12"):
+                arch_ver = ArchVer(arch, "12")
+            return self.revisions[arch_ver]
+        except KeyError:
+            log.debug("Incident %s does not have %s arch in %s" % (self.id, arch, ver))
+            return None
+
     @staticmethod
     def _rev(channels: List[Repos], project: str) -> Dict[ArchVer, int]:
         rev: Dict[ArchVer, int] = {}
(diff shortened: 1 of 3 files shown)


tests/test_incident.py
```diff
@@ -101,3 +101,9 @@ def test_inc_nochannels2(mock_good):
     ]
     with pytest.raises(EmptyChannels):
         Incident(bad_data)
+
+
+def test_inc_revisions(mock_good):
+    incident = Incident(test_data)
+    assert incident.revisions_with_fallback("x86_64", "15-SP4")
+    assert incident.revisions_with_fallback("aarch64", "15-SP4")
```
kalikiana, line 109: I can't figure out how to reproduce the case where we get a SLE12 with no versioning...

openqabot/types/incident.py
```diff
@@ -71,6 +71,17 @@ def __init__(self, incident):
         self.revisions = self._rev(self.channels, self.project)
         self.livepatch: bool = self._is_livepatch(self.packages)
 
+    def revisions_with_fallback(self, arch: str, ver: str):
+        try:
+            # An unversioned SLE12 module will have ArchVer version "12"
+            # but settings["VERSION"] can be any of "12","12-SP1" ... "12-SP5".
+            if ver.startswith("12"):
```
mimi1vx, line 78: this is wrong place , first full version check then with '12'

mimi1vx: something like 

```
if ArchVer(arch,ver) in self.revisions:
     return self.revisions[ArchVer(arch, ver)]
if ver.startswith("12") and ArchVer(arch, "12") in self.revisions:
      return self.revisions[ArchVer(arch, "12")]
return None
```

mimi1vx (changes requested):

tests/test_incident.py
kalikiana: Works now. Typos can be confusing in this context.

perlpunk (approved):

mimi1vx (approved):

openqabot/types/incidents.py
```diff
@@ -140,34 +136,10 @@ def __call__(
...
-                                )
-                                continue
-                        else:
-                            log.debug(
-                                "Incident %s does not have %s arch in %s"
-                                % (inc.id, arch, self.settings["VERSION"])
-                            )
-                            continue
-
+                    revs = inc.revisions_with_fallback(arch, self.settings("VERSION"))
```
mpagot, line 139: Should not be 
```
revs = inc.revisions_with_fallback(arch, self.settings["VERSION"])
```

Square bracket to accessing the **settings** dictionary?

PR #143, opened 2023-09-27
Add Unit test for Incidents

mpagot:
Add few tests to create a base coverage.
Fix a variable access issue in Incidents class probably from https://github.com/openSUSE/qem-bot/pull/142.

### DIFF
--- a/tests/test_incidents.py
+++ b/tests/test_incidents.py
@@ -1,0 +1,183 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: MIT
+
+from openqabot.types.incidents import Incidents
+from unittest import mock
+import pytest
+
+
+def test_incidents_constructor():
+    """
+    What is the bare minimal set of arguments
+    needed by the constructor?
+    """
+    test_config = {}
+    test_config["FLAVOR"] = {}
+    inc = Incidents(product="", settings=None, config=test_config, extrasettings=None)
+
+
+def test_incidents_printable():
+    """
+    Try the printable
+    """
+    test_config = {}
+    test_config["FLAVOR"] = {}
+    inc = Incidents(
+        product="hello", settings=None, config=test_config, extrasettings=None
+    )
+    assert "<Incidents product: hello>" == str(inc)
+
+
+def test_incidents_call():
+    """
+    What is the bare minimal set of arguments
+    needed by the callable?
+    """
+    test_config = {}
+    test_config["FLAVOR"] = {}
+    inc = Incidents(product="", settings=None, config=test_config, extrasettings=None)
+    res = inc(incidents=[], token={}, ci_url="", ignore_onetime=False)
+    assert res == []
+
+
+def test_incidents_call_with_flavors():
+    test_config = {}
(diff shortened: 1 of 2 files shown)


kalikiana (approved):
Nice! Thank you for helping with test coverage.

mpagot:
> Nice! Thank you for helping with test coverage.

yes but some disclaimer :-) 
Both UT added here and https://github.com/openSUSE/qem-bot/pull/139 are "not good"
They are not in the sense that tests in them as not been developed to test a specific behavior but just to reach coverage.

PROS:
   We have more coverage so we can detect python errors in the bot
   
 CONS:
   Detected errors will not be about behavioral changes. There could be false positive errors just because we change internal implementation.

ricardobranco777 (approved):

tests/test_incidents.py
```diff
@@ -0,0 +1,183 @@
...
+from unittest import mock
+import pytest
+
+
+def test_incidents_constructor():
+    """
+    What is the bare minimal set of arguments
+    needed by the constructor?
+    """
+    test_config = {}
```
asmorodskyi, line 14:
```suggestion
    test_config = { "FLAVOR": {}}
```

tests/test_incidents.py
```diff
@@ -0,0 +1,183 @@
...
+    test_config["FLAVOR"] = {}
+    inc = Incidents(product="", settings=None, config=test_config, extrasettings=None)
```
asmorodskyi, line 16:
```suggestion
    inc = Incidents(product="", config=test_config)
```

tests/test_incidents.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+def test_incidents_printable():
+    """
+    Try the printable
+    """
+    test_config = {}
+    test_config["FLAVOR"] = {}
+    inc = Incidents(
```
asmorodskyi, line 25: please create fixture which will return this object

tests/test_incidents.py
```diff
@@ -0,0 +1,183 @@
...
+            return [{"flavor": None}]
+
+    def mock_get(*args, **kwargs):
+        return MockResponse()
+
+    monkeypatch.setattr("openqabot.types.incidents.requests.get", mock_get)
+
+
+def test_incidents_call_with_channels(request_mock):
+    class MyIncident(object):
```
asmorodskyi, line 124: same for MyIncident class - you have a lot of duplicated code here which can be reduced with fixtures

mpagot: I agree for `Incidents` 
I probably disagree for `MyIncident`: each implementation in each test is adding the minimal amount of element needed in the class to reach the point of the Incidents code reached by the test (is it a good way to test it? Maybe not, see my disclaimer https://github.com/openSUSE/qem-bot/pull/143#issuecomment-1738666443 )

I could maybe address this suggestion as an improvement in a next PR

asmorodskyi: sure , it is just "nice to have " so up to you

Martchus (approved):

PR #144, opened 2023-09-28
SLES4SAP PINT query function

mpagot:
MU jobs for SAP and HA tests in the cloud are designed to:
- create a HA or SAP cluster in the cloud
- install the update on it
- perform some functional tests.

The cluster is created using various versions of SLES images from the CSP catalog.
This change is about always providing the openQA POST the name of the latest version of the catalog image.
This information is retrieved from PINT
 
Content of the PR:
 - Add API to PC helper to query could imaged from PINT.
 - Add Unit Testing coverage for that.
 - Add example of usage in pc_helper_online.

Ticket TEAM-8180

### DIFF
--- a/openqabot/pc_helper.py
+++ b/openqabot/pc_helper.py
@@ -11,9 +11,11 @@ log = getLogger("openqabot.pc_helper")
 
 def get_latest_tools_image(query):
     """
-    'publiccloud_tools_<BUILD NUM>.qcow2' is a generic name for an image used by Public Cloud tests to run
-    in openQA. A query is supposed to look like this "https://openqa.suse.de/group_overview/276.json" to get
-    a value for <BUILD NUM>
+    'publiccloud_tools_<BUILD NUM>.qcow2' is a generic name
+    for an image used by Public Cloud tests to run in openQA.
+    A query is supposed to look like this
+    "https://openqa.suse.de/group_overview/276.json"
+    to get a value for <BUILD NUM>
     """
 
     ## Get the first not-failing item
@@ -100,6 +102,133 @@ def apply_publiccloud_pint_image(settings):
     return settings
 
 
+def get_pint_image(name_filter, field, state, query, region=None):
+    """
+    Calculate the most recent image in PINT:
+    1. Compose PINT url using query+state,
+    2. get list of images from the url,
+    3. search the one with name matching the filter,
+    4. return the newest one.
+
+    :param str name_filter passed to get_recent_pint_image
+    :param str field key name in dictionary returned by PINT to be used as ID in the return dictionary
+    :param str state active/inactive used as name of the json file in PINT url
+    :param str query url of PINT, has not to have '/' at the end
+    :param str region optional filter, mostly needed by AWS
+    :return two keys dictionary: NAME and ID or an empty dictionary in case of error
+    """
+    settings = {}
+    url = f"{query}/{state}.json"
+    log.debug("Pint url:%s name_filter:%s region:%s", url, name_filter, region)
+    try:
+        images = pint_query(url)["images"]
+        image = get_recent_pint_image(images, name_filter, region=region, state=state)
+        if image is None:
+            raise ValueError(
+                f"Cannot find matching image in PINT with name:[{name_filter}] and state:{state}"
+            )
+        if field not in image.keys() or "name" not in image.keys():
+            raise ValueError(
+                f"Cannot find expected keys '{field}' in the selected image dictionary {image}"
+            )
+        settings["ID"] = image[field]
+        settings["NAME"] = image["name"]
+    except BaseException as e:
+        log.warning(f"get_pint_image handling failed: {e}")
+    log.debug("Return settings:%s", settings)
+    return settings
+
+
+def pint_url(pint_base_url, csp_name):
+    """
+    :param str pint_base_url, usually https://susepubliccloudinfo.suse.com/v1
+    :param str csp_name microsoft, amazon, google, ...
+    :return str PINT url for the images API
+    """
+    url = f"{pint_base_url}/{csp_name}/images"
+    log.debug("PINT url:%s", url)
+    return url
+
+
+def sles4sap_pint_gce(name_filter, state, pint_base_url):
+    """
+    Query PINT about GCE images and retrieve the latest one
+    """
+    job_settings = {}
+    ret = get_pint_image(
+        name_filter=name_filter,
+        field="project",
+        state=state,
+        query=pint_url(pint_base_url, "google"),
+    )
+    if any(ret):
+        job_settings["PUBLIC_CLOUD_IMAGE_NAME"] = f"{ret['ID']}/{ret['NAME']}"
+        job_settings["PUBLIC_CLOUD_IMAGE_STATE"] = state
+    return job_settings
+
+
+def sles4sap_pint_ec2(name_filter, state, pint_base_url, region_list):
+    """
+    Query PINT about EC2 images and retrieve the latest one
+    for each of the requested regions.
+    Returned data is organized in a different way from sles4sap_pint_azure
+    and sles4sap_pint_gce
+    """
+    job_settings = {}
+    images_list = {}
+    for this_region in region_list:
+        ret = get_pint_image(
+            name_filter=name_filter,
+            field="id",
+            state=state,
+            query=pint_url(pint_base_url, "amazon"),
+            region=this_region,
+        )
+        if any(ret):
+            images_list[this_region] = ret
+    if any(images_list):
+        # All element should have same name, just get the first
+        job_settings["PUBLIC_CLOUD_IMAGE_NAME"] = next(iter(images_list.items()))[1][
+            "NAME"
+        ]
+        job_settings["PUBLIC_CLOUD_IMAGE_STATE"] = state
+        job_settings["SLES4SAP_QESAP_OS_OWNER"] = "aws-marketplace"
+        # Pack all pairs region/AMI in a ';' separated values string
+        setting_regions = []
+        setting_ami = []
+        for image_region, image_settings in images_list.items():
+            setting_regions.append(image_region)
+            setting_ami.append(image_settings["ID"])
+        job_settings["PUBLIC_CLOUD_IMAGE_NAME_REGIONS"] = ";".join(setting_regions)
+        job_settings["PUBLIC_CLOUD_IMAGE_NAME_ID"] = ";".join(setting_ami)
+    return job_settings
+
+
+def apply_sles4sap_pint_image(
+    cloud_provider, pint_base_url, name_filter, region_list=None
+):
+    """
+    Return OS_IMAGE related settings based on the given SLES4SAP_IMAGE_REGEX
+    :param str cloud_provider uppercase cps name: could be one of AZURE|GCE|EC2
+    :param str pint_base_url forwarded to lower layer, usually is https://susepubliccloudinfo.suse.com/v1
+    :param str name_filter forwarded to lower layer to pint_url
+    :param str region_list=None only considered by EC2
+    :return dict Dictionary of OS_IMAGE related settings
+    """
+    job_settings = {}
+    for state in ["active", "inactive"]:
+        if "GCE" in cloud_provider:
+            job_settings = sles4sap_pint_gce(name_filter, state, pint_base_url)
+        elif "EC2" in cloud_provider:
+            job_settings = sles4sap_pint_ec2(
+                name_filter, state, pint_base_url, region_list
+            )
+        if any(job_settings):
+            break
+    log.debug("Sles4sap job settings:%s", job_settings)
+    return job_settings
+
+
 def get_recent_pint_image(images, name_regex, region=None, state="active"):
     """
     From the given set of images (received json from pint),
@@ -109,6 +238,11 @@ def get_recent_pint_image(images, name_regex, region=None, state="active"):
      - state given the state of the image
 
     Get the latest one based on 'publishedon'
+    :param list images field adressed by "images" key in json returned by pint_query
+    :param str name_regex has to be valid regexp, will be passed to re.compile
+    :param str region if provided only considers images for this region
+    :param str state only considers images with this state
+    :return dict selected image
     """
 
     def is_newer(date1, date2):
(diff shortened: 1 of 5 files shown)


openqabot/pc_helper.py
```diff
@@ -100,6 +100,148 @@ def apply_publiccloud_pint_image(settings):
...
+        job_settings["SLES4SAP_QESAP_OS_VER_REGIONS"] = ";".join(setting_regions)
+        job_settings["SLES4SAP_QESAP_OS_VER_ID"] = ";".join(setting_ami)
+    return job_settings
+
+
+def apply_sles4sap_pint_image(
+    cloud_provider, pint_base_url, name_filter, region_list=None
+):
+    """
+    Applies OS_IMAGE relates settings based on the given SLES4SAP_IMAGE_REGEX
```
alvarocarvajald, line 227: s/relates/related/

pc_helper_online.py
```diff
@@ -47,6 +51,23 @@ def main():
...
+                if "SLES4SAP_PINT_QUERY" in settings and "incidents" in data:
+                    for flavor in data["incidents"]["FLAVOR"]:
+                        if (
+                            "pint_name_regexp"
+                            not in data["incidents"]["FLAVOR"][flavor]
+                        ):
+                            continue
+                        cloud_provider = ""
+                        if "azure" in flavor.lower():
+                            cloud_provider = "AZURE"
```
alvarocarvajald, line 63: Why not
```
cloud_provider = flavor.lower()
```
Instead of these 3 lines? Do we need `cloud_provider` to be empty on GCE and EC2?

mpagot: To be more flexible

TL;DR
I'm not even sure that inspecting FLAVOR is a good idea at all. Problem is that we do not have any strongly enforced policy in the sles4sap metadata files about the flavor names for cloud build. Usually we fit in there the cloud provider name, we fit the BYOS/PAYG. So I decided to use this more generic test criteria to be more flexible. Another possible approach is not to use the FLAVOR at all and add an explicit.

alvarocarvajald (approved):
LGTM

lpalovsky (approved):
LGTM

okurz (changes requested):
This is adding quite a lot of changes. Could you please share more context about this, what benefit it brings, what it is required for, etc. The "TEAM-XXX" link does not help as it's not publically available.

@asmorodskyi as an important recent contributor please be aware about this change and potentially review

openqabot/pc_helper.py
```diff
@@ -100,6 +102,152 @@ def apply_publiccloud_pint_image(settings):
...
+    :param str region optional filter, mostly needed by AWS
+    :return Return dictionary has keys NAME and ID
+    """
+    settings = {}
+    try:
+        url = f"{query}{state}.json"
+        log.debug("Pint url:%s name_filter:%s region:%s", url, name_filter, region)
+        images = pint_query(url)["images"]
+        image = get_recent_pint_image(images, name_filter, region=region, state=state)
+        if image is None:
```
asmorodskyi, line 126: mmm will BaseException just catch error which you throwing ? why not just do return settings ?

asmorodskyi:
why do you need PINT for Azure and Google ? In Azure there is nice mechanism of "offer" and "sku" . You can just say Azure "give me latest SLE 15 SP2 image" and Azure will handle what you trying to do here in bot . Something similar about Google . There is good reason why we have such logic only in AWS because only AWS API refusing to "speak" with us using human understandable SLE versions and requires some wicked UUID's which only PINT knows but this is not the case with Azure or Google

mpagot:
> This is adding quite a lot of changes. Could you please share more context about this, what benefit it brings, what it is required for, etc. The "TEAM-XXX" link does not help as it's not publically available.

PR description extended with more context

mpagot:
> why do you need PINT for Azure and Google ? In Azure there is nice mechanism of "offer" and "sku" . You can just say Azure "give me latest SLE 15 SP2 image" and Azure will handle what you trying to do here in bot .

true, I can potentially simplify the PR

mpagot:
> Something similar about Google .

Do you have an example of the syntax to be used?

alvarocarvajald:
@asmorodskyi wrote:
> why do you need PINT for Azure and Google ?\

+1 on the Azure front as we can already reference the latest image with `:latest` in the offer (IIRC), but is there already such a mechanism for Google?

As far as we understood, Google images have to be referenced by project and image name (for example: `sles-12-sp5-sap-v20230718-x86-64`), and the image name contains a date.

openqabot/pc_helper.py
```diff
@@ -100,6 +102,133 @@ def apply_publiccloud_pint_image(settings):
...
+    job_settings = {}
+    ret = get_pint_image(
+        name_filter=name_filter,
+        field="project",
+        state=state,
+        query=pint_url(pint_base_url, "google"),
+    )
+    if any(ret):
+        job_settings["SLES4SAP_QESAP_OS_VER"] = f"{ret['ID']}/{ret['NAME']}"
+        job_settings["SLES4SAP_QESAP_OS_STATE"] = state
```
mpagot, line 166: @asmorodskyi yesterday we always talk about the fact that dumping `state` to a setting is not needed, but PC is dumping it too

https://github.com/openSUSE/qem-bot/pull/144/files#diff-5578fec6c7adac2cb35f299569fc8310e4a4e7d1005ac8d1802d3dc75b20ede1R84

seems nothing on the test side it using `PUBLIC_CLOUD_IMAGE_STATE`. Would you like me to remove `PUBLIC_CLOUD_IMAGE_STATE` too?

openqabot/pc_helper.py
```diff
@@ -100,6 +102,133 @@ def apply_publiccloud_pint_image(settings):
...
+            region=this_region,
+        )
+        if any(ret):
+            images_list[this_region] = ret
+    if any(images_list):
+        # All element should have same name, just get the first
+        job_settings["SLES4SAP_QESAP_OS_VER"] = next(iter(images_list.items()))[1][
+            "NAME"
+        ]
+        job_settings["SLES4SAP_QESAP_OS_VER_STATE"] = state
```
mpagot, line 194: At least I should rename this one from `SLES4SAP_QESAP_OS_VER_STATE` to `SLES4SAP_QESAP_OS_STATE`

openqabot/pc_helper.py
```diff
@@ -100,6 +102,133 @@ def apply_publiccloud_pint_image(settings):
...
+    """
+    job_settings = {}
+    ret = get_pint_image(
+        name_filter=name_filter,
+        field="project",
+        state=state,
+        query=pint_url(pint_base_url, "google"),
+    )
+    if any(ret):
+        job_settings["SLES4SAP_QESAP_OS_VER"] = f"{ret['ID']}/{ret['NAME']}"
```
mpagot, line 165: I can potentially rename this one to `PUBLIC_CLOUD_IMAGE_NAME`. I'd prefer to do it later as having a dedicated name could help me to debug the final result looking at the triggered job settings.

openqabot/pc_helper.py
```diff
@@ -100,6 +102,131 @@ def apply_publiccloud_pint_image(settings):
...
+            raise ValueError(
+                f"Cannot find matching image in PINT with name:[{name_filter}] and state:{state}"
+            )
+        if field not in image.keys() or "name" not in image.keys():
+            raise ValueError(
+                f"Cannot find expected keys '{field}' in the selected image dictionary {image}"
+            )
+        settings["ID"] = image[field]
+        settings["NAME"] = image["name"]
+    except BaseException as e:
```
ricardobranco777, line 136:
```suggestion
    except BaseException as e:
```
```suggestion
    except Exception as e:
```

openqabot/pc_helper.py
```diff
@@ -100,6 +102,131 @@ def apply_publiccloud_pint_image(settings):
...
+    return settings
+
+
+def pint_url(pint_base_url, csp_name):
+    """
+    :param str pint_base_url, usually https://susepubliccloudinfo.suse.com/v1
+    :param str csp_name microsoft, amazon, google, ...
+    :return str PINT url for the images API
+    """
+    url = "/".join([pint_base_url] + [csp_name, "images"]) + "/"
```
ricardobranco777, line 148:
```suggestion
    url = f"{pint_base_url}/{csp_name}/images/"
```

openqabot/pc_helper.py
```diff
@@ -100,6 +102,131 @@ def apply_publiccloud_pint_image(settings):
...
+    Query PINT about GCE images and retrieve the latest one
+    """
+    job_settings = {}
+    ret = get_pint_image(
+        name_filter=name_filter,
+        field="project",
+        state=state,
+        query=pint_url(pint_base_url, "google"),
+    )
+    if any(ret):
```
ricardobranco777, line 164: Why use `any` here?

openqabot/pc_helper.py
```diff
@@ -100,6 +102,131 @@ def apply_publiccloud_pint_image(settings):
...
+    images_list = {}
+    for this_region in region_list:
+        ret = get_pint_image(
+            name_filter=name_filter,
+            field="id",
+            state=state,
+            query=pint_url(pint_base_url, "amazon"),
+            region=this_region,
+        )
+        if any(ret):
```
ricardobranco777, line 186: Why use any() on a dict?

openqabot/pc_helper.py
pdostal: I don't understand how this can benefit. I only see yet another unnecessary variable.

openqabot/pc_helper.py
asmorodskyi: no we exposing it for informational purposes . So for example when you see job for SLE 15 SP5 which running with image in "deprecated" state you suppose to figure out why this is happening because SP5 which is latest currently suppose to have "active" images , so there is issue or with bot or with image publishing pipeline

pdostal (changes requested):
I don't see a single point why you do this.

All of this logic is already done. You just create your own metadata files.
If there is something different then just alter existing code please.

openqabot/pc_helper.py
asmorodskyi: why can't you use `PUBLIC_CLOUD_IMAGE_STATE` ?

openqabot/pc_helper.py
mpagot: in order to return to the user a more meaningful error message

a-kpappas:
@pdostal How would you 
>  provide the openQA POST the name of the latest version of the catalog image.
> This information is retrieved from PINT
?

openqabot/pc_helper.py
mpagot: renaming is one of the suggestion that I get from Anton yesterday

a-kpappas:
I added non-qesap members as stakeholders to the ticket. You should have visibility now.

openqabot/pc_helper.py
mpagot: is it `state` needed by the test code? I'm not able to find any occurrence of `PUBLIC_CLOUD_IMAGE_STATE` in https://github.com/os-autoinst/os-autoinst-distri-opensuse

openqabot/pc_helper.py
mpagot: I like the idea, I'll do it for sles4sap and reusing the `PUBLIC_CLOUD_IMAGE_STATE` name

pdostal:
So from my understanding this pull request does the following:
 * Move control over which region to test level:
   * Why? - It can be done in metadata.
   * In metadata every flavour just needs to be in separate file but it's anyways quite practical and easy to maintain.
 * Return multiple AMIs of multiple regions so each test suite can run in different region:
   * This is doable but needs to be better documented and of course paired with pull request to os-autoinst-distri-opensuse.

openqabot/pc_helper.py
asmorodskyi: ok

asmorodskyi:
> As far as we understood, Google images have to be referenced by project and image name (for example: `sles-12-sp5-sap-v20230718-x86-64`), and the image name contains a date.
> Yes, an example would help a lot.

sorry to confuse you , you totally right here , for Google we also using pint to figure out exact image name because it contains date which is always changing ...

openqabot/pc_helper.py
m-dati: I'd suggest to remove last **`'/'`**, in order to use then:
`url/<something else>`
instead of:
`ulr<something else> `less clear.

openqabot/pc_helper.py
mpagot: @ricardobranco777  could you please elaborate about why `Exception` is better?

openqabot/pc_helper.py
mpagot: done https://github.com/openSUSE/qem-bot/compare/f086781efbf8ac1d8e0528b93812a4d16e801136..9693a7e55323377acd5daa10a80b943dc07f48f5

pc_helper_online.py
```diff
@@ -47,6 +51,23 @@ def main():
                     apply_publiccloud_pint_image(settings)
                     if "PUBLIC_CLOUD_IMAGE_ID" not in settings:
                         log.error(f"Failed to get PUBLIC_CLOUD_IMAGE_ID from {data}")
+                if "SLES4SAP_PINT_QUERY" in settings and "incidents" in data:
+                    for flavor in data["incidents"]["FLAVOR"]:
+                        if (
+                            "pint_name_regexp"
```
mpagot, line 57: @asmorodskyi  @pdostal this is only partially implemented but relevant clue about how next PR will handle metadata. As reference here a metadata example:

```
 product: SAP-15-SP5
   2   │ settings:
   3   │   DISTRI: sle
   4   │   VERSION: '15-SP5'
   5 + │   SLES4SAP_PINT_QUERY: 'https://susepubliccloudinfo.suse.com/v1'
   6   │ aggregate:
...
  18   │ incidents:
  19   │   FLAVOR:
...
  65   │     Azure-SAP-BYOS-Incidents-saptune:
  66   │       aggregate_job: false
  67   │       archs:
  68   │         - x86_64
  69   │       issues: *test_issues_list
  70   │       required_issues: *required_list
  71   │       packages: *whitelist_saptune
  72 + │       pint_name_regexp: '.*sap-15-sp5-byos'

```

pdostal: Why `SLES4SAP_PINT_QUERY` and not `PUBLIC_CLOUD_PINT_QUERY`?

mpagot: In order to call `apply_sles4sap_pint_image` in place of `apply_publiccloud_pint_image`

PR #145, opened 2023-10-10
flavor_settings field within each Incidents flavor

mpagot:
Add support for a new Incidents configuration field that allow to
specify flavor specific settings.

Keep separated settings for each flavor loop

Ticket: TEAM-8180

### DIFF
--- a/openqabot/types/incidents.py
+++ b/openqabot/types/incidents.py
@@ -102,6 +102,15 @@ class Incidents(BaseConf):
                     full_post["openqa"] = {}
                     full_post["openqa"].update(self.settings)
                     if "flavor_settings" in data:
+                        if any(
+                            forbidden_key in data["flavor_settings"]
+                            for forbidden_key in ["DISTRI", "VERSION"]
+                        ):
+                            log.error(
+                                "flavor:%s ignored as DISTRI and VERSION not allowed in flavor_settings",
+                                flavor,
+                            )
+                            continue
                         full_post["openqa"].update(data["flavor_settings"])
 
                     full_post["qem"]["incident"] = inc.id
--- a/openqabot/types/incidents.py
+++ b/openqabot/types/incidents.py
@@ -96,12 +96,30 @@ class Incidents(BaseConf):
                             inc.id,
                         )
                         continue
+                    full_post: Dict[str, Any] = {}
+                    full_post["api"] = "api/incident_settings"
+                    full_post["qem"] = {}
+                    full_post["openqa"] = {}
+                    full_post["openqa"].update(self.settings)
+                    if "flavor_settings" in data:
+                        full_post["openqa"].update(data["flavor_settings"])
 
+                    full_post["qem"]["incident"] = inc.id
+                    full_post["openqa"]["ARCH"] = arch
+                    full_post["qem"]["arch"] = arch
+                    full_post["openqa"]["FLAVOR"] = flavor
+                    full_post["qem"]["flavor"] = flavor
+                    full_post["openqa"]["VERSION"] = self.settings["VERSION"]
+                    full_post["qem"]["version"] = self.settings["VERSION"]
+                    full_post["openqa"]["DISTRI"] = self.settings["DISTRI"]
+                    full_post["openqa"]["_ONLY_OBSOLETE_SAME_BUILD"] = "1"
+                    full_post["openqa"]["_OBSOLETE"] = "1"
+                    full_post["openqa"]["INCIDENT_ID"] = inc.id
+
+                    if ci_url:
+                        full_post["openqa"]["__CI_JOB_URL"] = ci_url
                     if inc.staging:
                         continue
-
-                    if "flavor_settings" in data:
-                        self.settings.update(data["flavor_settings"])
 
                     if "packages" in data:
                         if not inc.contains_package(data["packages"]):
@@ -111,22 +129,6 @@ class Incidents(BaseConf):
                         if inc.contains_package(data["excluded_packages"]):
                             continue
 
-                    full_post: Dict[str, Any] = {}
-                    full_post["api"] = "api/incident_settings"
-                    full_post["qem"] = {}
-                    full_post["openqa"] = {}
-                    full_post["qem"]["incident"] = inc.id
-                    full_post["openqa"]["ARCH"] = arch
-                    full_post["qem"]["arch"] = arch
-                    full_post["openqa"]["FLAVOR"] = flavor
-                    full_post["qem"]["flavor"] = flavor
-                    full_post["openqa"]["_ONLY_OBSOLETE_SAME_BUILD"] = "1"
-                    full_post["openqa"]["_OBSOLETE"] = "1"
-                    full_post["openqa"]["INCIDENT_ID"] = inc.id
-
-                    if ci_url:
-                        full_post["openqa"]["__CI_JOB_URL"] = ci_url
-
                     if inc.livepatch:
                         full_post["openqa"]["KGRAFT"] = "1"
 
@@ -135,11 +137,6 @@ class Incidents(BaseConf):
                     if inc.rrid:
                         full_post["openqa"]["RRID"] = inc.rrid
 
-                    # Settings for the POST command
-                    full_post["openqa"].update(self.settings)
-                    full_post["openqa"]["VERSION"] = self.settings["VERSION"]
-                    full_post["qem"]["version"] = self.settings["VERSION"]
-                    full_post["openqa"]["DISTRI"] = self.settings["DISTRI"]
                     # old bot used variable "REPO_ID"
                     revs = inc.revisions_with_fallback(arch, self.settings["VERSION"])
                     if not revs:
--- a/openqabot/types/incidents.py
+++ b/openqabot/types/incidents.py
@@ -96,27 +96,12 @@ class Incidents(BaseConf):
                             inc.id,
                         )
                         continue
-                    full_post: Dict[str, Any] = {}
-                    full_post["api"] = "api/incident_settings"
-                    full_post["qem"] = {}
-                    full_post["openqa"] = {}
-                    full_post["openqa"].update(self.settings)
-                    full_post["qem"]["incident"] = inc.id
-                    full_post["openqa"]["ARCH"] = arch
-                    full_post["qem"]["arch"] = arch
-                    full_post["openqa"]["FLAVOR"] = flavor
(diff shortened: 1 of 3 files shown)


openqabot/types/incidents.py
```diff
@@ -96,35 +96,37 @@ def __call__(
                             inc.id,
                         )
                         continue
+
+                    if inc.staging:
+                        continue
+
+                    flavor_settings = self.settings.copy()
```
mpagot, line 103: We can potentially further optimize the code in regard of this copy and the one in 
https://github.com/openSUSE/qem-bot/pull/145/files#diff-81a88cee51bd2524a7e938a6b6d53d79f526ffb436071ead2d3d83163b38b1ecR255

openqabot/types/incidents.py
```diff
@@ -134,8 +136,13 @@ def __call__(
                     if inc.rrid:
                         full_post["openqa"]["RRID"] = inc.rrid
 
+                    # Settings for the POST command
+                    full_post["openqa"].update(flavor_settings)
+                    full_post["openqa"]["VERSION"] = flavor_settings["VERSION"]
```
ricardobranco777, line 141:
```suggestion
                    # Settings for the POST command
                    full_post["openqa"].update(flavor_settings)
```

The assignment after the `.update()` is redundant.

openqabot/types/incidents.py
```diff
@@ -134,8 +136,13 @@ def __call__(
...
+                    full_post["qem"]["version"] = flavor_settings["VERSION"]
+                    full_post["openqa"]["DISTRI"] = flavor_settings["DISTRI"]
```
ricardobranco777, line 143:
```suggestion
```

Also redundant.

ricardobranco777 (changes requested):

Martchus:
The first commit message should likely start with "Add …". If there's a related ticket then it would also be good to reference it in the commit message bodies.

mpagot:
> The first commit message should likely start with "Add …". If there's a related ticket then it would also be good to reference it in the commit message bodies.

PR has intentionally two commits. There's a bug in the first one that I fixed in the second one, but I like, at least reviewer, to be aware about it.

Bug is potentially also related to a possible further optimization in the code in regard of

```
settings = full_post["openqa"].copy()
```

Please have a look and let me know what do you think.

I think that this double commit is more interesting from the point of view of the review. So maybe I can squash them at the moment of merge.  What do you think?

okurz:
You don't need to squash commits if they provide a reasonable story each by themselves. We commonly follow http://chris.beams.io/posts/git-commit/ so please take a look into that blog article if you don't know it yet. At the very least rephrase the git commit message subjects to start in imperative mood starting with capital letter as Martchus stated.

openqabot/types/incidents.py
```diff
@@ -96,35 +96,37 @@ def __call__(
...
                         continue
+
+                    if inc.staging:
+                        continue
+
+                    flavor_settings = self.settings.copy()
+                    if "flavor_settings" in data:
+                        flavor_settings.update(data["flavor_settings"])
+
+                    if "packages" in data:
```
asmorodskyi, line 107: how this related to introduction of `flavor_settings` ?

openqabot/types/incidents.py
```diff
@@ -96,35 +96,37 @@ def __call__(
...
+                        if not inc.contains_package(data["packages"]):
+                            continue
+
+                    if "excluded_packages" in data:
```
asmorodskyi, line 111: how this is related to introduction of flavor_settings ?

openqabot/types/incidents.py
```diff
@@ -96,35 +96,37 @@ def __call__(
...
                     full_post["api"] = "api/incident_settings"
                     full_post["qem"] = {}
                     full_post["openqa"] = {}
-                    full_post["openqa"].update(self.settings)
                     full_post["qem"]["incident"] = inc.id
                     full_post["openqa"]["ARCH"] = arch
                     full_post["qem"]["arch"] = arch
                     full_post["openqa"]["FLAVOR"] = flavor
                     full_post["qem"]["flavor"] = flavor
-                    full_post["openqa"]["VERSION"] = self.settings["VERSION"]
```
asmorodskyi, line 109: disagree , flavor specific settings should NOT allow to overwrite VERSION defined on product level , this will open gate for so many tricky bugs and data inconsistencies that we do NOT want to deal with

openqabot/types/incidents.py
```diff
@@ -96,35 +96,37 @@ def __call__(
...
+
+                    if "excluded_packages" in data:
+                        if inc.contains_package(data["excluded_packages"]):
+                            continue
+
                     full_post: Dict[str, Any] = {}
                     full_post["api"] = "api/incident_settings"
                     full_post["qem"] = {}
                     full_post["openqa"] = {}
-                    full_post["openqa"].update(self.settings)
```
asmorodskyi, line 103: IMHO all changes what you have done above and below may be replaced by just TWO lines added here ( and we NOT removing this one too ofc. ) 

```
 if "flavor_settings" in data:
                 full_post["openqa"].update(data["flavor_settings"])
```

to be on safe side we may add verification something like : 
```
if "VERSION" in data["flavor_settings"] or "DISTRI" in data["flavor_settings"]:
                    raise ValueError("Attempt to redifine VERSION/DISTRI on flavor level")
```

mpagot: I'm ok if it pass the UT I created. I'm giving it a try

mpagot:
> The first commit message should likely start with "Add …".

Done

openqabot/types/incidents.py
mpagot: > The assignment after the .update() is redundant.

Was there even before this PR. I'll address it in a next PR

openqabot/types/incidents.py
mpagot: > Also redundant.

Thanks, good catch but was there even before this PR. I'll address it in a next PR

openqabot/types/incidents.py
mpagot: This code has just be moved not added. 

As `flavor_settings` is updating `settings` I reorganized the code to have all stuff with different purpose as separated and sequentially ordered as possible. I can recognize at least 3 main category.

* check something and skip to schedule a POST, me moving https://github.com/openSUSE/qem-bot/pull/145/files#diff-81a88cee51bd2524a7e938a6b6d53d79f526ffb436071ead2d3d83163b38b1ecR100 is in this category  Any stuff belonging to this category can goes first
* settings related or dependent stuff
* everything else

Generally speaking this method seems to me quite long that does not help about understanding what is going one.

asmorodskyi: any chance to do it in separate PR ?  we introducing quite big functional change here . IMHO it would be nice not to mix this with ANY other changes , so in case something will go wrong we can more easily detect the problem . If you will fully agree that this code may and should be improved . I just don't like to mix this **refactoring** together with introduction of `flavor_settings`

asmorodskyi (approved):
LGTM

openqabot/types/incidents.py
mpagot: @asmorodskyi can we please further elaborate idea in
```
if "VERSION" in data["flavor_settings"] or "DISTRI" in data["flavor_settings"]:
                    raise ValueError("Attempt to redifine VERSION/DISTRI on flavor level")
```

There's no other `raise` in that function. Should we do it and stop everything else, every other `flavors`? Can we just `continue` and skip scheduling  `POST` for the "malicious flavor"? Can we simple remove any `DISTRI` and `VERSION` from `flavor_settings` before to merge it to the `full_post["openqa"]` ?

openqabot/types/incidents.py
```diff
@@ -101,6 +101,14 @@ def __call__(
                     full_post["qem"] = {}
                     full_post["openqa"] = {}
                     full_post["openqa"].update(self.settings)
+                    if "flavor_settings" in data:
+                        if any(
+                            forbidden_key in data["flavor_settings"]
+                            for forbidden_key in ["DISTRI", "VERSION"]
+                        ):
+                            continue
```
asmorodskyi, line 109: I would add some message here

alvarocarvajald (approved):
LGTM

asmorodskyi (approved):
LGTM

baierjan:
Just out of curiosity, is the usage somewhat similar to the `params_expand` parameter?

mpagot:
> `params_expand`

:face_palm: seems mostly or exactly the same

`param_expand` is there from the first commit in github, it has been cited in an example of the configure.md but nothing more. There's no unit testing for that. Apart from that, looking at the code, seems also to me that is capable to do exactly the same of this PR is proposing for `flavor_settings`. One missing feature is possibly the one to filter `DISTRI` and `VERSION`

mpagot:
can we consider to merge this one and try to unify the two in a next PR?

baierjan (approved):
It's OK I guess.

openqabot/types/incidents.py
```diff
@@ -236,6 +236,15 @@ def __call__(
 
                     # add custom vars to job settings
                     if "params_expand" in data:
+                        if any(
```
okurz, line 239: the indendation level seems to become excessive here. Could you please try to extract this and the neighboring code parts into a function so that the level of indendation can be reduced here?

okurz (changes requested):

openqabot/types/incidents.py
mpagot: `black` does not blame me about it

alvarocarvajald: @okurz can you unblock this merge? There already 2 approvals and @mpagot will address your suggestion in an upcoming PR (we're not even close to being done with this task)

okurz: > But it means a big change, big just from the point of view of how many lines of code are moved around.
> I propose to do it in a next PR.
> How it sound?

Sorry, but this should be done as part of this PR. Make it a separate commit, sure, but still part of this PR. Unfortunately I know how often it happens that after a change someone has no time to work on promised changes anymore and then the codebase deteriorates and invites for more mistakes by others. Maybe it wouldn't happen here but that's not an unusual experience. So please look into this before merging.

okurz: Nevermind. As baierjan stated "It's OK I guess." :)

okurz (approved):

PR #146, opened 2023-10-12
Give openQA tests a chance to finish with depriorization instead of obsoletion

okurz:
This gives older builds a chance to finish while new ones are being triggered
regardless of the reason. The approach taken here should still allow for
externally provided settings to override the new default of
"_DEPRIORITIZEBUILD" with the "_OBSOLETE" switch as required. See
http://open.qa/docs/#_spawning_multiple_jobs_based_on_templates_isos_post for
more details about the scheduling flags.

The only test I conducted so far is executing all unit tests which pass.
I did not check if that settings actually end up in the scheduled product
nor if overriding still works.

Related progress issue: https://progress.opensuse.org/issues/94606

### DIFF
--- a/openqabot/types/incidents.py
+++ b/openqabot/types/incidents.py
@@ -101,6 +101,7 @@ class Incidents(BaseConf):
                     full_post["qem"] = {}
                     full_post["openqa"] = {}
                     full_post["openqa"].update(self.settings)
+                    self.set_obsoletion(full_post["openqa"])
                     full_post["qem"]["incident"] = inc.id
                     full_post["openqa"]["ARCH"] = arch
                     full_post["qem"]["arch"] = arch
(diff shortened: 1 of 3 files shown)


openqabot/types/incidents.py
```diff
@@ -109,8 +110,6 @@ def __call__(
                     full_post["openqa"]["VERSION"] = self.settings["VERSION"]
                     full_post["qem"]["version"] = self.settings["VERSION"]
                     full_post["openqa"]["DISTRI"] = self.settings["DISTRI"]
-                    full_post["openqa"]["_ONLY_OBSOLETE_SAME_BUILD"] = "1"
```
baierjan, line 112: We might still want to set this `if "_OBSOLETE" in settings`?

okurz: oh geez, I removed too much. Thx. Reverted

baierjan: Also, the documentation is a little bit puzzling:
> _ONLY_OBSOLETE_SAME_BUILD | Only obsolete (or **deprioritize**) jobs for the same BUILD.

So it looks like it could be used with `_DEPRIORITIZEBUILD`, but on the other hand:

> Needs _OBSOLETE.

baierjan, line 112: My browser might be misbehaving, but I still see this line as removed. Is that correct?

baierjan (changes requested):

openqabot/types/incidents.py
okurz: I trust your browser more than me :)

okurz: I trust your browser more than me :)

baierjan (approved):
OK, let's give it a try.

PR #147, opened 2023-10-17
Add linting support

mpagot:
Include flake8 and pylint as dev dependency.
Add configuration file for pylint and flake8, for the moment all
warnings currently present in the HEAD are disabled.
Add lint targets to the make project.
This commit also result in the github action to run the linter.
Fix simplest warnings.

Verification: https://github.com/openSUSE/qem-bot/actions/runs/6553300955/job/17798390468?pr=147#step:5:1

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,8 @@ jobs:
           cache: 'pip'
       - name: Setup enviroment
         run: pip install -r requirements-dev.txt
+      - name: Linting
+        run: make lint
       - name: Run tests + cov report
         run: make test-with-coverage
       - name: Check style
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,8 +13,6 @@ jobs:
           cache: 'pip'
       - name: Setup enviroment
         run: pip install -r requirements-dev.txt
-      - name: Linting
-        run: make lint
       - name: Run tests + cov report
         run: make test-with-coverage
       - name: Check style
--- a/Makefile
+++ b/Makefile
@@ -5,8 +5,8 @@ all:
 only-test:
 	python3 -m pytest
 
-.PHONY: checkstyle
-checkstyle:
+.PHONY: black
+black:
 	black --check ./
 
 .PHONY: tidy
@@ -21,12 +21,17 @@ pylint:
 flake8:
 	flake8 ./openqabot pc_helper_online.py --config=setup.cfg
 
-.PHONY: lint
-lint: pylint flake8
-
 .PHONY: test-with-coverage
 test-with-coverage:
 	python3 -m pytest -v --cov=./openqabot --cov-report=xml --cov-report=term
 
+# aggregate targets
+
+.PHONY: lint
+lint: pylint flake8
+
+.PHONY: checkstyle
+checkstyle: black lint
+
 .PHONY: test
-test: only-test checkstyle lint
+test: only-test checkstyle
--- a/Makefile
+++ b/Makefile
@@ -13,9 +13,20 @@ checkstyle:
 tidy:
 	black ./
 
+.PHONY: pylint
+pylint:
+	pylint ./openqabot pc_helper_online.py --rcfile=pylintrc
+
+.PHONY: flake8
+flake8:
+	flake8 ./openqabot pc_helper_online.py --config=setup.cfg
+
+.PHONY: lint
+lint: pylint flake8
+
 .PHONY: test-with-coverage
 test-with-coverage:
 	python3 -m pytest -v --cov=./openqabot --cov-report=xml --cov-report=term
 
 .PHONY: test
-test: only-test checkstyle
+test: only-test checkstyle lint
--- a/openqabot/approver.py
+++ b/openqabot/approver.py
@@ -35,20 +35,25 @@ def _mi2str(inc: IncReq) -> str:
 def _handle_http_error(e: HTTPError, inc: IncReq) -> bool:
     if e.code == 403:
         log.info(
-            "Received '%s'. Request %s likely already approved, ignoring"
-            % (e.reason, inc.req)
+            "Received '%s'. Request %s likely already approved, ignoring",
+            e.reason,
+            inc.req,
         )
         return True
     elif e.code == 404:
         log.info(
-            "Received '%s'. Request %s removed or problem on OBS side: %s"
-            % (e.reason, inc.req, e.read().decode())
+            "Received '%s'. Request %s removed or problem on OBS side: %s",
+            e.reason,
+            inc.req,
+            e.read().decode(),
         )
         return False
     else:
         log.error(
-            "Received error %s, reason: '%s' for Request %s - problem on OBS side"
-            % (e.code, e.reason, inc.req)
+            "Received error %s, reason: '%s' for Request %s - problem on OBS side",
+            e.code,
+            e.reason,
+            inc.req,
         )
         return False
 
@@ -74,7 +79,7 @@ class Approver:
 
         log.info("Incidents to approve:")
         for inc in incidents_to_approve:
-            log.info("* %s" % _mi2str(inc))
+            log.info("* %s", _mi2str(inc))
 
         if not self.dry:
             osc.conf.get_config(override_apiurl=OBS_URL)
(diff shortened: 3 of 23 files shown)


openqabot/approver.py
```diff
@@ -35,22 +35,26 @@ def _mi2str(inc: IncReq) -> str:
...
         log.info(
-            "Received '%s'. Request %s likely already approved, ignoring"
-            % (e.reason, inc.req)
+            "Received '%s'. Request %s likely already approved, ignoring",
+            e.reason,
+            inc.req,
         )
         return True
-    elif e.code == 404:
+    if e.code == 404:
```
mpagot, line 43: [if/else] To be carefully reviewed

openqabot/loader/qem.py
```diff
@@ -236,15 +238,14 @@ def update_incidents(token: Dict[str, str], data, **kwargs) -> int:
         except Exception as e:
             log.exception(e)
             return 1
+        if ret.status_code == 200:
```
mpagot, line 241: [if/else] To be carefully reviewed

openqabot/openqa.py
```diff
@@ -84,7 +81,7 @@ def get_job_comments(self, job_id: int):
             )
             ret = list(map(lambda c: {"text": c.get("text", "")}, ret))
         except Exception as e:
-            (method, url, status_code, *other) = e.args
+            (_, _, status_code, *_) = e.args
```
mpagot, line 84: not sure about this as code is less readable

openqabot/openqabot.py
```diff
@@ -61,19 +62,18 @@ def __call__(self):
...
-            log.info("Triggering %d products in openQA" % len(post))
+            log.info("Triggering %d products in openQA", len(post))
             for job in post:
-                log.info("Triggering %s" % str(job))
+                log.info("Triggering %s", str(job))
                 try:
                     self.post_openqa(job["openqa"])
                 except PostOpenQAError:
                     log.info("POST failed, not updating dashboard")
-                    pass
```
mpagot, line 76: please check, seems to me not useful

openqabot/smeltsync.py
```diff
@@ -35,35 +35,25 @@ def __call__(self) -> int:
...
+    def _review_rrequest(request_set):
+        if not request_set:
             return None
-        else:
-            rr = sorted(requestSet, key=itemgetter("requestId"), reverse=True)[0]
-            if rr["status"]["name"] in ("new", "review", "accepted", "revoked"):
-                return rr
-            else:
-                return None
+        rr = sorted(request_set, key=itemgetter("requestId"), reverse=True)[0]
```
mpagot, line 41: [if/else] To be carefully reviewed

openqabot/smeltsync.py
```diff
@@ -35,35 +35,25 @@ def __call__(self) -> int:
...
     @staticmethod
     def _is_inreview(rr_number) -> bool:
         if rr_number["reviewSet"]:
-            if rr_number["status"]["name"] == "review":
-                return True
-            else:
-                return False
-        else:
-            return False
+            return rr_number["status"]["name"] == "review"
```
mpagot, line 49: [if/else] To be carefully reviewed

openqabot/smeltsync.py
```diff
@@ -35,35 +35,25 @@ def __call__(self) -> int:
...
     @staticmethod
     def _is_revoked(rr_number) -> bool:
         if rr_number["reviewSet"]:
-            if rr_number["status"]["name"] == "revoked":
-                return True
-            else:
-                return False
-        else:
-            return False
+            return rr_number["status"]["name"] == "revoked"
```
mpagot, line 55: [if/else] To be carefully reviewed

openqabot/utils.py
```diff
@@ -46,7 +46,7 @@ def walk(inc):
         if len(inc) == 1:
             if "edges" in inc:
                 return walk(inc["edges"])
-            elif "node" in inc:
+            if "node" in inc:
```
mpagot, line 49: [if/else] To be carefully reviewed

Makefile
```diff
@@ -13,9 +13,20 @@ checkstyle:
...
+.PHONY: pylint
+pylint:
+	pylint ./openqabot pc_helper_online.py --rcfile=pylintrc
+
+.PHONY: flake8
+flake8:
+	flake8 ./openqabot pc_helper_online.py --config=setup.cfg
+
+.PHONY: lint
+lint: pylint flake8
```
okurz, line 25: I suggest to combine all lint+flake8+black into checkstyle, so something like

```
black: …
pylint: …
flake8: …
checkstyle: black pylint flake8

File: openqabot/approver.py (hunk 1 above)
Comment (okurz) [reply openqabot/approver.py:43]: that's fine because the previous if concludes with a return so the code execution would never execute the above branch and reach here. So the change makes sense

File: openqabot/loader/qem.py (hunk 2 above)
Comment (okurz) [reply openqabot/loader/qem.py:241]: same as above, change is fine

File: openqabot/openqa.py (hunk 3 above)
Comment (okurz) [reply openqabot/openqa.py:84]: I agree with the change. It's a good practice to explicitly mark unused return variables so the change is good.

File: openqabot/openqabot.py (hunk 4 above)
Comment (okurz) [reply openqabot/openqabot.py:76]: change is good. `pass` should be used if there is no other execution in branch but with the log call there is, so no need for the pass.

File: openqabot/smeltsync.py (hunk 5 above)
Comment (okurz) [reply openqabot/smeltsync.py:41]: same as above, change is fine

File: openqabot/smeltsync.py (hunk 6 above)
Comment (okurz) [reply openqabot/smeltsync.py:49]: change is fine. an equality check directly returns a boolean result so it's equivalent to returning True/False and the change is good.

File: openqabot/smeltsync.py (hunk 7 above)
Comment (okurz) [reply openqabot/smeltsync.py:55]: same as above, change is fine

File: openqabot/smeltsync.py [hunk 10]
```diff
@@ -72,8 +62,7 @@ def _is_accepted(rr_number) -> bool:
             or rr_number["status"]["name"] == "new"
         ):
             return True
-        else:
-            return False
+        return False
```
okurz, line 65: same as above, change is fine

openqabot/utils.py
okurz: same as above, change is fine

okurz (changes requested):
all code changes look fine to me. Just some minor notes about the Makefile

Martchus (approved):
Looks all good to me, including the if-else changes. (I suppose the linter just doesn't like unnecessary elif/else branches when the if branch always returns.)

mpagot:
> Looks all good to me, including the if-else changes. (I suppose the linter just doesn't like unnecessary elif/else branches when the if branch always returns.)

Yes they are from pylint and named

```
  R1705, # no-else-return
```

- https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/no-else-return.html
-

Makefile
```diff
@@ -5,17 +5,34 @@ all:
...
 	python3 -m pytest -v --cov=./openqabot --cov-report=xml --cov-report=term
 
+# aggregate targets
+
+.PHONY: lint
+lint: pylint flake8
+
+.PHONY: checkstyle
+checkstyle:
+	black lint
```
okurz, line 35: does this actually work? IMHO it should be

```suggestion
checkstyle: black lint
```

.github/workflows/ci.yml
```diff
@@ -13,10 +13,12 @@ jobs:
           cache: 'pip'
       - name: Setup enviroment
         run: pip install -r requirements-dev.txt
+      - name: Linting
+        run: make lint
       - name: Run tests + cov report
         run: make test-with-coverage
       - name: Check style
-        run: make checkstyle
+        run: make black
```
okurz, line 21: I would combine those style checks so

```
      - name: Run tests + cov report
        run: make test-with-coverage
      - name: Check style
        run: make checkstyle
```

so effectively just leave it as is as you added lint+flake8 to checkstyle

okurz (changes requested):

.github/workflows/ci.yml
mpagot: I think it make sense for the user to have output from black separated from the one from (pylint and flake8)

pylint and flake8 also have some warning about formatting and indentation that black will detect as error too.

But the part I like the user perceive as lint is all the warning not about indentation like: unused variables, duplicated code, too complex code, ...

kalikiana: Ah, I guess the both of us had a similar idea. In #148 I actually used flake8-black which avoids having to call multiple tools and instead integrates them

okurz: > pylint and flake8 also have some warning about formatting and indentation that black will detect as error too.

You mean part of the checks are redundant? That's certainly true. But I don't understand why then you would need to separate.

> In order of importance the code should pass …

hm, I consider black always part of "static analysis". Another reason why I suggest to just call the meta-target checkstyle in CI as well is to make sure the target actually works to prevent errors like in https://github.com/openSUSE/qem-bot/pull/147#discussion_r1363818414 :)

mpagot: ok good for me to so :-)

okurz (approved):

PR #149, opened 2023-10-20
Refactor a bit

okurz:
* Make test_approve not rely on specific log level selection
* Refactor too big for-loop in types/incidents.py

### DIFF
--- a/openqabot/types/incidents.py
+++ b/openqabot/types/incidents.py
@@ -76,6 +76,199 @@ class Incidents(BaseConf):
 
         return False
 
+    def _handle_incident(
+        self,
+        inc: Incident,
+        arch,
+        flavor,
+        data,
+        token: Dict[str, str],
+        ci_url: Optional[str],
+        ignore_onetime: bool,
+    ) -> Dict[str, Any]:
+        DOWNLOAD_BASE = "http://download.suse.de/ibs/SUSE:/Maintenance:/"
+        BASE_PRIO = 50
+        if self.filter_embargoed() and inc.embargoed:
+            log.debug(
+                "Incident %s is embargoed and filtering embargoed updates enabled",
+                inc.id,
+            )
+            return None
+        full_post: Dict[str, Any] = {}
+        full_post["api"] = "api/incident_settings"
+        full_post["qem"] = {}
+        full_post["openqa"] = {}
+        full_post["openqa"].update(self.settings)
+        self.set_obsoletion(full_post["openqa"])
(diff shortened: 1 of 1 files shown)


openqabot/types/incidents.py
```diff
@@ -76,204 +76,215 @@ def _is_scheduled_job(
...
-                    # parse Public-Cloud pint query if present
-                    if "PUBLIC_CLOUD_PINT_QUERY" in settings:
-                        settings = apply_publiccloud_pint_image(settings)
-                        if not settings.get("PUBLIC_CLOUD_IMAGE_ID", False):
-                            continue
-
-                    full_post["openqa"] = settings
-                    full_post["qem"]["settings"] = settings
-                    ret.append(full_post)
+        ret = list(filter(None, ret))
```
mimi1vx, line 289: please no ... this is awful and really discouraged python syntax

`ret = [r for r in ret if r]`

mimi1vx (changes requested):

kalikiana:
I would prefer **Make test_approve not rely on specific log level selection** in its own PR rather than unrelated "bits"

mimi1vx:
> I would prefer **Make test_approve not rely on specific log level selection** in its own PR rather than unrelated "bits"

ah, its in different commits ...

Martchus (approved):

okurz:
> I would prefer **Make test_approve not rely on specific log level selection** in its own PR rather than unrelated "bits"

-> #150

openqabot/types/incidents.py
okurz: thank you. That's actually also what I prefer. Done

okurz:
> https://github.com/openSUSE/qem-bot/actions/runs/6586389281/job/17894634676?pr=149#step:6:17
> 
> ```
> ************* Module openqabot.types.incidents
> openqabot/types/incidents.py:79:4: R0911: Too many return statements (13/6) (too-many-return-statements)
> ```

Yeah, I guess that was expected due to the still many abort paths in the original code. I wonder why that's not an error aborting the CI run however I suggest we accept this refactoring first

a-kpappas:
This is the perfect opportunity to add some comments. What does 'handle' mean?

okurz:
> This is the perfect opportunity to add some comments. What does 'handle' mean?

It means "do stuff" as I don't have a better clue what it actually does ;)

okurz:
@mimi1vx do you request further changes or can approve?

mimi1vx (approved):

PR #150, opened 2023-10-20
Make test_approve not rely on specific log level selection

okurz:
Unit tests should not be too specific and we should be quite limited
with looking into actual log messages. This commit relaxes the checks
using caplog to the point what should be actually necessary for each
individual test step and not more. With this one can select an arbitrary
log level when calling pytest for easier development.

### DIFF
--- a/tests/test_approve.py
+++ b/tests/test_approve.py
@@ -193,7 +193,7 @@ def test_single_incident(fake_qem, caplog):
         ],
     )
     approver(incident=1)
-    assert len(caplog.records) == 5
+    assert len(caplog.records) >= 1, "we rely on log messages"
     messages = [x[-1] for x in caplog.record_tuples]
     assert (
         "SUSE:Maintenance:1:100 has at least one failed job in incident tests"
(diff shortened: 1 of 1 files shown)


kalikiana (approved):
Nice! That's definitely something that makes working on the code easier since several tests don't need to be updated when adding something new.

tests/test_approve.py
```diff
@@ -193,7 +193,7 @@ def test_single_incident(fake_qem, caplog):
         ],
     )
     approver(incident=1)
-    assert len(caplog.records) == 5
+    assert len(caplog.records) >= 1, "we rely on log messages"
```
mpagot, line 196: I suggest to also add short docstring to document which behavior this test is testing

okurz: Yes but I don't think this comment helps here.

kalikiana: The following asserts check that the expected messages are seen. Arguably this check is completely redundant.

okurz: Well, the check for the length of the arrays checks helps to prevent confusing IndexError exceptions

Martchus (approved):

PR #151, opened 2023-10-20
Partial refactoring of multi-level append into list concatenation - who can make it more pythonic and pretty?

okurz:
After
* #149

### DIFF
--- a/openqabot/types/incidents.py
+++ b/openqabot/types/incidents.py
@@ -278,13 +278,19 @@ class Incidents(BaseConf):
     ) -> List[Dict[str, Any]]:
         ret = []
 
+        def incident_post_settings(inc, arch, flavor, data):
+            return self._handle_incident(
+                inc, arch, flavor, data, token, ci_url, ignore_onetime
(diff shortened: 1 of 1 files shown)


okurz:
@Mergifyio rebase

openqabot/types/incidents.py
```diff
@@ -278,13 +278,19 @@ def __call__(
...
-                            inc, arch, flavor, data, token, ci_url, ignore_onetime
-                        )
-                    )
+                ret.append(
+                    [
+                        incident_post_settings(inc, arch, flavor, data)
+                        for inc in incidents
+                    ]
+                )
+        ret = [r for sublist in ret for r in sublist]
```
mimi1vx, line 294: maybe https://docs.python.org/3/library/itertools.html#itertools.chain.from_iterable here ? rest LGTM

okurz: ah, nice. Maybe I can try that eventually or somebody provides a working suggestions that I can take over.

PR #152, opened 2023-10-23
[WIP] Introduce mypy

mpagot:
Introduce mypy ad dev dependency.
Add mypy as make target.
Fix some mase errors.

### DIFF
--- a/openqabot/types/baseconf.py
+++ b/openqabot/types/baseconf.py
@@ -23,7 +23,8 @@ class BaseConf(metaclass=ABCMeta):
     ) -> List[Dict[str, Any]]:
         pass
 
-    @abstractstaticmethod
+    @abstractmethod
+    @staticmethod
     def normalize_repos(config):
         pass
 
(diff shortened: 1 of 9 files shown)


openqabot/types/baseconf.py
```diff
@@ -23,7 +23,8 @@ def __call__(
     ) -> List[Dict[str, Any]]:
         pass
 
-    @abstractstaticmethod
```
mimi1vx, line 26: why this ?

mpagot: https://docs.python.org/3/library/abc.html#abc.abstractclassmethod

okurz:
@mpagot hi, following your idea I have checked mypy and pyright and especially due to the much better runtime selected pyright in https://github.com/openSUSE/qem-bot/pull/312/commits/afb05e2dcab24ee1a6b2b611e50453cacfa2e483 . I hope you like it. Should we close this PR?

PR #154, opened 2024-01-30
Forbid failing incidents from being scheduled in aggregates

foursixnine:
@mimi1vx, @michaelgrifalconi

grafik

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ test: only-test checkstyle
 venv = qem-bot
 environment:
 	if [ ! -d "$(venv)" ]; then virtualenv -p python3 $(venv); fi
-	. $(venv)/bin/activate && pip install -r requirements.txt
+	. $(venv)/bin/activate && pip install -r requirements-dev.txt -r requirements.txt
 
 # devel: environment
 #   maybe use Makefile.venv instead to get a shell with virtualenv
--- a/Makefile
+++ b/Makefile
@@ -36,15 +36,15 @@ checkstyle: black lint
 .PHONY: test
 test: only-test checkstyle
 
-venv = qem-bot
-environment:
-	if [ ! -d "$(venv)" ]; then virtualenv -p python3 $(venv); fi
-	. $(venv)/bin/activate && pip install -r requirements-dev.txt -r requirements.txt
+VENV = qem-bot
+.PHONY environment:
+	if [ ! -d "$(VENV)" ]; then virtualenv -p python3 $(venv); fi
+	. $(VENV)/bin/activate && pip install -r requirements-dev.txt -r requirements.txt
 
 # devel: environment
-#   maybe use Makefile.venv instead to get a shell with virtualenv
+#   maybe use Makefile.VENV instead to get a shell with virtualenv
 # 	# we need to detect what shell we are using
 # 	shell=$$(basename $$SHELL); \
 # 	echo "Activating virtualenv for $$shell"; \
-# 	. $(venv) && \
+# 	. $(VENV) && \
 # 	exec $($(SHELL))
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ test: only-test checkstyle
 
 VENV = qem-bot
 .PHONY environment:
-	if [ ! -d "$(VENV)" ]; then virtualenv -p python3 $(venv); fi
+	if [ ! -d "$(VENV)" ]; then virtualenv -p python3 $(VENV); fi
 	. $(VENV)/bin/activate && pip install -r requirements-dev.txt -r requirements.txt
 
 # Developers have bad memory, so we need to remind them to activate the virtualenv
--- a/Makefile
+++ b/Makefile
@@ -41,8 +41,9 @@ VENV = qem-bot
 	if [ ! -d "$(VENV)" ]; then virtualenv -p python3 $(venv); fi
 	. $(VENV)/bin/activate && pip install -r requirements-dev.txt -r requirements.txt
 
-# devel: environment
-#   maybe use Makefile.VENV instead to get a shell with virtualenv
+# Developers have bad memory, so we need to remind them to activate the virtualenv
+# maybe use Makefile.VENV instead to get a shell with virtualenv
+# .PHONY devel: environment
 # 	# we need to detect what shell we are using
 # 	shell=$$(basename $$SHELL); \
 # 	echo "Activating virtualenv for $$shell"; \
--- a/Makefile
+++ b/Makefile
@@ -35,3 +35,16 @@ checkstyle: black lint
 
 .PHONY: test
 test: only-test checkstyle
+
+venv = qem-bot
+environment:
+	if [ ! -d "$(venv)" ]; then virtualenv -p python3 $(venv); fi
+	. $(venv)/bin/activate && pip install -r requirements.txt
+
+# devel: environment
+#   maybe use Makefile.venv instead to get a shell with virtualenv
+# 	# we need to detect what shell we are using
+# 	shell=$$(basename $$SHELL); \
+# 	echo "Activating virtualenv for $$shell"; \
+# 	. $(venv) && \
+# 	exec $($(SHELL))
--- a/openqabot/types/aggregate.py
+++ b/openqabot/types/aggregate.py
@@ -83,6 +83,9 @@ class Aggregate(BaseConf):
             # only testing queue and not livepatch
             valid_incidents = list()
             for i in incidents:
+                if i.has_failures(token):
+                    # filter out incidents which have already failed in single incident tests
+                    continue
                 if not any((i.livepatch, i.staging)):
                     # if filtering embargoed updates is on
                     if self.filter_embargoed(self.flavor) and i.embargoed:
@@ -96,15 +99,8 @@ class Aggregate(BaseConf):
                     else:
                         valid_incidents.append(i)
 
-            # filter out incidents which have already failed in single incident tests
-            filtered_incidents = [
-                incident
-                for incident in valid_incidents
-                if not incident.has_failures(token)
-            ]
-
             for issue, template in self.test_issues.items():
-                for inc in filtered_incidents:
+                for inc in valid_incidents:
                     if (
                         Repos(template.product, template.version, issues_arch)
                         in inc.channels
--- a/openqabot/types/aggregate.py
+++ b/openqabot/types/aggregate.py
@@ -26,6 +26,7 @@ class Aggregate(BaseConf):
         self.archs = config["archs"]
         self.onetime = config.get("onetime", False)
         self.test_issues = self.normalize_repos(config)
+        log.debug("Initialized Aggregate object for %s", self.product)
 
     @staticmethod
     def normalize_repos(config):
@@ -60,7 +61,9 @@ class Aggregate(BaseConf):
         ignore_onetime: bool = False,
     ) -> List[Dict[str, Any]]:
         ret = []
-
+        log.debug(
+            "Generating aggregate for %s the following incidents: %s", self, incidents
+        )
         for arch in self.archs:
             full_post: Dict["str", Any] = {}
             full_post["openqa"] = {}
--- a/openqabot/types/aggregate.py
+++ b/openqabot/types/aggregate.py
@@ -92,8 +92,16 @@ class Aggregate(BaseConf):
                     # if filtering embargoed updates is off we ignoring this field
                     else:
                         valid_incidents.append(i)
+
+            # filter out incidents which have already failed in single incident tests
+            filtered_incidents = [
+                incident
+                for incident in valid_incidents
+                if not incident.has_failures(token)
+            ]
+
             for issue, template in self.test_issues.items():
-                for inc in valid_incidents:
+                for inc in filtered_incidents:
                     if (
                         Repos(template.product, template.version, issues_arch)
                         in inc.channels
--- a/openqabot/types/incident.py
+++ b/openqabot/types/incident.py
@@ -3,10 +3,14 @@
 import re
 from logging import getLogger
 from typing import Dict, List, Tuple
+from urllib import request
 
+# Make sure we have the right dashboard URL
+from .. import QEM_DASHBOARD
 from . import ArchVer, Repos
-from ..errors import EmptyChannels, EmptyPackagesError, NoRepoFoundError
+from ..errors import EmptyChannels, EmptyPackagesError, NoRepoFoundError, NoResultsError
 from ..loader.repohash import get_max_revision
+
 
 log = getLogger("bot.types.incident")
 version_pattern = re.compile(r"(\d+(?:[.-](?:SP)?\d+)?)")
@@ -145,3 +149,30 @@ class Incident:
                 if package.startswith(req) and package != "kernel-livepatch-tools":
                     return True
         return False
+
+    def has_failures(self, token) -> bool:
+        results = request.get(
+            QEM_DASHBOARD + "/api/jobs/incident/" + str(self.id), headers=token
+        ).json()
+
+        # This should be wrapped by a verbose flag so we return faster
+        for res in results:
+            if res["status"] == "passed":
+                continue
+            else:
+                log.info(
+                    "Found failed, not-ignored job /t%s for incident %s",
+                    res["job_id"],
+                    self.id,
+                )
+                break
+
+        if not results:
+            raise NoResultsError(
+                "Job setting %s not found for incident" % (str(self.id))
+            )
+
+        # Return true if there are any failed jobs, so if results is not empty
+        # and there is at least one failed job, return true, any() returns false
+        # if the iterable is empty
+        return any(r["status"] == "failed" for r in results)
--- a/openqabot/types/incident.py
+++ b/openqabot/types/incident.py
@@ -6,7 +6,7 @@ from typing import Dict, List, Tuple
 from openqabot.utils import retry3 as requests
 
 # Make sure we have the right dashboard URL
-from .. import QEM_DASHBOARD
+from .. import QEM_DASHBOARD, OPENQA_URL
 from . import ArchVer, Repos
 from ..errors import EmptyChannels, EmptyPackagesError, NoRepoFoundError, NoResultsError
 from ..loader.repohash import get_max_revision
@@ -155,7 +155,7 @@ class Incident:
             QEM_DASHBOARD + "/api/jobs/incident/" + str(self.id), headers=token
         ).json()
 
-        failed_jobs = [res for res in results if res["status"] != "passed"]
+        failed_jobs = self.filter_failures(results)
         if failed_jobs:
             log.info("Found %s failed jobs for incident %s:", len(failed_jobs), self.id)
             if log.isEnabledFor(LOG_DEBUG_LEVEL):
@@ -177,3 +177,25 @@ class Incident:
             job_id,
             self.id,
         )
+
+    def filter_failures(self, results):
+        return [
+            res
+            for res in results
+            if res["status"] not in ("passed")
+            and not has_ignored_comment(res["job_id"], self.id)
+        ]
+
+
+@staticmethod
+def has_ignored_comment(job_id: int, inc: int):
+    ret = []
+    ret = requests.get(OPENQA_URL + "/api/v1/jobs/%s/comments" % job_id).json()
+    regex = re.compile(r"\@review\:acceptable_for\:incident_%s\:(.+)" % inc)
+    for comment in ret:
+        if regex.match(comment["text"]):
+            # leave comment for future debugging purposes
+            # log.debug("matched comment incident %s: with comment %s", inc, comment)
+            return True
+
+    return False
--- a/openqabot/types/incident.py
+++ b/openqabot/types/incident.py
@@ -1,9 +1,9 @@
 # Copyright SUSE LLC
 # SPDX-License-Identifier: MIT
 import re
-from logging import getLogger
+from logging import getLogger, DEBUG as LOG_DEBUG_LEVEL
 from typing import Dict, List, Tuple
-from urllib import request
+from openqabot.utils import retry3 as requests
 
 # Make sure we have the right dashboard URL
 from .. import QEM_DASHBOARD
@@ -151,21 +151,21 @@ class Incident:
         return False
 
     def has_failures(self, token) -> bool:
-        results = request.get(
+        results = requests.get(
             QEM_DASHBOARD + "/api/jobs/incident/" + str(self.id), headers=token
         ).json()
 
-        # This should be wrapped by a verbose flag so we return faster
-        for res in results:
-            if res["status"] == "passed":
-                continue
-            else:
-                log.info(
-                    "Found failed, not-ignored job /t%s for incident %s",
-                    res["job_id"],
+        failed_jobs = [res for res in results if res["status"] != "passed"]
+        if failed_jobs:
+            log.info("Found %s failed jobs for incident %s:", len(failed_jobs), self.id)
+            # if debug is enabled, print all failed jobs
+            if log.isEnabledFor(LOG_DEBUG_LEVEL):
+                log_debug = lambda job_id: log.debug(
+                    "Job %s is not marked as acceptable for incident %s",
+                    job_id,
                     self.id,
                 )
-                break
+                [log_debug(res["job_id"]) for res in failed_jobs]
 
         if not results:
             raise NoResultsError(
@@ -175,4 +175,4 @@ class Incident:
         # Return true if there are any failed jobs, so if results is not empty
         # and there is at least one failed job, return true, any() returns false
         # if the iterable is empty
-        return any(r["status"] == "failed" for r in results)
+        return any(failed_jobs)
--- a/openqabot/types/incident.py
+++ b/openqabot/types/incident.py
@@ -6,7 +6,7 @@ from typing import Dict, List, Tuple
 from openqabot.utils import retry3 as requests
 
 # Make sure we have the right dashboard URL
-from .. import QEM_DASHBOARD, OPENQA_URL
+from .. import QEM_DASHBOARD, OPENQA_URL, ACCEPTABLE_FOR_INCIDENT_REGEXP
 from . import ArchVer, Repos
 from ..errors import EmptyChannels, EmptyPackagesError, NoRepoFoundError, NoResultsError
 from ..loader.repohash import get_max_revision
@@ -188,13 +188,18 @@ class Incident:
 
 
 @staticmethod
+# TODO:
+# - move to utils
+# - remove almost duplicated code from Approver.is_job_marked_acceptable_for_incident
+#   as approver does not seem to operate over incidents
 def has_ignored_comment(job_id: int, inc: int):
     ret = []
     ret = requests.get(OPENQA_URL + "/api/v1/jobs/%s/comments" % job_id).json()
-    regex = re.compile(r"\@review\:acceptable_for\:incident_%s\:(.+)" % inc)
+    regex = re.compile(ACCEPTABLE_FOR_INCIDENT_REGEXP % inc)
     for comment in ret:
         if regex.match(comment["text"]):
-            # leave comment for future debugging purposes
+            # leave comment for future debugging purposes, but don't spam the log
+            # as it pollutes the log with irrelevant information
             # log.debug("matched comment incident %s: with comment %s", inc, comment)
             return True
 
--- a/openqabot/types/incident.py
+++ b/openqabot/types/incident.py
@@ -187,11 +187,13 @@ class Incident:
         ]
 
 
-@staticmethod
+# pylint: disable=fixme
 # TODO:
-# - move to utils
-# - remove almost duplicated code from Approver.is_job_marked_acceptable_for_incident
+#   - move to utils.py or a better place
+#   - remove almost duplicated code from Approver.is_job_marked_acceptable_for_incident
 #   as approver does not seem to operate over incidents
+#   about the TODO see discussion at https://github.com/openSUSE/qem-bot/pull/154#discussion_r1472721681
+@staticmethod
 def has_ignored_comment(job_id: int, inc: int):
     ret = []
     ret = requests.get(OPENQA_URL + "/api/v1/jobs/%s/comments" % job_id).json()
--- a/openqabot/types/incident.py
+++ b/openqabot/types/incident.py
@@ -159,12 +159,9 @@ class Incident:
         if failed_jobs:
             log.info("Found %s failed jobs for incident %s:", len(failed_jobs), self.id)
             if log.isEnabledFor(LOG_DEBUG_LEVEL):
-                log_debug = lambda job_id: log.debug(
-                    "Job %s is not marked as acceptable for incident %s",
-                    job_id,
-                    self.id,
+                list(
+                    map(lambda job: self.log_debug_incident(job["job_id"]), failed_jobs)
                 )
-                [log_debug(res["job_id"]) for res in failed_jobs]
 
         if not results:
             raise NoResultsError(
@@ -173,3 +170,10 @@ class Incident:
 
         # we only need to check if there are any failed jobs
         return any(failed_jobs)
+
+    def log_debug_incident(self, job_id):
+        log.debug(
+            "Job %s is not marked as acceptable for incident %s",
+            job_id,
+            self.id,
+        )
--- a/openqabot/types/incident.py
+++ b/openqabot/types/incident.py
@@ -158,7 +158,6 @@ class Incident:
         failed_jobs = [res for res in results if res["status"] != "passed"]
         if failed_jobs:
             log.info("Found %s failed jobs for incident %s:", len(failed_jobs), self.id)
-            # if debug is enabled, print all failed jobs
             if log.isEnabledFor(LOG_DEBUG_LEVEL):
                 log_debug = lambda job_id: log.debug(
                     "Job %s is not marked as acceptable for incident %s",
@@ -172,7 +171,5 @@ class Incident:
                 "Job setting %s not found for incident" % (str(self.id))
             )
 
-        # Return true if there are any failed jobs, so if results is not empty
-        # and there is at least one failed job, return true, any() returns false
-        # if the iterable is empty
+        # we only need to check if there are any failed jobs
         return any(failed_jobs)
--- a/openqabot/types/incidents.py
+++ b/openqabot/types/incidents.py
@@ -54,7 +54,7 @@ class Incidents(BaseConf):
                 f"{QEM_DASHBOARD}api/incident_settings/{inc.id}",
                 headers=token,
             ).json()
-        except Exception as e:  # pylint: disable=broad-except
+        except Exception as e:
             log.exception(e)
 
         if not jobs:
@@ -76,7 +76,8 @@ class Incidents(BaseConf):
 
         return False
 
-    def _handle_incident(  # pylint: disable=too-many-return-statements
+    # pylint: disable=too-many-return-statements
+    def _handle_incident(
         self,
         inc: Incident,
         arch,
--- a/tests/test_aggregate.py
+++ b/tests/test_aggregate.py
@@ -91,15 +91,24 @@ def incident_mock(monkeypatch):
         arch: str
 
     class MockIncident:
-        def __init__(self, repo, embargoed):
+        def __init__(self, repo, embargoed, has_failures, mocked_incident=42):
+            self.id = mocked_incident
             self.livepatch = None
             self.staging = None
             self.channels = [repo]
             self.embargoed = embargoed
+            self._has_failures = has_failures
 
-    def _func(product, version, arch, embargoed=False):
+        def has_failures(self, token):
+            return self._has_failures
+
+    def _func(
+        product, version, arch, embargoed=False, has_failures=False, mocked_incident=42
+    ):
         repo = Repos(product=product, version=version, arch=arch)
-        return MockIncident(repo, embargoed=embargoed)
+        return MockIncident(
+            repo, embargoed=embargoed, has_failures=False, mocked_incident=42
+        )
 
     return _func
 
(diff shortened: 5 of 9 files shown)


openqabot/types/incident.py
```diff
@@ -144,3 +148,30 @@ def contains_package(self, requires: List[str]) -> bool:
                 if package.startswith(req) and package != "kernel-livepatch-tools":
                     return True
         return False
+
+    def has_failures(self, token) -> bool:
+        results = request.get(
+            QEM_DASHBOARD + "/api/jobs/incident/" + str(self.id), headers=token
+        ).json()
+
+        # This should be wrapped by a verbose flag so we return faster
```
okurz, line 157: I don't understand this comment. Can you elaborate?

openqabot/types/incident.py
```diff
@@ -144,3 +148,30 @@ def contains_package(self, requires: List[str]) -> bool:
...
+        for res in results:
+            if res["status"] == "passed":
+                continue
+            else:
```
okurz, line 161: This seems redundant. How about just

```suggestion
            if res["status"] != "passed":
```

openqabot/types/aggregate.py
```diff
@@ -95,8 +95,16 @@ def __call__(
...
                     else:
                         valid_incidents.append(i)
+
+            # filter out incidents which have already failed in single incident tests
+            filtered_incidents = [
+                incident
+                for incident in valid_incidents
+                if not incident.has_failures(token)
+            ]
+
```
okurz, line 105:
```suggestion
            filtered_incidents = [i for i in valid_incidents if not i.has_failures(token)]
            log.info("filtered out incidents with failed single incident tests: %s", ",".join(i for i in filtered_incidents))
```

This way we don't need the comment but instead have an informative message

okurz (changes requested):

openqabot/types/incident.py
foursixnine: See above, I don't put too much brainpower on thtis

foursixnine:
@okurz since you're requesting changes already, can you guide me in the test/code paths? `the_demon_incident` in theory should not show up, any thoughts?

openqabot/types/aggregate.py
foursixnine: Nope, I prefer to leave the presence/absence of messages to has_failures

openqabot/types/incident.py
okurz: Well, the level of verbosity is already defined by choosing the log-level. I agree with the "info" level below so I suggest you just remove this comment then

openqabot/types/incident.py
okurz: That does not resolve the discussion point though. I might be wrong then please prove me wrong, otherwise my suggestion holds

openqabot/types/aggregate.py
okurz: That does not resolve my point though. Then log in has_failures and leave the the comment for good.

okurz (changes requested):

openqabot/types/incident.py
foursixnine: @okurz rather than wasting time over style issues, lets focus on the important part of the problem:

can you guide me in the test/code paths? `the_demon_incident` in theory should not show up, any thoughts?

openqabot/types/aggregate.py
foursixnine: That's whats gonna happen

foursixnine:
@okurz can you please tag this pr as `ai-assisted`?

Martchus:
We don't have that kind of label yet (in that repo). Are you saying you've been using AI here? If yes, why is that relevant and what would adding that tag change?

foursixnine:
> We don't have that kind of label yet (in that repo). Are you saying you've been using AI here? If yes, why is that relevant and what would adding that tag change?

It add traceability, of things that have been done using some sort of assistance, while not important for you, SUSE-wise it is.

tests/test_incident.py
```diff
@@ -117,3 +126,53 @@ def test_inc_revisions(mock_good):
...
+        caplog.records[1].message
+        == "Job 2 is not marked as acceptable for incident 24618"
+    )
+    assert len(caplog.records) == 2
+
+    caplog.set_level(logging.INFO)
+    caplog.clear()
+
+    # Assert that the method returns True since there is a failed job
+    inc.has_failures("token")
```
Martchus, line 175: This line does not assert anything.

Martchus (changes requested):
Looks generally good.

tests/test_incident.py
foursixnine: It doesn't need to, see assertion below

foursixnine: @Martchus I see it now :) fixed

openqabot/types/incident.py
```diff
@@ -150,21 +150,21 @@ def contains_package(self, requires: List[str]) -> bool:
...
-            if res["status"] == "passed":
-                continue
-            else:
-                log.info(
-                    "Found failed, not-ignored job /t%s for incident %s",
-                    res["job_id"],
+        failed_jobs = [res for res in results if res["status"] != "passed"]
+        if failed_jobs:
+            log.info("Found %s failed jobs for incident %s:", len(failed_jobs), self.id)
+            # if debug is enabled, print all failed jobs
```
okurz, line 160: that comment is redundant

tests/test_incident.py
```diff
@@ -117,3 +126,53 @@ def test_inc_revisions(mock_good):
     assert incident.revisions_with_fallback("x86_64", "12-SP5")
     assert not incident.revisions_with_fallback("aarch64", "12")
     assert not incident.revisions_with_fallback("aarch64", "12-SP5")
+
+
+class MockResponse:
+    # TODO: collect all instances where the same pattern is used and refactor,
+    # see openSUSE/qem-bot/issues/161
```
okurz, line 133: Don't add TODO. Better leave it out and keep in your local git repository, either in `git stash` or a temporary "WIP"-commit. We have the issue as reference after all which can suffice.

okurz (changes requested):

tests/test_incident.py
foursixnine: Nope, I'll keep the to-do for the sanity of whoever has to work on this in the future. grepping for `TODO` + support in modern-day editors makes it easier and less prone to be forgotten, should somebody work on it in the future.

If you _absolutely_ don't want to have the todo, do a commit to the branch

okurz:
> @okurz can you please tag this pr as `ai-assisted`?

As I explained I don't think it's a good idea but as you insist I created and added that label.

tests/test_incident.py
okurz: TODOs in code don't work. Everybody starting to apply changes to the code comes from a certain motivation and they want to get something done. In my experience it is hardly the case that somebody is on it developing a certain feature, stumbling about a TODO and saying "oh, I wanted to get $foo done. I will stop that for now and fix that issue before I continue". If we find other contributors to this repository to convince us that TODO's in code are good then I will accept :)

Makefile
```diff
@@ -35,3 +35,16 @@ checkstyle: black lint
 
 .PHONY: test
 test: only-test checkstyle
+
+venv = qem-bot
+environment:
```
okurz, line 40:
```suggestion
.PHONY: environment
environment:
```

Makefile
```diff
@@ -35,3 +35,16 @@ checkstyle: black lint
...
+	if [ ! -d "$(venv)" ]; then virtualenv -p python3 $(venv); fi
```
okurz, line 41: Where should the application "venv" come from? `$(venv)` would be the invocation of an application "venv" that should return a directory path. I don't see where this would come from

openqabot/types/aggregate.py
```diff
@@ -60,7 +61,9 @@ def __call__(
         ignore_onetime: bool = False,
     ) -> List[Dict[str, Any]]:
         ret = []
-
+        log.debug(
+            "Generating aggregate for %s the following incidents: %s", self, incidents
```
okurz, line 65: I don't understand that sentence. The object reference `self` would be put into the first `%s`? How would an example log line look like?

tests/test_aggregate.py
```diff
@@ -91,34 +93,106 @@ class Repos(NamedTuple):
...
+    these_incidents.append(the_moved_incident)
+    logger.info("Testing the the_moved_incident (%s)", len(res))
+
+    res = acc(
+        these_incidents,
+        token=None,
+        ci_url=None,
+    )
+
+    assert len(res) == 2  # the demon incident is not added, because it has failures
```
okurz, line 195: I am not sure what you are asserting here. This is where https://github.com/openSUSE/qem-bot/actions/runs/7724416299/job/21056528243?pr=154#step:5:178 fails. Your comment says "the demon incident is not added, because it has failures" which is what https://github.com/openSUSE/qem-bot/pull/154/files#diff-09e8e60e6d3ec2a94d9f6571f313fef2ea2e181b40892485011bea95c7e11f66R181 already stated. But you assume that the "moved incident" is then included, right? Unfortunately I don't have further hints. I would use the debugger to step through the execution here as my experience with this code is also quite limited.

openqabot/types/aggregate.py
foursixnine: https://github.com/openSUSE/qem-bot/actions/runs/7724416299/job/21056528243?pr=154#step:5:185

tests/test_incident.py
foursixnine: @okurz I'm not having the discussion about TODO's. If you don't want TODOs make an RFC for all of the projects, and we can bring up the same discussion from https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/12972#issuecomment-1073722468

You're subjecting everybody to empirical experience, while in reality [they are quite used](https://github.com/search?q=repo%3Atorvalds%2Flinux+TODO%3A&type=commits) and [helpful](https://github.com/torvalds/linux/commit/2ef08b98025bd09b74f68d1801995b0b068afbe7) overall.

So solve that one first, and then ping me and I'll add the comment :).

tests/test_aggregate.py
foursixnine: > I am not sure what you are asserting here. This is where https://github.com/openSUSE/qem-bot/actions/runs/7724416299/job/21056528243?pr=154#step:5:178 fails. Your comment says "the demon incident is not added, because it has failures" which is what https://github.com/openSUSE/qem-bot/pull/154/files#diff-09e8e60e6d3ec2a94d9f6571f313fef2ea2e181b40892485011bea95c7e11f66R181 already stated. But you assume that the "moved incident" is then included, right? Unfortunately I don't have further hints. I would use the debugger to step through the execution here as my experience with this code is also quite limited.

Problem was that `res` is a list of whatever, but the information about the incidents was hidden in the elements of it, so `res[0][qem][incidents]` was the code I was looking for

so all this time, the incidents were like:

grafik

foursixnine:
> > @okurz can you please tag this pr as `ai-assisted`?
> 
> As I explained I don't think it's a good idea

That wasn't entirely clear :), however to leave public evidence:

I’m doing it under the Copilot pilot thing, and I’m one of the persons pushing for proper traceability, because Stefano sucks at whatever he was trying to do:
https://suse.slack.com/archives/C05QMF9RSUB/p1700063860348349

>  but as you insist I created and added that label.

I appreciate it <3

Makefile
foursixnine: Any particular reason for the preference?

openqabot/types/aggregate.py
okurz: Got it, so

```
bot.types.aggregate:aggregate.py:64 Generating aggregate for <Aggregate product: > the following incidents: [<MockIncident id: 1284 with repos [Repos(product='BBBBBBBBB', version='CCCCCCCC', arch='ciao')]>]
```

not a perfect format but for a "debug" message acceptable

Makefile
okurz: well, because the target is not a file "environment" but an abstract target, same as in other ones

openqabot/types/incidents.py
```diff
@@ -54,7 +54,7 @@ def _is_scheduled_job(
                 f"{QEM_DASHBOARD}api/incident_settings/{inc.id}",
                 headers=token,
             ).json()
-        except Exception as e:  # pylint: disable=broad-except
+        except Exception as e:
```
okurz, line 57: this seems to have been lost in rebasing

foursixnine: Linter doesn't complain though

openqabot/types/incident.py
```diff
@@ -144,3 +148,31 @@ def contains_package(self, requires: List[str]) -> bool:
                 if package.startswith(req) and package != "kernel-livepatch-tools":
                     return True
         return False
+
+    def has_failures(self, token) -> bool:
+        results = requests.get(
+            QEM_DASHBOARD + "/api/jobs/incident/" + str(self.id), headers=token
+        ).json()
+
+        failed_jobs = [res for res in results if res["status"] != "passed"]
```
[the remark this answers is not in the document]
foursixnine: Good catch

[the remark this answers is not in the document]
okurz: I assume https://github.com/openSUSE/qem-bot/blob/9872fbc88ccc14968f066f54b8565049754541bc/openqabot/utils.py#L57 has been called already so that passed and soft failed are both treated the same. In openQA we call those "ok results"

Makefile
```diff
@@ -35,3 +35,16 @@ checkstyle: black lint
...
+	if [ ! -d "$(VENV)" ]; then virtualenv -p python3 $(venv); fi
+	. $(VENV)/bin/activate && pip install -r requirements-dev.txt -r requirements.txt
+
+# devel: environment
+#   maybe use Makefile.VENV instead to get a shell with virtualenv
+# 	# we need to detect what shell we are using
+# 	shell=$$(basename $$SHELL); \
+# 	echo "Activating virtualenv for $$shell"; \
+# 	. $(VENV) && \
+# 	exec $($(SHELL))
```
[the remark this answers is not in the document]
foursixnine: Not really, updated the comment though; Not adding another TODO, to avoid causing a heart attack to @okurz

tests/test_incident.py
```diff
@@ -117,3 +126,89 @@ def test_inc_revisions(mock_good):
...
+        url=url,
+        json_data=[
+            {"status": "passed", "job_id": 1},
+            {"status": "failed", "job_id": 1777},  # Accept the turk
+            {
+                "status": "softfailed",
+                "job_id": 2020,
+            },  # 2020 is the genesys of dark fate
+            {"status": "failed", "job_id": 2042},  # This one has a dark fate
+            {"status": "passed", "job_id": 3},
```
foursixnine, line 178: Python linting can be ugly at times 🗡️

Martchus (approved):

openqabot/types/incident.py
```diff
@@ -144,3 +148,53 @@ def contains_package(self, requires: List[str]) -> bool:
...
+            "Job %s is not marked as acceptable for incident %s",
+            job_id,
+            self.id,
+        )
+
+    def filter_failures(self, results):
+        return [
+            res
+            for res in results
+            if res["status"] not in ("passed", "softfailed")
```
okurz, line 184: Are you sure the status can ever be soft failed here? I don't think this is the direct openQA job result here

openqabot/types/incident.py
```diff
@@ -144,3 +148,53 @@ def contains_package(self, requires: List[str]) -> bool:
...
+
+@staticmethod
+def has_ignored_comment(job_id: int, inc: int):
+    ret = []
+    ret = requests.get(OPENQA_URL + "/api/v1/jobs/%s/comments" % job_id).json()
+    regex = re.compile(r"\@review\:acceptable_for\:incident_%s\:(.+)" % inc)
+    for comment in ret:
+        if regex.match(comment["text"]):
+            # leave comment for future debugging purposes
+            # log.debug("matched comment incident %s: with comment %s", inc, comment)
```
okurz, line 197: Don't add disabled code as nobody but you will understand why this is not active. Better leave it out and keep in your local git repository, either in `git stash` or a temporary "WIP"-commit

See https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/CONTRIBUTING.md#coding-style for more details

pylintrc
```diff
@@ -2,7 +2,6 @@
 
 disable=
   missing-docstring,
-  bad-continuation,
```
okurz, line 5: Why did you remove that?

openqabot/types/incident.py
```diff
@@ -144,3 +148,53 @@ def contains_package(self, requires: List[str]) -> bool:
...
+            if res["status"] not in ("passed", "softfailed")
+            and not has_ignored_comment(res["job_id"], self.id)
+        ]
+
+
+@staticmethod
+def has_ignored_comment(job_id: int, inc: int):
+    ret = []
+    ret = requests.get(OPENQA_URL + "/api/v1/jobs/%s/comments" % job_id).json()
+    regex = re.compile(r"\@review\:acceptable_for\:incident_%s\:(.+)" % inc)
```
okurz, line 193: Wait. This was already used somewhere. Did you just copy-paste that?

okurz (changes requested):

openqabot/types/incident.py
foursixnine: @okurz can you suggest a better way? wasn't really a fan of this, but didn't want to give you a heart attack by adding another TODO

pylintrc
foursixnine: Because the pylint would "warn" otherwise that this is deprecated.

openqabot/types/incident.py
foursixnine: > Don't add disabled code as nobody but you will understand why this is not active.

That only applies if the person looking at the code doesn't read the actual comment stating why it is there

openqabot/types/incident.py
foursixnine: Better safe than sorry. This is a result comming directly from smelt. One important thing of this code, is that it makes clear to whoever is looking at it, what are the conditions that need to be met. And I don't see a future where we would like to give a different treatment to softfails.

Makefile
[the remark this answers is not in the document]
foursixnine: @okurz thank you for the recommendation, however same comment applies:

https://github.com/openSUSE/qem-bot/pull/154#discussion_r1473869498

pylintrc
foursixnine:
```
************* Module pylintrc
pylintrc:1:0: R0022: Useless option value for '--disable', 'bad-continuation' was removed from pylint, see https://github.com/pylint-dev/pylint/pull/3571. (useless-option-value)
```

openqabot/types/incident.py
michaelgrifalconi: Had a look into the bot/dashboard code.
Seems like we are using  [dashboard data](https://github.com/openSUSE/qem-bot/pull/154/files/b194cd734e1da3cfe79a033a705aab6340a5a797#diff-2a7060f0a42828a4dbbbebdd465bf429c4c2d0e4d4760342da54546f1fd97480R154), and the dashboard [hides softfails as passed](https://github.com/openSUSE/qem-dashboard/blob/2d5dfece9fca7e7fc9f4c3970c096a57cbb44a8c/lib/Dashboard/Model/Jobs.pm#L103). 
So right now, if the dashboard do not change behavior, softfails will never happen in this code so we could skip it.
But that would mean building yet one more thing that only works on top of some "processed" data from the dashboard database and will break in weird ways if you try to run it with data direct from source.

foursixnine: > Had a look into the bot/dashboard code. Seems like we are using [dashboard data](https://github.com/openSUSE/qem-bot/pull/154/files/b194cd734e1da3cfe79a033a705aab6340a5a797#diff-2a7060f0a42828a4dbbbebdd465bf429c4c2d0e4d4760342da54546f1fd97480R154), and the dashboard [hides softfails as passed](https://github.com/openSUSE/qem-dashboard/blob/2d5dfece9fca7e7fc9f4c3970c096a57cbb44a8c/lib/Dashboard/Model/Jobs.pm#L103). So right now, if the dashboard do not change behavior, softfails will never happen in this code so we could skip it. But that would mean building yet one more thing that only works on top of some "processed" data from the dashboard database and will break in weird ways if you try to run it with data direct from source.

I wonder if it makes more sense to query openQA directly, since that's being done too when looking for the comments. (would be done in a separate PR though)

openqabot/types/incident.py
okurz: Extract the existing code into something reusable and call it in both places

pylintrc
okurz: Yes, I know. Unfortunately current pylint on current openSUSE Leap 15.5 still has the check included and would actually fail style checks if not ignored. We could only exclude "bad-continuation" based on parsing the version of pylint itself on invocation

openqabot/types/incident.py
okurz: Having a commented out log.debug line does not make any sense to me. I am fine to just have the log.debug enabled. If the debug log level is too noisy for anyone then they can set to only log with "info" level

openqabot/types/incident.py
okurz: Yes, querying live results can be considered but please leave that to a separate PR. And regarding "softfailed", same as @michaelgrifalconi explained, does not make sense here. The "passed" is not the openQA-"passed". And "better safe than sorry" is not a good reasoning to introduce wrong and confusing code. We don't want to have

```suggestion
            if res["status"] not in ("passed", "softfailed", "not-ok", "whoknows", "whatever")
```

"just to be safe", right?

alvarocarvajald, line 184: Sorry if this was already asked, but would using `if res["status"] not in ("passed", "softfailed")` treat running/not-completed jobs as failures, and filter more than what is intended?

alvarocarvajald (commented):
I left an in-line question, and I have another: suppose it's one of those weeks, and there are a bunch of failures in Incidents due to non-product issues (communication issues between SUSE's network and Cloud Service Providers, issues in SCC, Multi-Machine issues, the list does go on) ... this will cause many test failures, which in turn will cause multiple incidents to being incorrectly filtered. Is this an acceptable outcome? I have to admit that from my side, I tend to notice MM issues quicker when seeing the errors in Aggregates. If too many Incidents are filtered, and Aggregates don't run, I think we can arrive to the same conclusion but not as quickly. Just my $0.02.

openqabot/types/incident.py
foursixnine: This is addresed in https://github.com/openSUSE/qem-bot/pull/154#discussion_r1474042954 since we're querying smelt, it makes sense to use that data instead of anything else

michaelgrifalconi: That's is correct, but since aggregates are scheduled late in the evening (European time) I imagine the impact is not huge. And right now if a new RR comes out after aggregates are scheduled, they also wait another day, so we are enlarging that window a little bit (depending on how much time incidents need to finish)

kalikiana: Let's keep it technical, shall we? ;-) The dashboard can return passed, waiting, stopped or failed. If the concern here is that the dashboard might change behavior we should ensure it has according unit tests.

kalikiana: FYI I filed https://github.com/openSUSE/qem-dashboard/issues/810

openqabot/types/incident.py
kalikiana: The **TODO** says `remove almost duplicated code from Approver.is_job_marked_acceptable_for_incident`. I suggest you either kill the TODO or actually move the code if it's easy enough.

openqabot/types/incident.py
foursixnine: > Yes, querying live results can be considered but please leave that to a separate PR. And regarding "softfailed", same as @michaelgrifalconi explained, does not make sense here. The "passed" is not the openQA-"passed". And "better safe than sorry" is not a good reasoning to introduce wrong and confusing code. We don't want to have
> 
> "just to be safe", right?

Addressed in 62cb3d29d710c849c83effa4680ae2740a1800b0, so for now let's be exclusive

foursixnine:
@okurz I think that despite the arguments about the commented code, this PR is ready, I note the objections and do plan to follow up on the bits and pieces starting with #161

okurz (changes requested):
Sorry, but I don't see that we should merge this.

Please do not resolve discussions when they are actually not resolved.

You can of course run the code as is from your fork and you don't need this to be merged just to execute for testing purposes. Please consider https://github.com/openSUSE/qem-bot/?tab=readme-ov-file#local-testing if not already done.

pylintrc
okurz: But that's not the reason for failure, that is

```
************* Module openqabot.types.incident
openqabot/types/incident.py:190:1: W0511: TODO: (fixme)
```

openqabot/types/incidents.py
```diff
@@ -54,7 +54,8 @@ def _is_scheduled_job(
                 f"{QEM_DASHBOARD}api/incident_settings/{inc.id}",
                 headers=token,
             ).json()
-        except Exception as e:  # pylint: disable=broad-except
+        # pylint: disable=broad-except
+        except Exception as e:
```
okurz, line 58: why did you need to change those?

tests/test_aggregate.py
```diff
@@ -91,34 +96,99 @@ class Repos(NamedTuple):
...
+        def __init__(self, repo, embargoed, mocked_incident=42):
+            self.id = mocked_incident
             self.livepatch = None
             self.staging = None
             self.channels = [repo]
             self.embargoed = embargoed
+            self._has_failures = False
+
+        def has_failures(self, token):
+            does_it_have_failures = self.id == 666 or self._has_failures
```
okurz, line 108:
```suggestion
            has_failures = self.id == 666 or self._has_failures
```

tests/test_incident.py
okurz: We don't need an RFC when common reviewers agree on a style and enforce that by reviewing and approving. If there are multiple other project maintainers that tell me that we commonly add TODOs and it's a good thing then I will stand corrected and accept your code as is. So for sure I don't want to just make that my personal opinion.

tests/test_incident.py
```diff
@@ -117,3 +126,89 @@ def test_inc_revisions(mock_good):
...
+
+
+def mock_get(url, extra_data=None, headers=None):
+    return MockResponse(
+        url=url,
+        json_data=[
+            {"status": "passed", "job_id": 1},
+            {"status": "failed", "job_id": 1777},  # Accept the turk
+            {
+                "status": "softfailed",
```
okurz, line 175: @kraih this looks like it's mocking a response from qem-dashboard that can actually never happen in reality as the dashboard would not return a status "softfailed", am I right?

okurz (changes requested):

openqabot/types/incident.py
baierjan: What should be the difference between debug and trace level here? I also thing the message is perfectly fine there with debug level. If that message is helpful during debugging (as the comment suggests) it has no meaning to leave that commented out.

tests/test_incident.py
baierjan: I also guess so. The status from dashboard should be seen from the perspective of [approval job](https://github.com/openSUSE/qem-bot/pull/154/files#diff-d44cab80bf43c62865043529f6e9883cc12ab37d49f5fb5af9e1bdc7c012dcaaL168) which only has binary result: either approve or not. An openQA softfail should not block approval and hence is masked by dashboard as passed (as already linked in another conversation).

openqabot/types/incident.py
baierjan: Also, we shifted some messages between levels in https://github.com/openSUSE/qem-bot/pull/19 and similar and finally ended with limiting the automated pipeline to info level only. So the debug level is indeed intended for any helpful messages for manual debugging.

pylintrc
foursixnine: Both need to be removed iirc. let me remove the commit and see

************* Module pylintrc
pylintrc:1:0: R0022: Useless option value for '--disable', 'bad-continuation' was removed from pylint, see https://github.com/pylint-dev/pylint/pull/3571. (useless-option-value)
************* Module openqabot.types.incident
openqabot/types/incident.py:190:1: W0511: TODO: (fixme)

tests/test_incident.py
baierjan: Can't we just replace the TODO by a link to a progress ticket describing the problem? That can assure someone will solve that eventually. We do not have any process in place which involves greping code for TODOs when we are deciding what to do next.

Makefile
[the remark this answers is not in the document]
Martchus: It would make sense to at least state why this code has been disabled; e.g. why it is not good/useful enough for general use and in what situations it would make sense to use it nevertheless. The comments

```
# Developers have bad memory, so we need to remind them to activate the virtualenv
# maybe use Makefile.VENV instead to get a shell with virtualenv
```

don't make that clear to me at all.

Additionally, also if that code was not commented-out I'd frankly struggle to make sense of its intended use and purpose. So that should probably be clarified anyway.

openqabot/types/incident.py
Martchus: I'd also say this code should be either enabled or removed completely (but wouldn't make a big deal out of it either).

Martchus (changes requested):
see @okurz's comments

foursixnine:
> good idea

leaving here as by mistake I copied a message I sent myself:

> > @okurz can you please tag this pr as `ai-assisted`?
> 
> As I explained I don't think it's a good idea

That wasn't entirely clear :), however to leave public evidence:

I’m doing it under the Copilot pilot thing, and I’m one of the persons pushing for proper traceability:
https://suse.slack.com/archives/C05QMF9RSUB/p1700063860348349

>  but as you insist I created and added that label.

I appreciate it <3

tests/test_aggregate.py
foursixnine: fixed

openqabot/types/incidents.py
foursixnine: Messy rebases :D

pylintrc
foursixnine: @okurz removing that line, the warning still shows up: https://github.com/openSUSE/qem-bot/actions/runs/7757445750/job/21156903452#step:6:17

I leave it to you for a different PR if you want to remove such warning, I doubt Leap version has anything to do with linting rules.

openqabot/types/incident.py
foursixnine: In fact, after last changes, this is not necessary anymore and

Makefile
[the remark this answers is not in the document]
foursixnine: > Additionally, also if that code was not commented-out I'd frankly struggle to make sense of its intended use and purpose. So that should probably be clarified anyway.

Good point, updated the comment

tests/test_incident.py
foursixnine: This was a leftover

foursixnine:
> Sorry, but I don't see that we should merge this.

I didn't ask explicitly for a merge, that's for later if needed. No need to be sorry.

Also, if there's a deployment pipeline that doesn't have an E2E, merging of fridays is a big no.  this is something that can likely cause some people to go around screaming in chicken headless mode anywhere.

What I'm asking is: its important for me, for you to be through and focus on the technical side of the review; because of the criticallity of it.

> Please consider https://github.com/openSUSE/qem-bot/?tab=readme-ov-file#local-testing if not already done.

I'll consider it.

Makefile
```diff
@@ -35,3 +35,17 @@ checkstyle: black lint
 
 .PHONY: test
 test: only-test checkstyle
+
+VENV = qem-bot
+.PHONY environment:
+	if [ ! -d "$(VENV)" ]; then virtualenv -p python3 $(VENV); fi
```
mimi1vx, line 41: maybe better is use of `venv` module of python stdlib ?

openqabot/types/incident.py
```diff
@@ -144,3 +148,60 @@ def contains_package(self, requires: List[str]) -> bool:
...
+
+    def has_failures(self, token) -> bool:
+        results = requests.get(
+            QEM_DASHBOARD + "/api/jobs/incident/" + str(self.id), headers=token
+        ).json()
+
+        failed_jobs = self.filter_failures(results)
+        if failed_jobs:
+            log.info("Found %s failed jobs for incident %s:", len(failed_jobs), self.id)
+            if log.isEnabledFor(LOG_DEBUG_LEVEL):
```
mimi1vx, line 160: this is really ugly

openqabot/types/incident.py
```diff
@@ -144,3 +148,60 @@ def contains_package(self, requires: List[str]) -> bool:
...
+                list(
+                    map(lambda job: self.log_debug_incident(job["job_id"]), failed_jobs)
```
mimi1vx, line 162: this should be standard `for` loop

openqabot/types/incident.py
```diff
@@ -144,3 +148,60 @@ def contains_package(self, requires: List[str]) -> bool:
...
+        ]
+
+
+# pylint: disable=fixme
+# TODO:
+#   - move to utils.py or a better place
+#   - remove almost duplicated code from Approver.is_job_marked_acceptable_for_incident
+#   as approver does not seem to operate over incidents
+#   about the TODO see discussion at https://github.com/openSUSE/qem-bot/pull/154#discussion_r1472721681
+@staticmethod
```
mimi1vx, line 195: static method of what ??

openqabot/types/incident.py
```diff
@@ -144,3 +148,60 @@ def contains_package(self, requires: List[str]) -> bool:
...
+        failed_jobs = self.filter_failures(results)
+        if failed_jobs:
+            log.info("Found %s failed jobs for incident %s:", len(failed_jobs), self.id)
+            if log.isEnabledFor(LOG_DEBUG_LEVEL):
+                list(
+                    map(lambda job: self.log_debug_incident(job["job_id"]), failed_jobs)
+                )
+
+        if not results:
+            raise NoResultsError(
```
mimi1vx, line 166: is this exception anywhere caught? and resolved?, btw aggregates could be scheduled before any results are available ..

mimi1vx (changes requested):
few questions ..

openqabot/types/incident.py
Martchus: But it makes sense as it avoid running the code at all unless debug level has been enabled.

openqabot/types/incident.py
Martchus: Yes, if the result is not used then it makes no sense to use map.

mimi1vx: > Yes, if the result is not used then it makes no sense to use map.

in reality `map` `reduce` and `filter` and `lambda` should be avoided in standard python

Martchus (changes requested):
There are still pending questions. If that was an opt-in we would be able to merge it more easily (without everything being perfect but we could try it out in production without changing the deployment).

openqabot/types/incident.py
kraih: "If it can be a `for` loop, it probably should be", seems to be true for most programming languages. 😁

openqabot/types/incident.py
```diff
@@ -144,3 +148,60 @@ def contains_package(self, requires: List[str]) -> bool:
...
+            for res in results
+            if res["status"] not in ("passed")
+            and not has_ignored_comment(res["job_id"], self.id)
+        ]
+
+
+# pylint: disable=fixme
+# TODO:
+#   - move to utils.py or a better place
+#   - remove almost duplicated code from Approver.is_job_marked_acceptable_for_incident
```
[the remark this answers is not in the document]
foursixnine: > can we maybe just do them before merging this PR?

I'll leave it for a follow-up when addressing the rest of the changes, as that would need a bigger refactor, due to the incidents class not being used in the approver. thingie.

Martchus:
@Mergify-io rebase

okurz:
I also called the application locally but found no relevant logs are output:

```
./bot-ng.py --configs metadata --singlearch metadata/bot-ng/singlearch.yml -t 1234 --debug --dry updates-run
```

Possibly the relevant steps are not executed due to dry-run or something else preventing the evaluation of what products to trigger:

```
2024-03-06 12:34:58 INFO     Bot schedule starts now
2024-03-06 12:34:58 INFO     Project SUSE:Maintenance:17818 has empty channels - check incident in SMELT
2024-03-06 12:34:58 INFO     Project SUSE:Maintenance:17958 has empty channels - check incident in SMELT
…
2024-03-06 12:35:28 INFO     Project SUSE:Maintenance:18479 can't calculate repohash  .. skipping
2024-03-06 12:35:28 INFO     Project SUSE:Maintenance:18485 has empty channels - check incident in SMELT
…
2024-03-06 12:35:59 INFO     Project SUSE:Maintenance:19102 can't calculate repohash  .. skipping
2024-03-06 12:35:59 INFO     Project SUSE:Maintenance:24734 has empty channels - check incident in SMELT
…
2024-03-06 12:36:31 INFO     Project SUSE:Maintenance:28369 can't calculate repohash  .. skipping
…
2024-03-06 12:37:02 INFO     Project SUSE:Maintenance:28667 can't calculate repohash  .. skipping
…
2024-03-06 12:37:32 INFO     Project SUSE:Maintenance:28784 can't calculate repohash  .. skipping
2024-03-06 12:37:32 INFO     Project SUSE:Maintenance:29248 has empty packages - check incident in SMELT
2024-03-06 12:37:32 INFO     Project SUSE:Maintenance:30071 has empty channels - check incident in SMELT
2024-03-06 12:37:41 INFO     Project SUSE:Maintenance:31645 has empty channels - check incident in SMELT
2024-03-06 12:37:42 INFO     Project SUSE:Maintenance:32086 has empty channels - check incident in SMELT
2024-03-06 12:37:59 INFO     Project SUSE:Maintenance:32288 has empty channels - check incident in SMELT
2024-03-06 12:38:09 INFO     Project SUSE:Maintenance:32462 has empty channels - check incident in SMELT
2024-03-06 12:38:33 INFO     Project SUSE:Maintenance:32613 has empty channels - check incident in SMELT
2024-03-06 12:39:13 INFO     Project SUSE:Maintenance:32782 has empty channels - check incident in SMELT
2024-03-06 12:39:33 INFO     Project SUSE:Maintenance:32808 has empty channels - check incident in SMELT
2024-03-06 12:39:43 INFO     Project SUSE:Maintenance:32824 has empty channels - check incident in SMELT
2024-03-06 12:40:16 INFO     Project SUSE:Maintenance:32877 has empty channels - check incident in SMELT
…
2024-03-06 12:40:47 INFO     Project SUSE:Maintenance:32879 can't calculate repohash  .. skipping
2024-03-06 12:40:47 INFO     … incidents loaded from qem dashboard
2024-03-06 12:40:47 DEBUG    Skipping invalid config metadata/.gitlab-ci.yml
2024-03-06 12:40:47 DEBUG    Skipping invalid config metadata/products.yml
2024-03-06 12:40:47 INFO     Starting bot mainloop
2024-03-06 12:40:47 INFO     Would trigger 0 products in openQA
2024-03-06 12:40:47 INFO     End of bot run
```

PR #156, opened 2024-01-30
Enforce a pylint style of 10.0 with fine-grained exceptions

okurz:


baierjan (approved):

openqabot/approver.py
```diff
@@ -198,7 +198,7 @@ def osc_approve(inc: IncReq) -> bool:
             )
         except HTTPError as e:
             return _handle_http_error(e, inc)
-        except Exception as e:
+        except Exception as e:  # pylint: disable=broad-except
```
perlpunk, line 201: These are a bit annoying. Can't this be configured to be ignored in general?

perlpunk (approved):
Approving anyway

openqabot/loader/config.py
```diff
@@ -25,7 +25,7 @@ def load_metadata(
     for p in get_yml_list(path):
         try:
             data = loader.load(p)
-        except Exception as e:
+        except Exception as e:  # pylint: disable=broad-except
```
Martchus, line 28: Maybe we should actually make the exception type more specific?

Martchus (approved):

PR #157, opened 2024-01-30
wip -- why does flake8 not fail CI checks?

okurz:


okurz:
I guess because I have multiple flake8 plugins locally. So I will take a look into adding some more in requirements-dev.txt

okurz:
ok, https://github.com/openSUSE/qem-bot/actions/runs/7709081576/job/21009493899?pr=157#step:6:1 fails now as expected. I will include the relevant changes in another PR

PR #158, opened 2024-01-30
Add more flake8 checks which are already handled

okurz:


baierjan (approved):

openqabot/loader/qem.py
```diff
@@ -115,21 +115,19 @@ def get_incident_settings_data(token: Dict[str, str], number: int) -> Sequence[D
...
-                number,
-                d["id"],
-                d["flavor"],
-                d["arch"],
-                d["settings"]["DISTRI"],
-                d["version"],
-                d["settings"]["BUILD"],
-                "",
-            )
+    ret = [
```
Martchus, line 118: Maybe also remove the variable completely and return the rvalue directly?

okurz: Sure, done

Martchus (approved):

PR #162, opened 2024-02-01
Extract dashboard helper methods

okurz:
This commit extracts requests get/put/patch calls to its own module so
that the common URL construction using the dashboard URL is not repeated
over and over again.


kalikiana (approved):
I like it. Little steps towards more maintainable code

Martchus (approved):

PR #164, opened 2024-02-08
Filter embargoed also by Flavor

mimi1vx:
https://progress.opensuse.org/issues/153334

### DIFF
--- a/openqabot/types/baseconf.py
+++ b/openqabot/types/baseconf.py
@@ -27,8 +27,10 @@ class BaseConf(metaclass=ABCMeta):
     def normalize_repos(config):
         pass
 
-    def filter_embargoed(self) -> bool:
-        return any(k.startswith("PUBLIC") for k in self.settings.keys())
+    def filter_embargoed(self, flavor: str) -> bool:
+        return any(k.startswith("PUBLIC") for k in self.settings) or any(
+            flavor.startswith(s) for s in ("Azure", "EC2", "GCE")
+        )
 
     @staticmethod
     def set_obsoletion(settings: dict) -> None:
--- a/tests/test_aggregate.py
+++ b/tests/test_aggregate.py
@@ -8,7 +8,7 @@ def test_aggregate_constructor():
     needed by the constructor?
     """
     config = {}
-    config["FLAVOR"] = None
+    config["FLAVOR"] = "None"
     config["archs"] = None
     config["test_issues"] = {}
     acc = Aggregate("", None, config)
@@ -19,7 +19,7 @@ def test_aggregate_printable():
     Try the printable
     """
     config = {}
-    config["FLAVOR"] = None
+    config["FLAVOR"] = "None"
     config["archs"] = None
     config["test_issues"] = {}
     acc = Aggregate("hello", None, config)
@@ -32,7 +32,7 @@ def test_aggregate_call():
     needed by the callable?
     """
     config = {}
-    config["FLAVOR"] = None
+    config["FLAVOR"] = "None"
     config["archs"] = []
     config["test_issues"] = {}
     acc = Aggregate("", None, config)
@@ -69,7 +69,7 @@ def test_aggregate_call_with_archs(request_mock):
     Configure an archs to enter in the function main loop
     """
     my_config = {}
-    my_config["FLAVOR"] = None
+    my_config["FLAVOR"] = "None"
     my_config["archs"] = ["ciao"]
     my_config["test_issues"] = {}
     acc = Aggregate("", settings={}, config=my_config)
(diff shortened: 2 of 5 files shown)


tests/test_baseconf.py
```diff
@@ -34,12 +34,13 @@ def test_baseconf_init(baseconf_gen):
...
     baseconf_gen.settings["PUBLIC_CLOUD_SOMETHING"] = ""
 
-    assert baseconf_gen.filter_embargoed()
+    assert baseconf_gen.filter_embargoed("None")
 
     baseconf_gen.settings = {}
 
-    assert baseconf_gen.filter_embargoed() is False
+    assert baseconf_gen.filter_embargoed("Noone") is False
+    assert baseconf_gen.filter_embargoed("Azure-test")
```
okurz, line 46: I guess this line is the only real test addition for the new change of behaviour, right?

mimi1vx:
@asmorodskyi

tests/test_baseconf.py
mimi1vx: yes, real change is only `or any(
            flavor.startswith(s) for s in ("Azure", "EC2", "GCE")` , rest is only call semantic and remove one layer of indentation in aggregates :D

asmorodskyi (approved):
LGTM

tests/test_aggregate.py
```diff
@@ -8,7 +8,7 @@ def test_aggregate_constructor():
     needed by the constructor?
     """
     config = {}
-    config["FLAVOR"] = None
+    config["FLAVOR"] = "None"
```
baierjan, line 11: I guess `FLAVOR` is actually a key name from metadata so in reality it cannot be `None` so the change to string make sense.

okurz (approved):

baierjan (approved):

openqabot/types/baseconf.py
```diff
@@ -27,8 +27,10 @@ def __call__(
     def normalize_repos(config):
         pass
 
-    def filter_embargoed(self) -> bool:
-        return any(k.startswith("PUBLIC") for k in self.settings.keys())
+    def filter_embargoed(self, flavor: str) -> bool:
+        return any(k.startswith("PUBLIC") for k in self.settings) or any(
+            flavor.startswith(s) for s in ("Azure", "EC2", "GCE")
```
mpagot, line 32: Is using of  `startswith`  too restrictive?

It will works as expected but I'm not sure there's any "official rule" that say that name of the cloud provider has to be used at the beginning of the `FLAVOR` name.

More deeply in the PC test lib there's a more coded convention based on `PUBLIC_CLOUD_PROVIDER` and not on `FLAVOR` 

There's a PC lib setting named `PUBLIC_CLOUD_PROVIDER` that has a coded naming convention: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/2e0b43a6f0cea5d2e8856e2a35d8bb23bb625325/lib/publiccloud/basetest.pm#L33

Then, at least for the QE-SAP jobs the connection between FLAVOR and PUBLIC_CLOUD_PROVIDER happens at JobGroup level (have a look for the setting name in JobGroup file `openQA/yaml_job_groups/maintenance/maintenance_sle15_sp5_sap.yml` in repo `openqa_ha_sap`

openqabot/types/baseconf.py
```diff
@@ -27,8 +27,10 @@ def __call__(
     def normalize_repos(config):
         pass
 
-    def filter_embargoed(self) -> bool:
-        return any(k.startswith("PUBLIC") for k in self.settings.keys())
+    def filter_embargoed(self, flavor: str) -> bool:
+        return any(k.startswith("PUBLIC") for k in self.settings) or any(
```
mpagot, line 31: I think here it is safe to have `startswith` but `PUBLIC` can be extended to `PUBLIC_CLOUD`

mpagot:
I'd like to propose a possible different approach to drastically simplify logic in `filter_embargoed`. What about to add a field in the metadata to opt-in the check?

```
    Azure-SAP-PAYG-Incidents:
      embargo_sensible: true       <<<-----
      aggregate_job: false
      archs:
        - x86_64
      issues: *test_issues_list
      required_issues: *required_list
      packages: *whitelist_ha_hana
      params_expand:
        PUBLIC_CLOUD_IMAGE_ID: 'suse:sles-sap-15-sp5:gen2:latest'
```

openqabot/types/baseconf.py
asmorodskyi: the problem which @mimi1vx  was trying to address here that currently https://gitlab.suse.de/qa-maintenance/metadata does not have `PUBLIC_CLOUD_PROVIDER` for SAP related products . And if you will add it there than you won't need second part at all . Because `any(k.startswith("PUBLIC") for k in self.settings)` would already return true for `PUBLIC_CLOUD_PROVIDER` variable no matter what value it has

PR #165, opened 2024-02-20
Parallelize job posting

mimi1vx:


okurz (changes requested):
s/Paralelize/Parallelize/

The commit message could have a bit more details, e.g. what issue is fixed, why this needs to change, to which versions of which product it applies, link to a bug or a fate entry, the choices you made, etc. Also see https://commit.style/ or http://chris.beams.io/posts/git-commit/ as a helpful guide how to write good commit messages.

openqabot/openqabot.py
```diff
@@ -75,6 +78,10 @@ def __call__(self):
                 else:
                     self.post_qem(job["qem"], job["api"])
 
+            with ThreadPoolExecutor() as executor:
+                tasks = [executor.submit(poster, job) for job in post]
+                wait(tasks)
```
Martchus, line 83:
```suggestion
                wait([executor.submit(poster, job) for job in post])
```

Martchus (approved):

PR #168, opened 2024-02-21
 Approve jobs if at least older jobs passed

michaelgrifalconi:
https://progress.opensuse.org/issues/97118

- if aggregate update failed, do not give up immediately
- look at openQA previous jobs, if present, green, not too old and including the update under test: ignore that failure

This is to reduce the impact of one test being broken one day, a different test another day and the update not being approved even if combined result give all green, just not at the same time.

I did not touch the tests yet. Before investing more I would like to discuss about the new logic and it's implementation.

### DIFF
--- a/openqabot/approver.py
+++ b/openqabot/approver.py
@@ -5,6 +5,7 @@ from functools import lru_cache
 from logging import getLogger
 from typing import List
 from urllib.error import HTTPError
+from datetime import timedelta, datetime
 import re
 
 import osc.conf
@@ -15,7 +16,7 @@ from openqabot.errors import NoResultsError
 from openqabot.openqa import openQAInterface
 from openqabot.dashboard import get_json
 
-from . import OBS_GROUP, OBS_MAINT_PRJ, OBS_URL, QEM_DASHBOARD
+from . import OBS_GROUP, OBS_MAINT_PRJ, OBS_URL, QEM_DASHBOARD, OLDEST_APPROVAL_JOB_DAYS
 from .loader.qem import (
     IncReq,
     JobAggr,
@@ -131,11 +132,96 @@ class Approver:
             pass
         return False
 
-    def job_acceptable(self, inc: int, res) -> bool:
+    @lru_cache(maxsize=512)
+    def validate_job_qam(self, job: int) -> bool:
+        # Check that valid test result is still present in the dashboard (see https://github.com/openSUSE/qem-dashboard/pull/78/files) to avoid using results related to an old release request
+        qam_data = get_json("api/jobs/" + str(job), headers=self.token)
+        if not qam_data:
+            return False
+        if "error" in qam_data:
+            log.info(
+                "Cannot find job %s in the dashboard database to make sure it is valid",
+                job,
+            )
+            return False
+        if qam_data["status"] != "passed":
+            log.info(
+                'Job %s is not recorded as "passed" in the qam-dashboard database',
+                job,
+            )
+            return False
+        return True
+
+    @lru_cache(maxsize=512)
+    def was_ok_before(self, failed_job_id: int, inc: int) -> bool:
+        # We need a considerable amount of older jobs, since there could be many failed manual restarts from same day
+        older_jobs = self.client.get_older_jobs(failed_job_id, 20)
+        if older_jobs == []:
+            log.info("Cannot find older jobs for %s", failed_job_id)
+            return False
+
+        current_build = older_jobs["data"][0]["build"][:-2]
+        current_build_date = datetime.strptime(current_build, "%Y%m%d")
+
+        # Use at most X days old build. Don't go back in time too much to reduce risk of using invalid tests
+        oldest_build_usable = current_build_date - timedelta(
+            days=OLDEST_APPROVAL_JOB_DAYS
+        )
+
+        regex = re.compile(r"(.*)Maintenance:/%s/(.*)" % inc)
+        # Skipping first job, which was the current one
+        for i in range(1, len(older_jobs["data"])):
+            job = older_jobs["data"][i]
+            job_build = job["build"][:-2]
+            job_build_date = datetime.strptime(job_build, "%Y%m%d")
+
+            # Check the job is not too old
+            if job_build_date < oldest_build_usable:
+                log.info(
+                    "Cannot ignore aggregate failure %s for update %s because: Older jobs are too old to be considered"
+                    % (failed_job_id, inc)
+                )
+                return False
+
+            if job["result"] != "passed" and job["result"] != "softfailed":
+                continue
+
+            # Check the job contains the update under test
+            job_settings = self.client.get_single_job(job["id"])
+            if not regex.match(str(job_settings)):
+                # Likely older jobs don't have it either. Giving up
+                log.info(
+                    "Cannot ignore aggregate failure %s for update %s because: Older passing jobs do not have update under test"
+                    % (failed_job_id, inc)
+                )
+                return False
+
+            if not self.validate_job_qam(job["id"]):
+                log.info(
+                    "Cannot ignore failed aggregate %s using %s for update %s because is not present in qem-dashboard. It's likely about an older release request"
+                    % (failed_job_id, job["id"], inc)
+                )
+                return False
+
+            log.info(
+                "Ignoring failed aggregate %s and using instead %s for update %s"
+                % (failed_job_id, job["id"], inc)
+            )
+            return True
+
+        log.info(
+            "Cannot ignore aggregate failure %s for update %s because: Older usable jobs did not succeed. Run out of jobs to evaluate."
+            % (failed_job_id, inc)
+        )
+        return False
+
+    def job_acceptable(self, inc: int, api: str, res) -> bool:
         """
         Check each job if it is acceptable for different reasons.
 
         Keep jobs marked as acceptable for one incident by openQA comments.
+
+        Keep jobs marked as acceptable if are aggregate and were ok in the previous days.
         """
         if res["status"] == "passed":
             return True
@@ -143,6 +229,13 @@ class Approver:
         if self.is_job_marked_acceptable_for_incident(res["job_id"], inc):
             log.info(
                 "Ignoring failed job %s for incident %s due to openQA comment", url, inc
+            )
+            return True
+        if api == "api/jobs/update/" and self.was_ok_before(res["job_id"], inc):
+            log.info(
+                "Ignoring failed aggregate job %s for incident %s due to older eligible openQA job being ok",
+                url,
+                inc,
             )
             return True
         log.info("Found failed, not-ignored job %s for incident %s", url, inc)
@@ -156,7 +249,7 @@ class Approver:
                 "Job setting %s not found for incident %s"
                 % (str(job_aggr.id), str(inc))
             )
-        return all(self.job_acceptable(inc, r) for r in results)
+        return all(self.job_acceptable(inc, api, r) for r in results)
 
     def get_incident_result(self, jobs: List[JobAggr], api: str, inc: int) -> bool:
         res = False
--- a/openqabot/openqa.py
+++ b/openqabot/openqa.py
@@ -105,3 +105,28 @@ class openQAInterface:
         return (
             ret[0]["parent_id"] == DEVELOPMENT_PARENT_GROUP_ID if ret else True
         )  # ID of Development Group
+
+    @lru_cache(maxsize=256)
+    def get_single_job(self, job_id: int):
+        ret = None
+        try:
+            ret = self.openqa.openqa_request(
+                "GET",
+                "jobs/%s" % job_id,
+            )["job"]
+        except RequestError as e:
+            log.exception(e)
+        return ret
+
+    @lru_cache(maxsize=256)
+    def get_older_jobs(self, job_id: int, limit: int):
+        ret = []
+        try:
+            ret = self.openqa.openqa_request(
+                "GET",
+                "/tests/%s/ajax?previous_limit=%s&next_limit=0" % (job_id, limit),
+                retries=self.retries,
+            )
+        except RequestError as e:
+            log.exception(e)
+        return ret
--- a/openqabot/__init__.py
+++ b/openqabot/__init__.py
@@ -9,3 +9,4 @@ OBS_GROUP = "qam-openqa"
 OPENQA_URL = "openqa.suse.de"
 DEVELOPMENT_PARENT_GROUP_ID = 9
 DOWNLOAD_BASE = "http://download.suse.de/ibs/SUSE:/Maintenance:/"
+OLDEST_APPROVAL_JOB_DAYS = 6
--- a/tests/test_approve.py
+++ b/tests/test_approve.py
@@ -24,6 +24,57 @@ _namespace = namedtuple(
 openqa_instance_url = urlparse("http://instance.qa")
 
 
+@pytest.fixture(scope="function")
+def fake_responses_for_unblocking_incidents_via_older_ok_result(request):
+    responses.add(
+        responses.GET,
+        f"{QEM_DASHBOARD}api/jobs/update/20005",
+        json=[
+            {"job_id": 100000, "status": "passed"},
+            {"job_id": 100002, "status": "failed"},
+            {"job_id": 100003, "status": "passed"},
+        ],
+    )
+    add_two_passed_response()
+    responses.add(
+        responses.GET,
+        url="http://instance.qa/api/v1/jobs/100002/comments",
+        json=[{"text": "@review:acceptable_for:incident_555:foo"}],
+    )
+    responses.add(
+        responses.GET,
+        re.compile(r"http://instance.qa/tests/.*/ajax\?previous_limit=.*&next_limit=0"),
+        json={
+            "data": [
+                {"build": "20240115-1", "id": 100002, "result": "failed"},
+                {"build": "20240114-1", "id": 100004, "result": "failed"},
+                {"build": "20240113-1", "id": 100005, "result": "softfailed"},
+            ]
+        },
+    )
+    responses.add(
+        responses.GET,
+        re.compile(f"http://instance.qa/api/v1/jobs/.*"),
+        json={
+            "job": {
+                "settings": {
+                    "BASE_TEST_REPOS": "http://download.suse.de/ibs/SUSE:/Maintenance:/1111/SUSE_Updates_SLE-Module-Basesystem_15-SP5_x86_64/,http://download.suse.de/ibs/SUSE:/Maintenance:/%s/SUSE_Updates_SLE-Module-Basesystem_15-SP5_x86_64/"
+                    % request.param
+                }
+            }
+        },
+    )
+
+
+@pytest.fixture(scope="function")
+def fake_openqa_older_jobs_api():
+    responses.add(
+        responses.GET,
+        re.compile(r"http://instance.qa/tests/.*/ajax\?previous_limit=.*&next_limit=0"),
+        status=404,
+    )
+
+
 def add_two_passed_response():
     responses.add(
         responses.GET,
@@ -404,7 +455,11 @@ def test_one_incident_failed(
 @responses.activate
 @pytest.mark.parametrize("fake_qem", [("NoResultsError isn't raised")], indirect=True)
 def test_one_aggr_failed(
-    fake_qem, fake_openqa_comment_api, fake_responses_updating_job, caplog
+    fake_qem,
+    fake_openqa_comment_api,
+    fake_responses_updating_job,
+    fake_openqa_older_jobs_api,
+    caplog,
 ):
     caplog.set_level(logging.DEBUG, logger="bot.approver")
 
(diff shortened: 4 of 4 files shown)


openqabot/approver.py
```diff
@@ -131,6 +132,53 @@ def is_job_marked_acceptable_for_incident(self, job_id: int, inc: int) -> bool:
             pass
         return False
 
+    @lru_cache(maxsize=512)
+    def was_green_before(self, job_id: int, inc: int) -> bool:
```
okurz, line 136:
```suggestion
    def was_ok_before(self, job_id: int, inc: int) -> bool:
```

openqabot/approver.py
```diff
@@ -131,6 +132,53 @@ def is_job_marked_acceptable_for_incident(self, job_id: int, inc: int) -> bool:
...
+        log.info(current_build)
+        current_build_date = date(
+            int(current_build[0:4]),
+            int(current_build[4:6]),
+            int(current_build[6:8])
+            )
+        log.info("Failed job date: %s",current_build_date)
+
+        # Use at most X days old build. Don't go back in time too much to reduce risk of using invalid tests
+        oldest_build_usable = current_build_date - timedelta(days=6)
```
okurz, line 151: ok, that makes sense in general but why 6 then?

I suggest to put that magic number into a configuration variable or a constant with a good name and the according comment for the value selected.

openqabot/approver.py
```diff
@@ -131,6 +132,53 @@ def is_job_marked_acceptable_for_incident(self, job_id: int, inc: int) -> bool:
...
+                        log.info("Not found an older passing job with update %s under test present. Giving up" % inc)
+                        return False
+
+                    log.info("Found a valid job result %s for update %s" % (job['id'], inc))
+                    return True
+            else:
+                log.info("Job is too old. Giving up")
+                return False
+        # Run out of jobs. Giving up
+        log.info("Run out of jobs to try. Giving up")
```
okurz, line 179: The comment is redundant due to the log message in the next line. I also often prefer having log messages over comments.

```suggestion
        log.info("Ran out of jobs to try. Giving up")
```

okurz (changes requested):
Neat idea! I wonder if this can actually work because so far the approver workflow does AFAIK never contact openQA directly.

openqabot/approver.py
michaelgrifalconi: Same as the decision of asking for 20 older jobs to openQA and not 15 or 25,  I don't see a reasonable number but a "bunch of tests" to be collected.

For the days: more days means more chance to find all green test at some point (but useless if update request is new). I considered reasonable a week, even though more might be useful (some updates might be stuck for quite some time).

 But more days means also more risk of taking a "unfaithful" test, where in meantime the update request was changed this opens the pandora's box since it's an issue that is already present, and incident would get retriggered anyway so they provide some safety. 
 
 Anyway, I agree on making this number more visible, can move it [here maybe](https://github.com/openSUSE/qem-bot/blob/master/openqabot/__init__.py)

openqabot/approver.py
michaelgrifalconi: I have no strong opinion on names, will change as requested!

michaelgrifalconi:
If the code it's looking good enough, I will move on adapting the tests!

openqabot/approver.py
```diff
@@ -131,6 +132,52 @@ def is_job_marked_acceptable_for_incident(self, job_id: int, inc: int) -> bool:
             pass
         return False
 
+    @lru_cache(maxsize=512)
+    def was_ok_before(self, job_id: int, inc: int) -> bool:
+        log.info("Current aggregate test %s failed. Will look at older jobs" % job_id)
```
okurz, line 137: Depending on how many such log messages appear in a usual run consider to make that debug. Also I would phrase it

```suggestion
        log.info("Current aggregate test %s failed. Looking up job history" % job_id)
```

openqabot/approver.py
```diff
@@ -131,6 +132,52 @@ def is_job_marked_acceptable_for_incident(self, job_id: int, inc: int) -> bool:
...
+        older_jobs = self.client.get_older_jobs(job_id, 20)
+
+        current_build = older_jobs['data'][0]['build'].replace('-1','')
+        log.info(current_build)
+        current_build_date = date(
+            int(current_build[0:4]),
+            int(current_build[4:6]),
+            int(current_build[6:8])
+            )
+        log.info("Failed job date: %s",current_build_date)
```
okurz, line 148: I don't understand this log message. What should it tell me?

okurz, line 148: I don't understand this log message. What should it tell me?

okurz (changes requested):
As you stated you already tested manually can you provide the logs from such run so that we can see the flow of execution from the log messages?

openqabot/approver.py
michaelgrifalconi: Resolving as duplicate of same conversation below

openqabot/approver.py
```diff
@@ -131,6 +132,52 @@ def is_job_marked_acceptable_for_incident(self, job_id: int, inc: int) -> bool:
...
         return False
 
+    @lru_cache(maxsize=512)
+    def was_ok_before(self, job_id: int, inc: int) -> bool:
+        log.info("Current aggregate test %s failed. Will look at older jobs" % job_id)
+        # We need a considerable amount of older jobs, since there could be many failed manual restarts from same day
+        older_jobs = self.client.get_older_jobs(job_id, 20)
+
+        current_build = older_jobs['data'][0]['build'].replace('-1','')
+        log.info(current_build)
```
michaelgrifalconi, line 142: Will remove this as well

openqabot/approver.py
michaelgrifalconi: I will remove that, was used to make sure job date and older job date were handled correctly

michaelgrifalconi:
>As you stated you already tested manually can you provide the logs from such run so that we can see the flow of execution from the log messages?

https://gist.github.com/michaelgrifalconi/4d07eee0197c929db7c2a11b85759edb


Mind that stuff like 
```
2024-02-22 08:13:59 INFO     20240221
2024-02-22 08:13:59 INFO     Failed job date: 2024-02-21
```
Will be removed as I mention on the other conversations

openqabot/approver.py
michaelgrifalconi: will remove this and rethink logging for this thing, reducing duplicated informations (if later on there is a log for each case (job found and ignore failure, job not found and reason) there is no need of this log announcing the start of this search

michaelgrifalconi:
New log with less frequent/redundant and more descriptive logs: https://gist.github.com/michaelgrifalconi/41b6451b36cdfb87814b9ce9635a9459

Would not reduce too much logging, since it as something you would look only in case of problems, and do not disturb anyone (of course as long as it's not so huge to be expensive on resources or just clutter when debugging)

openqabot/approver.py
```diff
@@ -131,6 +132,49 @@ def is_job_marked_acceptable_for_incident(self, job_id: int, inc: int) -> bool:
             pass
         return False
 
+    @lru_cache(maxsize=512)
+    def was_ok_before(self, failed_job_id: int, inc: int) -> bool:
+        # We need a considerable amount of older jobs, since there could be many failed manual restarts from same day
+        older_jobs = self.client.get_older_jobs(failed_job_id, 20)
```
asmorodskyi, line 138: I think we need to add sorting here just to be sure that jobs are ordered in right way

openqabot/approver.py
```diff
@@ -131,6 +132,49 @@ def is_job_marked_acceptable_for_incident(self, job_id: int, inc: int) -> bool:
...
+
+        current_build = older_jobs['data'][0]['build'].replace('-1','')
```
asmorodskyi, line 140: at some point of time we had two triggers during a day . Is it fully disabled currently ?  Is there chance that we will get builds ending with `-2` ?

openqabot/approver.py
```diff
@@ -131,6 +132,49 @@ def is_job_marked_acceptable_for_incident(self, job_id: int, inc: int) -> bool:
...
+                    job_settings = self.client.get_single_job(job['id'])
+                    if not regex.match(str(job_settings)):
+                        # Likely older jobs don't have it either. Giving up
+                        log.info("Cannot ignore aggregate failure %s for update %s because: Older passing jobs do not have update under test" %(failed_job_id, inc))
+                        return False
+
+                    log.info("Ignoring failed aggregate %s and using instead %s for update %s" % (failed_job_id, job['id'], inc))
+                    return True
+            else:
+                log.info("Cannot ignore aggregate failure %s for update %s because: Older usable jobs did not succed and remaining jobs are too old to be considered" %(failed_job_id, inc))
```
b10n1k, line 173: s/succed/succeed/

openqabot/approver.py
```diff
@@ -131,6 +132,49 @@ def is_job_marked_acceptable_for_incident(self, job_id: int, inc: int) -> bool:
...
+                return False
+        log.info("Cannot ignore aggregate failure %s for update %s because: Older usable jobs did not succed and run out of jobs to try." %(failed_job_id, inc))
```
b10n1k, line 175: s/succed/succeed/

openqabot/approver.py
michaelgrifalconi: I think they do come sorted already, see https://openqa.suse.de/tests/13576926/ajax as example.
Anyway, a follow-up plan is to contribute on openQA an official API for this, instead of stealing the one used for the NextPrevious table (even though is supposed to be pretty stable)

openqabot/approver.py
michaelgrifalconi: We don't have that anymore. We opted for more time to review failures instead of more tests left unreviewed,  In any case we don't care about that id, since we look at build date to make sure build is not too old than specified number of days and we also check that the selected job has the update under test.
We could improve the code to remove the last 2 chars in the date to be compatible, instead of looking for "-1"

okurz (changes requested):
My feedback after the QE Tools workshop about this topic:
1. What are alternatives that we can consider, how about again enabling automatic retry?
2. How about shifting more aggregate tests to incidents, especially the tricky ones more prone to failing?
3. There is no feedback to openQA so it will make it even harder for reviewers to find which jobs are actually blocking -> Hence my suggestion is instead to cover this logic in a separate application which simply writes "approvable_for" comments on openQA directly. This way there is direct feedback to reviewers same as to qem-bot.

openqabot/approver.py
```diff
@@ -145,6 +202,13 @@ def job_acceptable(self, inc: int, res) -> bool:
                 "Ignoring failed job %s for incident %s due to openQA comment", url, inc
             )
             return True
+        if api == "api/jobs/update/" and self.was_green_before(res["job_id"], inc):
+            log.info(
+                "Ignoring failed aggregate job %s for incident %s due to openQA older eligible job being green",
```
Martchus, line 207:
```suggestion
                "Ignoring failed aggregate job %s for incident %s due to older eligible openQA job being green",
```

okurz:
Before too much developent+review effort is invested please keep https://github.com/openSUSE/qem-bot/pull/168#pullrequestreview-1897640058 in mind which I think we agreed upon to follow up with.

michaelgrifalconi:
>     1. What are alternatives that we can consider, how about again enabling automatic retry?
I agree on making sure automatic retries should be set at least at RETRY=1 globally. Where would be a good place to set that? Medium types?
This is anyhow separate and does not substitute this PR. This is for reproducible issues (either test or product issues, unrelated to the update that shall be approved and was green the day before).

>     2. How about shifting more aggregate tests to incidents, especially the tricky ones more prone to failing?
Sure, there are tickets open for that. Still this is a different topic and I believe both things do help.


>     3. There is no feedback to openQA so it will make it even harder for reviewers to find which jobs are actually blocking -> Hence my suggestion is instead to cover this logic in a separate application.[...]
No, introducing a new thing that changes the behavior of something, from a different place makes it even more frustrating for an engineer to understand "what is happening and why".
I would agree on moving out the **entire approval logic** and handle it in a simpler way (no hand-crafted caching and syncing of data) but it's a future topic.

Right now this is **the smallest tweak possible to increase the quality of life for maintenance test developers**.


About the visibility issue, I think that all current test failure should be fixed/softfailed at some point, since they surely block updates that were not tested the day before(like newly released one). 
In addition, this is just about aggregates, and they run every day. I see no reason why anyone would spend time fixing a failure from the previous day instead of focusing on the present ones.

I see that would still be nice to have visibility of what is really blocking an update and what could be ignored for a specific update request (by "approvable for comment"). Too bad that the dashboard does not show that either AFAIK.

For this topic I would either:
- switch from ignoring the failure to writing a comment "Approvable for" (before looking at comments and ignore it)
- merge it as it is to get the benefit of getting stuff approved, and then start discussing on a better solution, like the one in the line above or something else

I have no strong opinions for either option, as long as we don't spend too much time discussing on cosmetics and also new cosmetics changes do not require me to do more rebase of this PR (which i believe to be more important than suddenly changing code styles)

okurz:
> switch from ignoring the failure to writing a comment "Approvable for" (before looking at comments and ignore it)
Yes, that could be done in here as well. I just think that this is easier to implement and also run in a separate script. Or, maybe a compromise, run that pre-approval with writing the comments in a separate command within qem-bot? I am fine with either approach, just pointing out ideas. You can choose and we will support you either way :)

asmorodskyi:
DISCLAIMER : discussion below has nothing to do with this certain PR but I am replying here because Oli raised it here . **Outcome of this discussion should not influence on decision about merging a PR**

> 2. How about shifting more aggregate tests to incidents, especially the tricky ones more prone to failing?

@okurz I understand what the intention is behind , but I am not sure that it is easily doable in setup which we have and that effect of such "solution" would be exactly what we want . I will try to explain in example (ofc. I will use my lovely Public Cloud as example, but I will simplify some things to keep it simple enough to prove my point  ) :

So we have aggregates test run which are triggered on update of each and every package and we run there console_tests and we have incidents which are triggered **only** for kernel/grub2/systemd updates and we run LTP tests. If we see that console_tests are unstable and we will decide to move them to incidrents this will mean that we will run console_tests only for kernel/grub2/systemd which will certainly decrease  quantity of failures but also we dramatically decreasing coverage for this test and also probably not to the list of packages which suits well to this certain test. 

I agree that we should decrease coverage for unstable tests, but I don't think that "just move it to incidents " is acceptable option here. To follow this path we need new functionality which allowing to define packages which triggers testing not in flavor level but on test level . Basically we need interface allowing to say "I want to trigger this test for this set of packages"

michaelgrifalconi:
@asmorodskyi I think the best place to continue that discussion is here: https://progress.opensuse.org/issues/106930

openqabot/approver.py
```diff
@@ -145,6 +196,13 @@ def job_acceptable(self, inc: int, res) -> bool:
                 "Ignoring failed job %s for incident %s due to openQA comment", url, inc
             )
             return True
+        if api == "api/jobs/update/" and self.was_ok_before(res["job_id"], inc):
+            log.info(
+                "Ignoring failed aggregate job %s for incident %s due to older eligible openQA job being green",
```
okurz, line 201:
```suggestion
                "Ignoring failed aggregate job %s for incident %s due to older eligible openQA job being ok",
```

okurz (changes requested):
One minor phrasing issue left, rest is fine

michaelgrifalconi:
Considering that: 
- we see more and more situations where this logic would help in real situations
- changing the design to writing comments will require more time and make it more difficult to test dry runs 

I would like to proceed as it is, add the necessary tests and then as soon as it is merged we can start a discussion on how to improve visibility and consider all various options like commenting and or moving to different bot job, etc.

okurz:
> I would like to proceed as it is, add the necessary tests and then as soon as it is merged we can start a discussion on how to improve visibility and consider all various options like commenting and or moving to different bot job, etc.

Yes, I can accept that although be aware that concerns were raised by others than just me about the overall approach. So I see only two points missing before I can approve:
1. The PR is marked as "WIP" and I won't approve before you confirm that the PR is ready to be approved, i.e. remove the WIP
2. Squash the commits that just fixup the original commit

michaelgrifalconi:
Right when I was feeling ready with this, I found a confirmation of the issue caused by using SMELT ids instead of RR ids.
Moving the discussion about it here: https://github.com/openSUSE/qem-bot/issues/174

michaelgrifalconi:
> Right when I was feeling ready with this, I found a confirmation of the issue caused by using SMELT ids instead of RR ids. Moving the discussion about it here: #174

Having lunch brought me some wisdom. We can make use of the same workaround described in the open issue to avoid falling in the rabbit hole of fixing everything in the system at once.
I can add a check [here](https://github.com/openSUSE/qem-bot/pull/168/files#diff-d44cab80bf43c62865043529f6e9883cc12ab37d49f5fb5af9e1bdc7c012dcaaR170) to make sure the selected job is still present in the qem-dashboard. If it's not, then it means it was removed by [qem-dashboard/#78](https://github.com/openSUSE/qem-dashboard/pull/78/files) and should not be used.

openqabot/approver.py
```diff
@@ -132,6 +132,32 @@ def is_job_marked_acceptable_for_incident(self, job_id: int, inc: int) -> bool:
...
+            return False
+        if "error" in qam_data:
+            log.info(
+                "Cannot find job %s in the dashboard database to make sure it is valid",
+                job,
+            )
+            return False
+        if qam_data["status"] != "passed":
+            log.info(
+                'Job %s does not appear "passed" in the qam-dashboard database',
```
okurz, line 155: what do you mean with "appear" here? Why not just "is not recorded as passed"?

michaelgrifalconi: Sure, will change!

openqabot/approver.py
```diff
@@ -132,6 +132,32 @@ def is_job_marked_acceptable_for_incident(self, job_id: int, inc: int) -> bool:
...
+        except Exception as e:  # pylint: disable=broad-except
+            log.info(
+                "Cannot reach the dashboard database to check if job %s is still valid",
+                job,
+            )
+            log.exception(e)
+            return False
+        if "error" in qam_data:
+            log.info(
+                "Cannot find job %s in the dashboard database to make sure it is valid",
```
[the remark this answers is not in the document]
michaelgrifalconi: This happens when the job is not in the dashboard anymore, so this is very important to avoid the situation described in the issue I just opened. When that happens, we do not use that job to ignore the failure because it is likely about a different release request

openqabot/approver.py
```diff
@@ -132,6 +132,32 @@ def is_job_marked_acceptable_for_incident(self, job_id: int, inc: int) -> bool:
             pass
         return False
 
+    @lru_cache(maxsize=512)
+    def validate_job_qam(self, job: int) -> bool:
+        # Check that valid test result is still present in the dashboard (see https://github.com/openSUSE/qem-dashboard/pull/78/files) to avoid using results related to an old release request
+        try:
+            qam_data = get_json("api/jobs/" + str(job), headers=self.token)
+        except Exception as e:  # pylint: disable=broad-except
```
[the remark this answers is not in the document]
michaelgrifalconi: This come from an equivalent call of that function already in the bot, and I took the same exception and pylint comment.
In this case I would like to not fail the entire bot run but just give up trying to approve the update. Do you have any suggestions for which exception to use? We could change that in other places as well

openqabot/approver.py
```diff
@@ -132,6 +132,32 @@ def is_job_marked_acceptable_for_incident(self, job_id: int, inc: int) -> bool:
...
+            log.info(
+                "Cannot reach the dashboard database to check if job %s is still valid",
```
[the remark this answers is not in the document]
michaelgrifalconi: I think it's the same topic and my other reply cover both. I would propose to continue the discussion there !

openqabot/approver.py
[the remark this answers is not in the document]
okurz: I don't know which exception would be relevant here. I would only add exception handling for cases that would actually happen as part of testing.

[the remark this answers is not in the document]
michaelgrifalconi: Fair point. I can remove it and if it crashes due to that we can think about exceptions, if never happens then it was not needed anyway!

michaelgrifalconi:
I think I am ready for the last review. If it's all fine I will do one last consolidation of commits and we can merge! 
By the way, this is what would be approved by the new logic:
```
2024-03-19 08:14:56 INFO     * SUSE:Maintenance:32461:324048
2024-03-19 08:14:56 INFO     * SUSE:Maintenance:32575:324054
2024-03-19 08:14:56 INFO     * SUSE:Maintenance:32795:324057
2024-03-19 08:14:56 INFO     * SUSE:Maintenance:32898:324045
2024-03-19 08:14:56 INFO     * SUSE:Maintenance:32912:324047
2024-03-19 08:14:56 INFO     * SUSE:Maintenance:32934:324023
2024-03-19 08:14:56 INFO     * SUSE:Maintenance:32940:323930
2024-03-19 08:14:56 INFO     * SUSE:Maintenance:32951:323996
2024-03-19 08:14:56 INFO     * SUSE:Maintenance:32956:324041
2024-03-19 08:14:56 INFO     * SUSE:Maintenance:32959:324042
2024-03-19 08:14:56 INFO     * SUSE:Maintenance:32960:324072
2024-03-19 08:14:56 INFO     * SUSE:Maintenance:32967:324105
2024-03-19 08:14:56 INFO     * SUSE:Maintenance:32971:324109
```

okurz (commented):
+1

okurz (commented):
ok, now we need to find a better subject line than "Enhance bot approval logic". How about "Approve jobs if at least older jobs passed"?

openqabot/approver.py
```diff
@@ -131,11 +132,94 @@ def is_job_marked_acceptable_for_incident(self, job_id: int, inc: int) -> bool:
...
+                        "Cannot ignore failed aggregate %s using %s for update %s because is not present in qem-dashboard. It's likley about an older release request"
+                        % (failed_job_id, job["id"], inc)
+                    )
+                    return False
+
+                log.info(
+                    "Ignoring failed aggregate %s and using instead %s for update %s"
+                    % (failed_job_id, job["id"], inc)
+                )
+                return True
```
okurz, line 208: This for-body is now indented a bit too much. Can you extract a method here?

openqabot/approver.py
```diff
@@ -131,11 +132,94 @@ def is_job_marked_acceptable_for_incident(self, job_id: int, inc: int) -> bool:
...
+
+        log.info(
+            "Cannot ignore aggregate failure %s for update %s because: Older usable jobs did not succeed and run out of jobs to try."
```
okurz, line 211:
```suggestion
            "Cannot ignore aggregate failure %s for update %s because: Older usable jobs did not succeed. ran out of jobs to evaluate."
```

openqabot/approver.py
```diff
@@ -131,11 +132,94 @@ def is_job_marked_acceptable_for_incident(self, job_id: int, inc: int) -> bool:
...
+        )
+        return False
+
+    def job_acceptable(self, inc: int, api: str, res) -> bool:
         """
         Check each job if it is acceptable for different reasons.
 
         Keep jobs marked as acceptable for one incident by openQA comments.
+
+        Keep jobs marked as acceptable if are aggregate and were green in the previous days.
```
okurz, line 222:
```suggestion
        Keep jobs marked as acceptable if are aggregate and were ok in the previous days.
```

openqabot/openqa.py
```diff
@@ -105,3 +105,33 @@ def is_devel_group(self, groupid: int) -> bool:
...
+
+    @lru_cache(maxsize=256)
+    def get_single_job(self, job_id: int):
+        ret = None
+        try:
+            ret = self.openqa.openqa_request(
+                "GET",
+                "jobs/%s" % job_id,
+            )["job"]
+        except Exception as e:
```
okurz, line 117: why do we need a broad exception here?

openqabot/openqa.py
```diff
@@ -105,3 +105,33 @@ def is_devel_group(self, groupid: int) -> bool:
...
+    @lru_cache(maxsize=256)
+    def get_older_jobs(self, job_id: int, limit: int):
+        ret = []
+        try:
+            ret = self.openqa.openqa_request(
+                "GET",
+                "/tests/%s/ajax?previous_limit=%s&next_limit=0" % (job_id, limit),
+                retries=self.retries,
+            )
+        except Exception as e:  # pylint: disable=broad-except
```
okurz, line 131: why do we need a broad exception here?

openqabot/openqa.py
michaelgrifalconi: Again, because it was done all over the place already and I kept it consistent. Will remove it as requested.

openqabot/openqa.py
michaelgrifalconi: Again, because it was done all over the place already and I kept it consistent. Will remove it as requested.

openqabot/openqa.py
michaelgrifalconi: https://github.com/openSUSE/qem-bot/blob/master/openqabot/openqa.py#L50
https://github.com/openSUSE/qem-bot/blob/master/openqabot/openqa.py#L73
https://github.com/openSUSE/qem-bot/blob/master/openqabot/openqa.py#L86
https://github.com/openSUSE/qem-bot/blob/master/openqabot/openqa.py#L100

openqabot/approver.py
michaelgrifalconi: I refactored a bit the code, without the need of a new method I removed a nested if to get down of one level of indentation and make it more readable

openqabot/approver.py
```diff
@@ -131,11 +132,96 @@ def is_job_marked_acceptable_for_incident(self, job_id: int, inc: int) -> bool:
...
+                # Likely older jobs don't have it either. Giving up
+                log.info(
+                    "Cannot ignore aggregate failure %s for update %s because: Older passing jobs do not have update under test"
+                    % (failed_job_id, inc)
+                )
+                return False
+
+            if not self.validate_job_qam(job["id"]):
+                log.info(
+                    "Cannot ignore failed aggregate %s using %s for update %s because is not present in qem-dashboard. It's likley about an older release request"
```
okurz, line 201:
```suggestion
                    "Cannot ignore failed aggregate %s using %s for update %s because is not present in qem-dashboard. It's likely about an older release request"
```

okurz (approved):

okurz (changes requested):
oh, sorry. You added another commit. Please squash, then I can approve

michaelgrifalconi:
Yeah I usually try to show what I change instead of force push on every change since it makes more difficult to check and might hide some stuff! Rebased now

michaelgrifalconi:
Thank for the approval! I have no rights to merge so someone else will have to do it :)

kalikiana (approved):

PR #169, opened 2024-02-23
Simplify job evaluation avoiding a double-loop

okurz:
In before there was a for-loop and after that a call to "all". Instead
we can just move the body of the for-loop into the evaluation called by
"all". This is cleaner, avoids the double-loop and also helps to
unindent code preventing more annoying log-message line-wrapping.

Motivated by #168 . @michaelgrifalconi if we accept this PR before #168 then you would need to only change https://github.com/openSUSE/qem-bot/pull/168/files#diff-d44cab80bf43c62865043529f6e9883cc12ab37d49f5fb5af9e1bdc7c012dcaaR195 and additionally pass "api" to the method "job_acceptable"

### DIFF
--- a/openqabot/approver.py
+++ b/openqabot/approver.py
@@ -131,39 +131,32 @@ class Approver:
             pass
         return False
 
+    def job_acceptable(self, inc: int, res) -> bool:
+        """
+        Check each job if it is acceptable for different reasons.
+
+        Keep jobs marked as acceptable for one incident by openQA comments.
+        """
+        if res["status"] == "passed":
+            return True
+        url = "{}/t{}".format(self.client.url.geturl(), res["job_id"])
+        if self.is_job_marked_acceptable_for_incident(res["job_id"], inc):
+            log.info(
+                "Ignoring failed job %s for incident %s due to openQA comment", url, inc
+            )
+            return True
+        log.info("Found failed, not-ignored job %s for incident %s", url, inc)
+        return False
+
     @lru_cache(maxsize=128)
     def get_jobs(self, job_aggr: JobAggr, api: str, inc: int) -> bool:
         results = get_json(api + str(job_aggr.id), headers=self.token)
-
-        # keep jobs explicitly marked as acceptable for this incident by openQA comments
-        for res in results:
-            ok_job = res["status"] == "passed"
-            if ok_job:
-                continue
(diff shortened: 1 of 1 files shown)


openqabot/approver.py
```diff
@@ -131,39 +131,32 @@ def is_job_marked_acceptable_for_incident(self, job_id: int, inc: int) -> bool:
...
-            if ok_job:
-                continue
-            if self.is_job_marked_acceptable_for_incident(res["job_id"], inc):
-                log.info(
-                    "Ignoring failed job %s/t%s for incident %s due to openQA comment",
-                    str(self.client.url.geturl()),
-                    res["job_id"],
-                    inc,
-                )
-                res["status"] = "passed"
```
Martchus, line 150: The previous code sets `res["status"]` here. Is this something we need to keep? I'm just asking because the commit starting with "Simplify …" doesn't make it sound like you have considered that.

michaelgrifalconi:
Actually, while working on my PR, I did change that in a different way that sounded better and simpler and then come back to the original to avoid refactoring unrelated stuff.

I am not a big fan of your proposal, while it makes it a bit cleaner, fees a touch more difficult to follow/debug/understand.

How about getting rid of the all() call, and instead return "acceptable" bool variable which starts as true and become false if there is a failure?
I can provide the change if I did not mange to explain what I mean :)

okurz:
> Actually, while working on my PR, I did change that in a different way that sounded better and simpler and then come back to the original to avoid refactoring unrelated stuff.
> 
> I am not a big fan of your proposal, while it makes it a bit cleaner, fees a touch more difficult to follow/debug/understand.
> 
> How about getting rid of the all() call, and instead return "acceptable" bool variable which starts as true and become false if there is a failure? I can provide the change if I did not mange to explain what I mean :)

Hm, I quite like the functional approach with "all" which is akin to the list comprehension in python. Many people use too much procedural code. I doubt you will convince me with an extended for-loop here :)

openqabot/approver.py
okurz: Yes, I have considered that explicitly. In my understanding the field is only set so that the subsequent "all" evaluates to passed so in a two-step process which I short-cut to a single loop here.

kalikiana: Ah. I missed that. So your approach is definitely cleaner then

Martchus (approved):

PR #173, opened 2024-02-29
Avoid error if priority is absent after f1d4d5c4a6

Martchus:
See https://progress.opensuse.org/issues/156301


Martchus:
This should mitigate the mentioned issue. I'm not sure whether it is all that needs fixing, though.

PR #175, opened 2024-03-21
PoC for incident result sync after AMQP job done message

baierjan:
Reference: https://progress.opensuse.org/issues/154498


okurz:
Obsolete with https://github.com/openSUSE/qem-bot/pull/177

PR #176, opened 2024-04-12
Document proper meaning of OPENQA_URL constant

baierjan:
_I learned this the hard way_ 😅


baierjan:
> Uh,I still don't understand

Ok, help me rephrase. It is about this piece of code https://github.com/openSUSE/qem-bot/blob/00b400050c974de1fc818acec04a57a1919febff/openqabot/openqa.py#L34

And it is apparently just a filter when to skip updating the dashboard with the newly created openQA jobs during scheduling.

Martchus (approved):

okurz (approved):
Don't know how to phrase it better. However your comment clearly states that something might not be what it appears to be so I am sure it helps nevertheless ;)

PR #177, opened 2024-04-15
Allow result sync after AMQP job done messages

Martchus:
See https://progress.opensuse.org/issues/157741

---

This PR adds mainly tests on top of https://github.com/openSUSE/qem-bot/pull/175. What's still missing is handling aggregates, documentation and perhaps a systemd-unit file to run this on the dashboard host.

### DIFF
--- a/openqabot/__init__.py
+++ b/openqabot/__init__.py
@@ -8,6 +8,7 @@ OBS_MAINT_PRJ = "SUSE:Maintenance"
 OBS_GROUP = "qam-openqa"
 DEVELOPMENT_PARENT_GROUP_ID = 9
 DOWNLOAD_BASE = "http://download.suse.de/ibs/SUSE:/Maintenance:/"
+AMQP_URL = "amqps://suse:suse@rabbit.suse.de"
 OLDEST_APPROVAL_JOB_DAYS = 6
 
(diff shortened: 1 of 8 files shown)


baierjan:
In regards of aggregates, that could be probably doing in a similar matter to incidents (or maybe could be even left out entirely as dashboard already should do some synchronization after AMQP events?).

One option to deploy would be a separate CI pipeline with a long job timeout, that way we would have all the relevant logs still on one place. Parallel runs can be disabled via resource group.

Martchus:
> One option to deploy would be a separate CI pipeline with a long job timeout, that way we would have all the relevant logs still on one place. Parallel runs can be disabled via resource group.

This is what it would look like: https://gitlab.suse.de/qa-maintenance/bot-ng/-/merge_requests/68

I've also already created a [schedule](https://gitlab.suse.de/qa-maintenance/bot-ng/-/pipeline_schedules/179/edit?id=179) but kept it inactive.

---

> In regards of aggregates, that could be probably doing in a similar matter to incidents (or maybe could be even left out entirely as dashboard already should do some synchronization after AMQP events?).

Judging by the code it seems more difficult but I'll have a look. Although it would be good to determine whether it could be left out before I dig too deep. So maybe I'll look into the dashboard's AMQP handling first.

nicksinger (approved):

PR #178, opened 2024-04-29
Don't crash in case of unexpected build date format

michaelgrifalconi:
try to parse the date, if unsuccessful, give up instead of crashing entirely

This is to fix cases like https://gitlab.suse.de/qa-maintenance/bot-ng/-/jobs/2463822 
where a manually cloned job with unexpected build strings confuse the bot logic.

### DIFF
--- a/openqabot/approver.py
+++ b/openqabot/approver.py
@@ -168,7 +168,7 @@ class Approver:
         try:
             current_build_date = datetime.strptime(current_build, "%Y%m%d")
         except (ValueError, TypeError):
-            log.info("Could not parse build date %s", current_build)
+            log.info("Could not parse build date %s. Won't try to look at older jobs for approval.", current_build)
             return False
 
         # Use at most X days old build. Don't go back in time too much to reduce risk of using invalid tests
@@ -184,7 +184,7 @@ class Approver:
             try:
                 job_build_date = datetime.strptime(job_build, "%Y%m%d")
             except (ValueError, TypeError):
-                log.info("Could not parse build date %s", current_build)
+                log.info("Could not parse build date %s. Won't consider this job as alternative for approval.", current_build)
                 continue
 
             # Check the job is not too old
--- a/openqabot/approver.py
+++ b/openqabot/approver.py
@@ -165,7 +165,11 @@ class Approver:
             return False
 
         current_build = older_jobs["data"][0]["build"][:-2]
-        current_build_date = datetime.strptime(current_build, "%Y%m%d")
+        try:
+            current_build_date = datetime.strptime(current_build, "%Y%m%d")
+        except (ValueError, TypeError):
+            log.info("Could not parse build date %s", current_build)
+            return False
 
         # Use at most X days old build. Don't go back in time too much to reduce risk of using invalid tests
         oldest_build_usable = current_build_date - timedelta(
@@ -177,7 +181,11 @@ class Approver:
         for i in range(1, len(older_jobs["data"])):
             job = older_jobs["data"][i]
             job_build = job["build"][:-2]
-            job_build_date = datetime.strptime(job_build, "%Y%m%d")
+            try:
+                job_build_date = datetime.strptime(job_build, "%Y%m%d")
+            except (ValueError, TypeError):
+                log.info("Could not parse build date %s", current_build)
+                continue
 
             # Check the job is not too old
             if job_build_date < oldest_build_usable:


Martchus (approved):

openqabot/approver.py
```diff
@@ -165,7 +165,11 @@ def was_ok_before(self, failed_job_id: int, inc: int) -> bool:
             return False
 
         current_build = older_jobs["data"][0]["build"][:-2]
-        current_build_date = datetime.strptime(current_build, "%Y%m%d")
+        try:
+            current_build_date = datetime.strptime(current_build, "%Y%m%d")
+        except (ValueError, TypeError):
+            log.info("Could not parse build date %s", current_build)
```
okurz, line 171: I suggest to make it clear what the consequence is. "Could not parse build date", so what? If it does not have any effect why parse the build date at all? If it does have an impact, what will it be? Will it fallback to a default value?

okurz (changes requested):

okurz:
And https://github.com/openSUSE/qem-bot/actions/runs/8875669610/job/24365587576?pr=178#step:6:19 says "openqabot/approver.py:160:4: R0911: Too many return statements (7/6) (too-many-return-statements)"

michaelgrifalconi:
> And https://github.com/openSUSE/qem-bot/actions/runs/8875669610/job/24365587576?pr=178#step:6:19 says "openqabot/approver.py:160:4: R0911: Too many return statements (7/6) (too-many-return-statements)"

About this, I would vote for disabling the check in this specific case. 
The example provided in pylint docs is for a very different situation that does not apply here: https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/too-many-return-statements.html

The reasons of this many returns are to "fail early" instead of going to the end of the function for no reason. But we can argue to still do that.

So the options I see are:
- move the for loop into a different function: I don't like this since would be a 'function' only called once from one single place, making it more difficult to understand the flow of the code due to jumping around in the file.

- change `return` to result=False/True and return only at the end, adding more line of code just to make the check happy

I can do whatever is requested, but I would like to find something that still makes sense to me, what is your opinion here?

openqabot/approver.py
```diff
@@ -165,7 +165,11 @@ def was_ok_before(self, failed_job_id: int, inc: int) -> bool:
...
+            return False
```
okurz, line 172: Moving your discussion into a thread here for easier handling:
>>  And https://github.com/openSUSE/qem-bot/actions/runs/8875669610/job/24365587576?pr=178#step:6:19 says "openqabot/approver.py:160:4: R0911: Too many return statements (7/6) (too-many-return-statements)"
>
> About this, I would vote for disabling the check in this specific case.
The example provided in pylint docs is for a very different situation that does not apply here: https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/too-many-return-statements.html

I don't think we are generally smarter than the people refining the proper python style over years so I don't think we should disable that check :)

> The reasons of this many returns are to "fail early" instead of going to the end of the function for no reason. But we can argue to still do that.
>
> So the options I see are:
> 1.  move the for loop into a different function: I don't like this since would be a 'function' only called once from one single place, making it more difficult to understand the flow of the code due to jumping around in the file.
> 2. change return to result=False/True and return only at the end, adding more line of code just to make the check happy
>
> I can do whatever is requested, but I would like to find something that still makes sense to me, what is your opinion here?

My opinion would actually depend on an answer in https://github.com/openSUSE/qem-bot/pull/178/files#r1582721997 so we should resolve that first.
In general I would be ok with both options. Additionally you can also consider using Exceptions instead of returns over multiple levels and reconsider how this function is called in the caller and also how the caller is called in _its_ caller

okurz (changes requested):

openqabot/approver.py
michaelgrifalconi: > I don't think we are generally smarter
Well they offer a way to ignore the check so their "smartness" considered the option to do so :)

I proposed a new log message for both cases (failed parsing on main job and on possible alternative job). Feel free to propose better wording, I hope what I try to explain is clear!

Then I wait for the proposal to handle the issue and keep the code easy to understand, avoiding more complicated workarounds just to hide a few simple returns because the tool say so.

okurz (changes requested):
The log messages are good. I would try the next days to extract the body of the for loop which would effectively reduce the return-statements. You could try the same or other means to address the reported style issue

openqabot/approver.py
okurz: Please see my suggestion #180 regarding how to handle the "too many return statements". Feel welcome to incorporate my changes into your PR or ask me to push the according changes with conflict resolution into your branch

michaelgrifalconi: Hello and thanks for the proposal! Wouldn't have made more sense if you submitted a pr for this branch, so your changes could be applied directly instead of having me (once again :P) resolve conflicts on an ongoing PR with the risk of messing something up?

okurz: sure, will do

okurz: … or rather: Can do but first I would like to resolve https://github.com/openSUSE/qem-bot/pull/180#pullrequestreview-2045737349 and any other potential subsequent comments

okurz:
Included with https://github.com/openSUSE/qem-bot/pull/181

PR #180, opened 2024-05-08
approver: Extract function 'was_older_job_ok'

okurz:


openqabot/approver.py
```diff
@@ -156,15 +156,61 @@ def validate_job_qam(self, job: int) -> bool:
...
+        # Check the job is not too old
+        if job_build_date < oldest_build_usable:
+            log.info(
+                "Cannot ignore aggregate failure %s for update %s because: Older jobs are too old to be considered"
+                % (failed_job_id, inc)
+            )
+            return False
+
+        if job["result"] != "passed" and job["result"] != "softfailed":
+            return False
```
Martchus, line 179:
```suggestion
            return None
```

openqabot/approver.py
```diff
@@ -173,46 +219,11 @@ def was_ok_before(self, failed_job_id: int, inc: int) -> bool:
...
-                "Ignoring failed aggregate %s and using instead %s for update %s"
-                % (failed_job_id, job["id"], inc)
-            )
-            return True
-
+        for job in older_jobs:
+            if self._was_older_job_ok(
+                failed_job_id, inc, job, oldest_build_usable, regex
+            ):
+                return True
```
Martchus, line 226:
```suggestion
            was_ok = self._was_older_job_ok(
                failed_job_id, inc, job, oldest_build_usable, regex
            )
            if was_ok is not None:
                return was_ok
```

Martchus (changes requested):
Doesn't this break the cases where the loop previously logged an error message (like `"Cannot ignore aggregate failure %s for update %s because: Older passing jobs do not have update under test"`) and then returned `False`? It would now continue iterating and log `"Cannot ignore aggregate failure %s for update %s because: Older usable jobs did not succeed. Run out of jobs to evaluate."` in addition to the actually relevant log message. I guess my two inline-comment suggestions would make it behave correctly (or at last as before).

openqabot/approver.py
Martchus: Well, I don't really care much about what this code does. I'm just pointing out that it would behave differently with your change. In case it is acceptable the commit message should be changed to reflect that.

However, I tend to keep the behavior the same because introducing a change nobody has asked for seems very needless. I also don't think the code readability is notably worse with three return values. (There's still the same number of branches.)

okurz: yes, you are right. That shouldn't be hidden in a "refactoring" commit. I will use your proposed approach and we can keep the true-boolean approach in mind as a potential explicit later approach

PR #181, opened 2024-05-08
date parsing rebased on extracted method - combining #180+#178

okurz:
This is
* #178
rebased onto
* #180

### DIFF
--- a/openqabot/approver.py
+++ b/openqabot/approver.py
@@ -165,7 +165,11 @@ class Approver:
         regex: Pattern[str],
     ) -> Optional[bool]:
         job_build = job["build"][:-2]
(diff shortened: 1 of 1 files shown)


okurz:
@michaelgrifalconi ok like this?

michaelgrifalconi:
I still feel it got more complicated just to please the formatting rules, but looks overall fine. 
Nice thing the `return None`, was not aware of that.

perlpunk (approved):

okurz:
> I still feel it got more complicated just to please the formatting rules, but looks overall fine.

It's important to just agree on styling rules, enforce them with automation and stick to it. If you feel in general a rule should be changed I suggest you propose that in a separate pull request but chances are slim we will agree on diluting the upstream python community rules

PR #183, opened 2024-10-15
Allow for more tolerant comment parsing

r-richardson:
Make comment parser more tolerant by allowing non-printable characters such as e.g. carriage returns, as well as preceding and trailing comment text

Related issue: https://progress.opensuse.org/issues/167836


openqabot/approver.py
```diff
@@ -58,6 +59,14 @@ def _handle_http_error(e: HTTPError, inc: IncReq) -> bool:
     return False
 
 
+def sanitize_comment_text(
+    text,
+):
+    text = "".join(filter(lambda x: x in string.printable, text))
```
okurz, line 65:
```suggestion
    text = "".join(x for x in text if x in string.printable)
```

okurz (approved):

Martchus (commented):
I guess this change will do it. It looks like you'll have to fix (or explicitly silence) the pylint errors.

r-richardson:
> I guess this change will do it. It looks like you'll have to fix (or explicitly silence) the pylint errors.

yes i'm looking into it though i am a little confused as these pylint errors do not seem to be related to my changes.

baierjan:
> > I guess this change will do it. It looks like you'll have to fix (or explicitly silence) the pylint errors.
> 
> yes i'm looking into it though i am a little confused as these pylint errors do not seem to be related to my changes.

It could just be that a newer pylint version is not happy with some old code.

Martchus (approved):

PR #184, opened 2024-10-30
Replace download.suse.de with mirror.nue2.suse.org

dzedro:
I tested allpatterns test with mirror.nue2.suse.org

https://progress.opensuse.org/issues/168097
VR: https://dzedro.suse.cz/tests/1674#step/patch_and_reboot/66

### DIFF
--- a/openqabot/__init__.py
+++ b/openqabot/__init__.py
@@ -7,7 +7,7 @@ OBS_URL = "https://api.suse.de"
 OBS_MAINT_PRJ = "SUSE:Maintenance"
 OBS_GROUP = "qam-openqa"
 DEVELOPMENT_PARENT_GROUP_ID = 9
-DOWNLOAD_BASE = "http://download.suse.de/ibs/SUSE:/Maintenance:/"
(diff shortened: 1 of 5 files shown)


Martchus:
Shouldn't this be configurable? I suppose we need to be able to deal with tests using embargoed data at some point. So we cannot completely avoid using `download.suse.de`.

okurz (changes requested):
We must not just use a different place. This would mean to rely on a mirror running not in prg2 which is both more remote and less reliable. This would deny access to embargoed updates which still need to be tested. Also this would mean that we trigger tests before the required assets are even synced.

dzedro:
Git it, it will be more complicated and done within. https://progress.opensuse.org/issues/169129

okurz:
No, this won't be done within https://progress.opensuse.org/issues/169129 which is only about time synchronisation to prevent starting a test before assets are available on a mirror. What should be done to use a local mirror explicitly is to look into the worker location and then select an according mirror. Alternatively we can set a mirror variable in each worker and read that with fallback to the default

PR #185, opened 2024-11-11
Reference internal gitlab for deployment

foursixnine:
Since deployment will style will change, reference documentation at time of writing

See also: https://gitlab.suse.de/qa-maintenance/bot-ng/-/merge_requests/72

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -55,6 +55,10 @@ tool for schedule maintenance jobs + sync SMELT/OpenQA to QEM-Dashboard
 * If there is a non-zero amount of related openQA jobs *and* none of them
   failed then qem-bot approves in IBS (`inc-approve`)
 
+### Deployment
+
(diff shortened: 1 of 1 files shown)


Martchus (approved):

foursixnine:
> I would refrain from stating any specific deployments based on SUSE internal resources. We can reference public projects in internal resources but shouldn't do the other way around.

I leave it up to qe-tools how you would like to document the change on the public repo, for me a reference is sufficient enough, as this project is a SUSE only thing, that depends on SUSE applications that aren't open sourced (nor should be).

Martchus:
I think we should be pragmatic here and simply add that link.

okurz (approved):
Fine. Some people complained in the past about "leaking information" but maybe that URL is innocent enough :)

PR #196, opened 2025-05-19
Add container definition and quadlets

baierjan:
As a proof of concept for running qem-bot in container and to support testing features from poo#180809, the gitea-sync action service is added first. This works best together with the qem-dashboard feature from https://github.com/openSUSE/qem-dashboard/pull/2472 as both services can be running inside one pod.

Although this is independent on #195, the service itself calls the feature which will be implemented in said PR. Additionally, the deployment requires dashboard URL to be configurable (and set to localhost) which is not yet the case outside of that PR.


Containerfile
```diff
@@ -0,0 +1,28 @@
...
+    python3-pika \
+    python3-requests \
+    python3-ruamel.yaml \
+    python3-jsonschema
+
+# Install application
+WORKDIR /app
+COPY . .
+# Alternatively, use sources from git
+# RUN git clone https://github.com/openSUSE/qem-bot .
```
Martchus, line 17: I suppose this `COPY` command is also where node modules are "installed" which is therefore an undocumented prerequisite. So just uncommenting the alternative to clone directly via Git is probably not going to work.

Martchus (commented):
I don't know about quadlet but I guess the change looks good. I only have one remark as inline comment.

By the way, I'll probably make the dashboard URL configurable in my PR today.

Containerfile
```diff
@@ -0,0 +1,28 @@
...
+
+# Install internal CA (use internal rpm as a quick hack, could be also done via adding a repo)
+RUN zypper in -y http://dist.nue.suse.com/ibs/SUSE:/CA/openSUSE_Tumbleweed/noarch/ca-certificates-suse-1.0-8.12.noarch.rpm
```
okurz, line 20: Please use the according file from https://download.opensuse.org/

okurz (changes requested):

okurz:
> As a proof of concept for running qem-bot in container […]

But we always ran qem-bot in container to be fair :)

Martchus:
With my latest changes you can use something like `export QEM_DASHBOARD_URL=http://127.0.0.1:3000/`.

Containerfile
Martchus: Oh, right.

Martchus (approved):

Containerfile
```diff
@@ -0,0 +1,30 @@
...
+
+# Install dependencies
+RUN zypper in -y -C \
+    git \
+    python3-osc \
+    python3-openqa_client \
+    python3-pika \
+    python3-requests \
+    python3-ruamel.yaml \
+    python3-jsonschema
```
okurz, line 11:
```suggestion
    python3-jsonschema && \
    zypper clean
```

okurz: Actually, better do that after the next zypper call

baierjan: True. And I guess I can probably merge the two installation in one zypper install call as this is not intended to be build by OBS anyway.

Martchus (approved):

okurz (approved):

PR #197, opened 2025-06-03
Consider concrete reviews on Gitea via our bot account

Martchus:
So far the code only checked reviews requested for the team the bot is part
of. It should also consider concrete reviews it has already made and thus
we also need to check `user.login`.

This PR also contains a few changes making development and demoing
easier. Those changes won't affect production.

Related ticket: https://progress.opensuse.org/issues/180812

### DIFF
--- a/openqabot/loader/config.py
+++ b/openqabot/loader/config.py
@@ -22,6 +22,12 @@ def load_metadata(
 
     loader = YAML(typ="safe")
 
+    log.debug(
+        "Loading meta-data from %s (with incidents: %r, with aggregates: %r)",
+        path,
+        not incidents,
+        not aggregate,
+    )
     for p in get_yml_list(path):
         try:
             data = loader.load(p)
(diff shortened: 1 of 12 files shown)


okurz (approved):

openqabot/loader/config.py
```diff
@@ -32,7 +38,7 @@ def load_metadata(
         try:
             settings = data.get("settings")
         except AttributeError:
-            # not valid yaml for bot settings
+            log.error("The YAML file '%s' contains no valid data for bot settings.", p)
```
b10n1k, line 41: suggestion: `"The YAML file '%s' contains invalid settings."

baierjan: That's not correct, it is possible the yaml is missing some data but it should not contain invalid one as the pipeline for metadata should check for that. And there is a distinction between invalid settings and no settings.

PR #208, opened 2025-07-02
Avoid `inconsistent scminfo …` log messages

Martchus:
It is expected that the build info of the different projects on OBS have different scminfo values.

This changes tracks these scminfo values separately. The scminfo of the codestream project is assigned as `scminfo` and the scminfo of sub projects assigned as e.g. `scminfo_SLES`. That latter is of course not supported by the dashboard and thus so far not actually synced there. Not sure whether this is wanted.

If we actually want to consider the scminfo of e.g. the `SLES` sub project as main scminfo we can easily flip the logic and adapt the test change. I'm just going here for *something* not really knowing what the requirements are.


openqabot/loader/gitea.py
```diff
@@ -155,22 +155,29 @@ def add_build_result(
...
-        return
+    project = res.get("project")
     for scminfo_element in res.findall("scminfo"):
         found_scminfo = scminfo_element.text
-        existing_scminfo = incident.get("scminfo", None)
+        project_match = re.search(".*:PullRequest:\\d+:(.*)", project)
+        if project_match:
+            scm_info_key = "scminfo_" + project_match.group(1)
+        else:
+            scm_info_key = "scminfo"
```
okurz, line 165:
```suggestion
        scm_info_key = "scminfo_" + project_match.group(1) if project_match else "scminfo"
```

this way it's clear that the scm_info_key is always set, just to different values

okurz (changes requested):

okurz:
mergify check now says "#commits-behind = 0 [📌 fast-forward merge requirement]" so that's good. @Martchus please rebase

PR #222, opened 2025-07-17
Add sub command for product increment approval

Martchus:
Related ticket: https://progress.opensuse.org/issues/184690

### DIFF
--- a/openqabot/args.py
+++ b/openqabot/args.py
@@ -78,6 +78,13 @@ def do_sync_aggregate_results(args):
     return syncer()
 
 
+def do_increment_approve(args):
+    from .incrementapprover import IncrementApprover
+
+    approve = IncrementApprover(args)
+    return approve()
+
+
 def do_amqp(args):
     from .amqp import AMQP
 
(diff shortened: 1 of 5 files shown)


Martchus:
I added the "not-ready" label because this needs https://github.com/os-autoinst/openQA/pull/6592.

okurz (changes requested):

openqabot/args.py
```diff
@@ -233,6 +240,51 @@ def get_parser():
...
 
+    cmdincrementapprove = commands.add_parser(
+        "increment-approve",
+        help="Approve the specified product increment if tests passed",
+    )
+    cmdincrementapprove.add_argument(
+        "--obs-project",
+        required=False,
+        type=str,
+        default="SUSE:SLFO:Products:SLES:15.99:TEST",
```
b10n1k, line 251: I found it a bit weird that the parameter is implies the build.opesuse.org but the default project cant be found there. I guess just `--project` would do better fit. wdyt?

tests/test_incrementapprover.py
```diff
@@ -0,0 +1,143 @@
...
+        "Skipping approval, some jobs on openQA are in pending states (running, scheduled)"
+        in messages
+    )
+
+
+@responses.activate
+def test_not_ok_openqa_jobs(caplog, fake_not_ok_jobs, monkeypatch):
+    run_approver(caplog, monkeypatch)
+    last_message = [x[-1] for x in caplog.record_tuples][-1]
+    assert "The following openQA jobs ened up with result 'failed'" in last_message
```
b10n1k, line 134: typo: s/ened/end/

openqabot/args.py
Martchus: But the OBS prefix makes it clear that this relates to the configured OBS instance (which can of course be our internal one) and not to some of the other services (like SMELT, the dashboard and openQA).

Note that OBS is the name of the software, not the name/domain of a specific instance of it (like build.opensuse.org).

openqabot/incrementapprover.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+class IncrementApprover:
+    def __init__(self, args: Namespace) -> None:
+        self.args = args
+        self.token = {"Authorization": "Token {}".format(args.token)}
+        self.client = openQAInterface(args)
+        osc.conf.get_config(override_apiurl=OBS_URL)
+
+    def __call__(self) -> int:
```
okurz, line 24: this method is very long. Can you extract methods or helper functions?. For example whenever you used a comment like below "find the latest request on OBS", why not make that a function?

okurz (approved):

b10n1k:
> @b10n1k I'm not sure what you're getting at. These tests can be considered unit tests as they mainly test `IncrementApprover` itself. This should be clear by the name `test_incrementapprover.py` of the file containing these tests. The function names the test is comprised of should be seen in this context as well. It makes no sense to rename e.g. `test_not_ok_openqa_jobs` to `test_increment_approver_with_not_ok_openqa_jobs`. It would just be wordy and doesn't add anything that isn't already clear from the context/file this function is defined in.

`test_pending_openqa_jobs` could be for instance `test_skip_approval_on_pending_openqa_jobs` 
So the test results tells you which functionality was tested.
You can say that the test is very small and if I read the function it wii be clear, but I dont think you should have to go to dig in the code to find out what is broken when the test fails/passes.
Maybe I cant convince you, so a simple search gave me this answer on stackoverflow. I think it describes what I am looking for https://stackoverflow.com/a/1594049

Martchus:
> `test_pending_openqa_jobs` could be for instance `test_skip_approval_on_pending_openqa_jobs` So the test results tells you which functionality was tested.

Ok, that makes actually sense.

b10n1k (approved):
+1 but there is a `not_ready` flag yet!

kalikiana (approved):

PR #225, opened 2025-08-12
Skip Gitea PRs if no packages have been build/published

Martchus:
So far the check for skipping PRs only skipped PRs if there were were failing/unpublished packages. Based on the feedback from comment https://progress.opensuse.org/issues/180812#note-64 we also need to check whether there is at least one successfully built and published package as IBS can apparently be in a state where no packages are present at all (or only excluded packages which we ignore).

### DIFF
--- a/openqabot/loader/gitea.py
+++ b/openqabot/loader/gitea.py
@@ -370,6 +370,16 @@ def add_packages_from_files(
             add_packages_from_patchinfo(incident, token, raw_url, dry)
 
 
+def is_build_acceptable_and_log_if_not(incident: Dict[str, Any], number: int) -> bool:
+    if len(incident["failed_or_unpublished_packages"]) > 0:
+        log.info("Skipping PR %s, not all packages succeeded and published", number)
+        return False
+    if len(incident["successful_packages"]) < 1:
+        log.info("Skipping PR %s, no packages have been built/published", number)
+        return False
+    return True
+
+
 def make_incident_from_pr(
     pr: Dict[str, Any],
     token: Dict[str, str],
@@ -418,11 +428,9 @@ def make_incident_from_pr(
         if len(incident["channels"]) == 0:
             log.info("Skipping PR %s, no channels found/considered", number)
             return None
-        if (
-            only_successful_builds
-            and len(incident["failed_or_unpublished_packages"]) > 0
+        if only_successful_builds and not is_build_acceptable_and_log_if_not(
+            incident, number
         ):
-            log.info("Skipping PR %s, not all packages succeeded and published", number)
             return None
         add_packages_from_files(incident, token, files, dry)
         if len(incident["packages"]) == 0:
--- a/responses/empty-build-results.xml
+++ b/responses/empty-build-results.xml
@@ -1,0 +1,5 @@
+<resultlist>
+  <result project="SUSE:SLFO:1.1.99:PullRequest:124" repository="product" arch="aarch64" state="published">
+    <!-- no packages here -->
+  </result>
+</resultlist>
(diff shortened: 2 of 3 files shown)


openqabot/loader/gitea.py
```diff
@@ -418,12 +418,17 @@ def make_incident_from_pr(
...
-            only_successful_builds
-            and len(incident["failed_or_unpublished_packages"]) > 0
-        ):
-            log.info("Skipping PR %s, not all packages succeeded and published", number)
-            return None
+        if only_successful_builds:
+            if len(incident["failed_or_unpublished_packages"]) > 0:
+                log.info(
+                    "Skipping PR %s, not all packages succeeded and published", number
+                )
```
Wabri, line 425: is this the tidied version of it?

Martchus: Me neither. Changing the formatting rules is a bit out of scope for this PR, though.

Wabri: Yes, I just wanted to discuss and understand. I didnt request changes!

setup.cfg
```diff
@@ -1,3 +1,3 @@
 [flake8]
 # E501 line too long
-extend-ignore = E501,D100,D101,D102,D103,D104,D105,D107,Q000
+extend-ignore = E501,D100,D101,D102,D103,D104,D105,D107,Q000,PLR911
```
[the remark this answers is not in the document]
Martchus: Here it is not a simple if-elseif structure that can be replaced by a map lookup. (Otherwise I probably would have written it like this in the first place.) Splitting the function would also make it less readable. Using more intermediate variables and a loop with break would make things even worse. I would not overthink this.

> We should generally trust the rules …

Sounds a bit boring :-)

openqabot/loader/gitea.py
```diff
@@ -418,12 +418,17 @@ def make_incident_from_pr(
...
+                return None
+            if len(incident["successful_packages"]) < 1:
+                log.info(
+                    "Skipping PR %s, no packages have been built/published", number
+                )
+                return None
```
okurz, line 431: I think this can be reworked by extracting a function and fulfilling the style request about not too many return statements

```suggestion
def packages_ok(incident):
    if len(incident["failed_or_unpublished_packages"]) > 0:
        log.info("Skipping PR %s, not all packages succeeded and published", number)
        return False
    if len(incident["successful_packages"]) < 1:
        log.info("Skipping PR %s, no packages have been built/published", number)
        return False
    return True

…
        if only_successful_builds and not packages_ok:
            return None
```

okurz (changes requested):

setup.cfg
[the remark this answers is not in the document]
okurz: please see https://github.com/openSUSE/qem-bot/pull/225#discussion_r2274695647

openqabot/loader/gitea.py
Martchus: I find that splitting this function makes it less readable but I can do that if you prefer this.

okurz: Well, I prefer to not disable style rules. Same as the discussion about line length both you and me prefer to stay consistent. I consider this approach the same :)

okurz: Well, commonly "too long functions" are seen as complicated and so are functions with "too many return statements". I think your change is good. Of course overall this adds about 2-3 lines but they are mostly blank lines for separation and the new function's declaration+call. Still I think an approach like this can help readability. Admittedly often a completely different approach can help much more not even getting into the challenge about "many return statements" but as neither you or me had a good idea right now how to do it I guess the current style is good enough

okurz (approved):

kalikiana (approved):

PR #226, opened 2025-09-02
Update logic to determine relevant Gitea PRs for new bot workflow

Martchus:
* Keep the old approach is still working in case it will be needed again after all
* See https://src.suse.de/products/SLFO/pulls/304 for an example PR
* See https://jira.suse.com/browse/MSQA-1009 for the related feature request


Martchus:
That's when you know that there are too many bots.

mimi1vx (approved):

PR #228, opened 2025-09-05
Allow specifying product repo in meta-data

Martchus:
This allows specifying e.g. `product_repo: SLES-SAP` in the meta-data. Then this specific product repo will be used to compute the repo hash and to populate the `INCIDENT_REPO` variable (instead of a default derived from the PR which is usually `SLES`).

Related ticket: https://progress.opensuse.org/issues/188328


kalikiana:
@mpagot FYI

PR #242, opened 2025-10-20
t: Add tests for AggregateResultsSync

okurz:


tests/test_aggrsync.py
```diff
@@ -0,0 +1,60 @@
...
+        token="null",
+        instance="null",
+        openqa_instance=urlparse("http://localhost"),
+    )
+    sync = AggregateResultsSync(args)
+    sync.client = MagicMock()
+    sync.client.get_jobs.return_value = [{"id": 1}]
+    assert sync() == 0
+    mock_normalize.assert_not_called()
+    mock_post.assert_not_called()
```
Martchus: This test should check what jobs have been posted. Otherwise the added coverage is misleading us into thinking this part of the code has tests that would fail if we break something.

okurz: Could you provide more specific suggestions because I don't understand the code enough yet to be able to solve that easily

Martchus: You can check the tests for Gitea. There I didn't completely mock away the client but used responses and its API to check what calls were made. (But maybe you can also change `@patch("openqabot.aggrsync.SyncRes.post_result")` so that invocations are recorded somehow. I didn't use `@patch` so far so I don't know what's possible with that.)

tests/test_aggrsync.py
```diff
@@ -0,0 +1,58 @@
...
+    mock_syncres.assert_called_once_with(args)
+
+
+@patch("openqabot.aggrsync.read_products", return_value=["product"])
+@patch("openqabot.aggrsync.get_aggregate_settings_data", return_value=["some_data"])
+@patch("openqabot.aggrsync.SyncRes.filter_jobs", return_value=False)
+@patch("openqabot.aggrsync.SyncRes.normalize_data")
+@patch("openqabot.aggrsync.SyncRes.post_result")
+def test_call_filter_jobs(mock_post, mock_normalize, mock_filter, mock_settings, mock_products):
```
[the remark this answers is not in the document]
okurz: You mean the patch call? With just the function name as first argument you can read like "patch them out for testing", like "mock as no-op"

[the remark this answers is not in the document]
Martchus: I meant the function parameters, e.g. `mock_post` (and not the decorators). However, when I understand now correctly, then these decorators are supplying the function parameters.

PR #246, opened 2025-10-21
Use more sensible linelength of 120 characters

okurz:
Same as for example in openQA the restriction of by default 88
characters causes ugly wrapping and reduces readability. This commit
changes the line-length that black formatting applies to 120 characters
and updates all code accordingly.


Martchus (approved):
Looks good but according to the failing CI check you probably missed some changed.

d3flex (approved):

PR #247, opened 2025-10-21
Use more efficient and simple ruff replacing flake8+pylint+black

okurz:
* Enable additional style checks for T100, T2, TD
* Replace flake8+black+pylint checks with ruff completely
* Fix style issues reported by extended ruff checks

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,12 @@ line-length = 120
 # D: pydocstyle (for D100, D101, etc.)
 # Q: flake8-quotes (for Q000)
 # PL: Pylint
-select = ["E", "F", "D", "Q", "PL"]
+extend-select =[
+    "E", "F", "D", "Q", "PL",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -0,0 +1,50 @@
...
+# PL: Pylint
+extend-select =[
+    "E", "F", "D", "Q", "PL",
+    "T100", # https://docs.astral.sh/ruff/rules/#flake8-debugger-t10
+    "T2", # https://docs.astral.sh/ruff/rules/#flake8-print-t20
+    "TD", # https://docs.astral.sh/ruff/rules/#flake8-todos-td
+]
+
+ignore = [
+    "E501",   # line-too-long
```
d3flex: Will the check complain if the line is > 120 with this as ignore?

d3flex: closing with a note. ruff will not complain. the style will need to get formatted by tidy.

PR #249, opened 2025-10-22
Enable more style checks

okurz:
* Fix and enable ruff 'ERA' style checks
* Fix and enable ruff 'N' style checks
* Fix and enable ruff 'I' style checks
* Fix and enable ruff 'W' style checks


perlpunk (approved):

d3flex (commented):
Do you know why is some files are still using the `List` type and some others are modified to the built-in `list`?

PR #250, opened 2025-10-22
More recommended style checks

okurz:
* Fix and enable ruff 'SIM' style checks
* Fix and enable ruff 'B' style checks
* CI: Remove redundant checkstyle call

### DIFF
--- a/openqabot/amqp.py
+++ b/openqabot/amqp.py
@@ -1,5 +1,6 @@
 # Copyright SUSE LLC
 # SPDX-License-Identifier: MIT
+import contextlib
 import json
 import re
 from argparse import Namespace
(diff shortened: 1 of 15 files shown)


perlpunk (commented):
> Fix and enable ruff 'B' style checks
> As recommend on https://docs.astral.sh/ruff/rules/printf-string-formatting/

I don't see the connection between that link and the changes

edit: also not for the 3d commit

okurz:
> > Fix and enable ruff 'B' style checks
> > As recommend on https://docs.astral.sh/ruff/rules/printf-string-formatting/
> 
> I don't see the connection between that link and the changes
> 
> edit: also not for the 3d commit

True. Removed those wrong references in both commits.

d3flex (approved):

PR #252, opened 2025-10-22
PART 2: Enable all ruff checks that potentially make sense - After #303

okurz:
* Fix and enable ruff 'FA' style checks
* Fix and enable ruff 'UP' style checks

After:
* #250
* #255
* #258
* #259
* #260
* #281
* #286
* #283
* #303

Before:
* #298

### DIFF
--- a/openqabot/loader/config.py
+++ b/openqabot/loader/config.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 from logging import getLogger
 from pathlib import Path
-from typing import List, Set, Union
 
 from ruamel.yaml import YAML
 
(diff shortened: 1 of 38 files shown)


openqabot/loader/config.py
```diff
@@ -1,24 +1,32 @@
 # Copyright SUSE LLC
 # SPDX-License-Identifier: MIT
+from __future__ import annotations
```
d3flex: Note that 3.14 doesnt need this.the import will be redundant and should be removed. Maybe add a comment like for <3.14

openqabot/loader/gitea.py
```diff
@@ -16,78 +19,78 @@
...
 def changed_files_url(repo_name: str, number: int) -> str:
     # https://docs.gitea.com/api/1.20/#tag/repository/operation/repoGetPullRequestFiles
-    return "repos/%s/pulls/%s/files" % (repo_name, number)
+    return "repos/{}/pulls/{}/files".format(repo_name, number)
 
 
 def comments_url(repo_name: str, number: int) -> str:
     # https://docs.gitea.com/api/1.20/#tag/issue/operation/issueCreateComment
-    return "repos/%s/issues/%s/comments" % (repo_name, number)
+    return "repos/{}/issues/{}/comments".format(repo_name, number)
```
d3flex: nitpick: Can we have fstring for those changes? somewhere below you do use fstring. I would say to adapt one of the two. ofc I am in favor of the fstring.

d3flex (approved):

Martchus (approved):

PR #255, opened 2025-10-23
Enable more style checks from list of all

okurz:
* Fix and enable ruff 'FURB' style checks
* Fix and enable ruff 'SLF' style checks
* Fix and enable ruff 'RET' style checks
* Fix and enable ruff 'RSE' style checks
* Fix and enable ruff 'T20' style checks
* Fix and enable ruff 'PIE' style checks
* Fix and enable ruff 'INP' style checks
* Fix and enable ruff 'ICN' style checks
* Fix and enable ruff 'ISC' style checks
* Fix and enable ruff 'FIX' style checks
* Fix and enable ruff 'T10' style checks
* Fix and enable ruff 'C4' style checks
* Apply more style suggestion from ruff "preview" rules

After
* #250

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,6 +31,7 @@ extend-select =[
     "RSE",
     "RET",
     "SLF",
+    "FURB",
 ]
 
 ignore = [


openqabot/amqp.py
```diff
@@ -94,6 +98,3 @@ def handle_incident(self, inc_nr: int, message) -> None:
         # Try to approve incident
         approve = Approver(self.args, inc_nr)
         approve()
-
-    def handle_aggregate(self, unused_build: str, unused_message) -> None:
-        return
```
Martchus: Sometimes it can be useful to have a function within a `class` even if `self` isn't used, e.g. to be able to override it in a derived class or if `self` might be used later. And here most likely the latter is the case. So I would avoid this change. (And your version also prompts the question: Why is `handle_incident` part of the `class` and `handle_aggregate` not? Both functions will be similar once fully implemented. So this is not very consistent.)

pyproject.toml
```diff
@@ -24,6 +24,7 @@ extend-select =[
     "C4",
     "T10",
     "FIX",
+    "ISC",
```
Martchus: https://docs.astral.sh/ruff/rules/#flake8-implicit-str-concat-isc

I guess this is ok.

pyproject.toml
```diff
@@ -32,6 +32,7 @@ extend-select =[
     "RSE",
     "RET",
     "SLF",
+    "FURB",
```
Martchus: This is actually a long list of changes: https://docs.astral.sh/ruff/rules/#pyupgrade-up

I guess I'm good with it but I only skimmed though the list.

Martchus (changes requested):
I guess most of these changes make sense but there are a few exceptions.

I guess if almost all of these style nitpicks can be applied automatically then it won't become too annoying to work with them. (Otherwise I would consider most of these style nitpicks way too annoying.)

okurz:
> I guess most of these changes make sense but there are a few exceptions.
> 
> I guess if almost all of these style nitpicks can be applied automatically then it won't become too annoying to work with them. (Otherwise I would consider most of these style nitpicks way too annoying.)

I addressed all your points. Some already in #250 which is merged. Others I have applied here. Yes, most of the style changes enforced by the rules in this PR can actually be fixed automatically with `ruff check --fix` and I am sure that the rules here are generally acceptable and help us in development *and* review. Later rules, e.g. the ones in #252 might be more debatable though :)

Martchus (approved):

r-richardson:
LGTM

PR #258, opened 2025-10-24
PART 2: Enable more style checks from list of all - After #255

okurz:
* Fix and enable ruff 'PGH' style checks
* Fix and enable ruff 'PERF' style checks
* Fix and enable ruff 'C90' style checks
* Fix and enable ruff 'G' style checks
* Fix and enable ruff 'LOG' style checks
* Fix and enable ruff 'DTZ' style checks
* Fix and enable ruff 'A' style checks
* Fix and enable ruff 'BLE' style checks
* Fix and enable ruff 'EXE' style checks
* CI: Bump Python version setup to 3.11

After:
* #255

### DIFF
--- a/openqabot/approver.py
+++ b/openqabot/approver.py
@@ -28,12 +28,7 @@ from .loader.qem import (
     get_incidents_approver,
     get_single_incident,
 )
-
-try:
-    from datetime import UTC
-except ImportError:  # python <3.11 compatibility
-    from datetime.timezone import utc as UTC
-
+from .utc import UTC
 
 log = getLogger("bot.approver")
 
--- a/openqabot/osclib/comments.py
+++ b/openqabot/osclib/comments.py
@@ -6,10 +6,7 @@ from xml.etree import ElementTree as ET
 
 from osc.core import http_DELETE, http_GET, http_POST, makeurl
 
-try:
-    from datetime import UTC
-except ImportError:  # python <3.11 compatibility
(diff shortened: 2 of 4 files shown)


openqabot/aggrsync.py
```diff
@@ -22,11 +21,7 @@ def __init__(self, args: Namespace) -> None:
     def __call__(self) -> int:
         update_setting = []
         for product in self.product:
-            try:
-                update_setting += get_aggregate_settings_data(self.token, product)
-            except EmptySettings as e:
-                log.info(e)
-                continue
```
Martchus: Did you change the behavior of `get_aggregate_settings_data`? I didn't find that so I'm wondering whether the changed version is correct. (I wouldn't trust the fact the unit tests are still passing.)

openqabot/loader/gitea.py
```diff
@@ -326,7 +326,7 @@ def determine_relevant_archs_from_multibuild_info(obs_project: str, dry: bool) -
     else:
         try:
             multibuild_data = get_multibuild_data(obs_project)
-        except Exception as e:
+        except Exception as e:  # noqa: BLE001
```
okurz:
```suggestion
        except Exception as e:  # noqa: BLE001 true-positive: Consider to use fine-grained exceptions
```

okurz (commented):
see comment

Martchus (approved):

PR #260, opened 2025-10-27
PART 4: Enable more stable style checks + fix xfail - After #259

okurz:
After:
* #259

Plus:
* tests: Fix xfail in test_approve


okurz:
Peer-reviewed and approved.

PR #261, opened 2025-10-28
Fix very misleading logging in `increment-approve`

Martchus:
Use the actual value for variables like flavor that has been derived from the config in log message like "Skipping approval …" (and not the default value). Otherwise it is very misleading what was skipped, scheduled or approved.


okurz:
@Martchus can you rebase?

PR #267, opened 2025-10-30
Approve increments only if all configurations/flavors look good

Martchus:
Having multiple configurations was initially added to support multiple products (e.g. SLES and SL Micro which would always have distinct increment requests). It was later extended to support scheduling additional flavors of the same product. For this to be useful we actually need to check all scheduled products of all flavors first and only decide whether to approve the request in the end taking all scheduled products into account.

### DIFF
--- a/tests/test_incrementapprover.py
+++ b/tests/test_incrementapprover.py
@@ -156,8 +156,8 @@ def prepare_approver(
 
 
 def run_approver(
-    caplog,
-    monkeypatch,
+    caplog: LogCaptureFixture,
+    monkeypatch: MonkeyPatch,
     schedule: bool = False,
     reschedule: bool = False,
     diff_project_suffix: str = "none",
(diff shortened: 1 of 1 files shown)


openqabot/incrementapprover.py
```diff
@@ -125,41 +136,41 @@ def _check_openqa_jobs(
...
                 else:
                     not_ok_jobs[result].update(info["job_ids"])
-        return ok_jobs
 
     def _evaluate_list_of_openqa_job_results(
         self, list_of_results: List[Dict[str, Dict[str, Dict[str, Any]]]]
-    ) -> Tuple[int, List[str]]:
-        ok_jobs = 0  # count ok jobs
+    ) -> Tuple[Set[int], List[str]]:
+        ok_jobs = set()  # keep track of ok jobs
```
d3flex: This function seems can use the new `ApprovalStatus` class. Both `ok_jobs` and `reasons_to_disapprove` are members of that new class. If I look at the names of the objects I would also say that the `ApprovalStatus` should be a dependency in `IncrementApprover`. 

All this implies a bigger refactoring rather a single change.

Martchus: I don't really know what you're saying. What do you mean exactly when you say class `ApprovalStatus` should be a *dependency* in `IncrementApprover`?

PR #273, opened 2025-11-10
Prevent CI warning about ruff version

d3flex:
this is to prevent the following warning in CI

```Found ruff version in pyproject.toml: undefined
Warning: Could not parse version from
/home/runner/work/qem-bot/qem-bot/pyproject.toml. Using latest version.
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,8 @@ jobs:
     steps:
       - uses: actions/checkout@v5
       - uses: astral-sh/ruff-action@v3
+        with:
+          version: "latest"
       - name: Check format
         run: ruff format --check
       - uses: actions/setup-python@v6


perlpunk:
> this is to prevent the following warning in CI
> 
> ```
> Warning: Could not parse version from
> /home/runner/work/qem-bot/qem-bot/pyproject.toml. Using latest version.
> ```

And where does this warning actually appear?
There are several CI checks, and they have several steps. The warning could appear in any of them.
Have you thought about just linking to a previous CI job and make our lives a bit easier?

d3flex:
> If the warning is this one: https://github.com/openSUSE/qem-bot/actions/runs/19231251449/job/54970083681?pr=272#step:3:8 it actually comes from the astral-sh/ruff-action and looks like the action itself wants the version explicitly: https://github.com/astral-sh/ruff-action#install-specific-versions
> 
> ```
> - uses: astral-sh/ruff-action@v3
>   with:
>     version: "latest"
> ```
> 
> Then it should be fine with not specifying the version in the pyproject.toml.

works. thanks

perlpunk (approved):

PR #274, opened 2025-11-12
Postpone job_result when reschedule

mpagot:
Postpone the call to request_openqa_job_result. This allow to completly skip this openQA request when IncrementApprover is configured with reschedule argument.

### DIFF
--- a/tests/test_incrementapprover.py
+++ b/tests/test_incrementapprover.py
@@ -186,12 +186,27 @@ def test_approval_if_there_are_only_ok_openqa_jobs(caplog: LogCaptureFixture, mo
 
 
 @responses.activate
-@pytest.mark.usefixtures("fake_ok_jobs", "fake_product_repo")
 def test_skipping_if_rescheduling(caplog: LogCaptureFixture, monkeypatch: MonkeyPatch) -> None:
-    run_approver(caplog, monkeypatch, reschedule=True)
-    last_message = [x[-1] for x in caplog.record_tuples][-1]
-    assert "have passed" not in last_message
-    assert "Re-scheduling jobs for" in last_message
+    # Mock the openQA API to return existing failed jobs.
+    # These would normally block approval, but reschedule=True should ignore them.
+    resp = responses.add(
+        GET,
+        openqa_url,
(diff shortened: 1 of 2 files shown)


tests/test_incrementapprover.py
```diff
@@ -186,12 +186,36 @@ def test_approval_if_there_are_only_ok_openqa_jobs(caplog: LogCaptureFixture, mo
...
-    last_message = [x[-1] for x in caplog.record_tuples][-1]
-    assert "have passed" not in last_message
-    assert "Re-scheduling jobs for" in last_message
+    # Mock the openQA API to return existing failed jobs.
+    # These would normally block approval, but reschedule=True should ignore them.
+    resp = responses.add(
+        GET,
+        openqa_url,
+        json={"done": {"passed": {"job_ids": [20]}, "failed": {"job_ids": [21]}}},
+    )
```
Martchus: This should use `fake_not_ok_jobs` instead of repeating that code here.

okurz: We shouldn't accept this PR as long as it's introducing this duplication. How you solve it is up to your preferences

tests/test_incrementapprover.py
```diff
@@ -186,12 +186,28 @@ def test_approval_if_there_are_only_ok_openqa_jobs(caplog: LogCaptureFixture, mo
...
+    # Mock the OBS API to return a list of available updates.
+    test_product_repo = read_json("test-product-repo")
+    responses.add(
+        GET,
+        OBS_DOWNLOAD_URL + "/OBS:/PROJECT:/TEST/product/?jsontable=1",
+        json=test_product_repo,
+    )
+
+    (errors, jobs) = run_approver(caplog, monkeypatch, reschedule=True)
+
```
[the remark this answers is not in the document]
mpagot: I disagree: I like to have a visual separation between:

```
def test_something():
    <SETUP>

    <EXERCISE>

    <VERIFY>
```

okurz:
> Not interested in proceed with this PR. Thanks for all the suggestions.

That's unfortunate to hear. The original idea is still valid, isn't it?

PR #277, opened 2025-11-12
Improve/extend repo-diff computation

Martchus:
This is for https://progress.opensuse.org/issues/190542 as it turns out that zstd compression is required for the real SLES repo. This explains why kernel tests wrongly scheduled (see https://progress.opensuse.org/issues/190542#note-26) but there's most likely more to be fixed.

### DIFF
--- a/openqabot/repodiff.py
+++ b/openqabot/repodiff.py
@@ -8,7 +8,7 @@ from collections import defaultdict
 from logging import getLogger
 from typing import Any, DefaultDict, Dict, List, NamedTuple, Optional, Set, Tuple, Union
 
-import pyzstd
+import zstandard
 
 from . import OBS_DOWNLOAD_URL
 from .utils import retry10 as requests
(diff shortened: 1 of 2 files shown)


Martchus:
For now I first want to see what else I need to make this work at all with the production SLES repo. If I have everything together I can extend tests.

pyproject.toml
```diff
@@ -17,6 +17,7 @@ dependencies = [
     "osc",
     "openqa-client",
     "pika",
+    "pyzstd",
```
okurz: this package does not seem to be available on the current Leap 15.6 based system. Can we consider alternatives? Otherwise wait for https://progress.opensuse.org/issues/191617

openqabot/repodiff.py
```diff
@@ -116,5 +128,5 @@ def __call__(self) -> int:
             count,
             args.repo_a,
         )
-        log.info(diff)
+        print(json.dumps(diff, indent=4, default=list))  # noqa: T201
```
d3flex: why did you replace the log with print?

tests/test_repodiff.py
```diff
@@ -0,0 +1,23 @@
...
+from collections.abc import Generator
+
+from _pytest.capture import CaptureFixture
+
+from openqabot.repodiff import RepoDiff
+
+_namespace = namedtuple("Namespace", ("dry", "fake_data", "repo_a", "repo_b"))
+
+
+def test_repodiff(capsys: Generator[CaptureFixture[str]]) -> None:
```
d3flex: Also this is quite new to me. do you really need `Generator` here?
According to https://docs.pytest.org/en/7.1.x/reference/reference.html?highlight=capsys#pytest.capture.capsys the following would work as well, no?

```import pytest

def test_repodiff(capsys: pytest.CaptureFixture[str]) -> None:
```

okurz (approved):

PR #278, opened 2025-11-12
Ensure complete patch coverage as target

okurz:


perlpunk (approved):

okurz:
> please not yet this change. There are modules which likely we do not touch yet. And modules still not active as the amqp.py. We still experiment with it and it is not expected to cover everything. But i am not against the TDD or DDD to move on. So the core parts of the tool is in track

This is only about the patch coverage. If you need/want to implement something then add according failing tests, add implementation to fix tests, 100% patch coverage done :)

PR #279, opened 2025-11-13
Extend package filtering for creating additional builds for product increments

Martchus:
See particular commit messages and https://progress.opensuse.org/issues/190542#note-32


d3flex (approved):
I think this is a good example which a `generator` can be used. But I think you would like to close the ticket. If you dont mind i will approve and propose a change above it later.

PR #283, opened 2025-11-15
Enable all ruff checks up to 'TID'

okurz:
* Fix and enable ruff 'TID' style checks
* Fix and enable ruff 'SLOT' style checks
* Fix and enable ruff 'PT' style checks
* Fix and enable ruff 'PYI' style checks
* Fix and enable ruff 'INT' style checks
* Fix and enable ruff 'EM' style checks
* Fix and enable ruff 'CPY' style checks
* Fix and enable ruff 'COM' style checks
* Fix and enable ruff 'ASYNC' style checks
* Fix and enable ruff 'YTT' style checks
* Fix and enable ruff 'DOC' style checks
* Fix and enable ruff preview style checks

After:
* #281
* #286

Before:
* #252

### DIFF
--- a/openqabot/loader/config.py
+++ b/openqabot/loader/config.py
@@ -6,11 +6,11 @@ from typing import List, Set, Union
 
 from ruamel.yaml import YAML
 
-from ..errors import NoTestIssues
-from ..types import Data
-from ..types.aggregate import Aggregate
-from ..types.incidents import Incidents
-from ..utils import get_yml_list
+from openqabot.errors import NoTestIssues
+from openqabot.types import Data
+from openqabot.types.aggregate import Aggregate
+from openqabot.types.incidents import Incidents
+from openqabot.utils import get_yml_list
 
 log = getLogger("bot.loader.config")
 
--- a/openqabot/loader/gitea.py
+++ b/openqabot/loader/gitea.py
@@ -18,9 +18,9 @@ import urllib3.exceptions
 from lxml import etree
 from osc.core import MultibuildFlavorResolver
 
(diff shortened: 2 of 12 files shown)


tests/test_amqp.py
```diff
@@ -11,12 +11,30 @@
...
+    openqa_instance=urlparse("http://instance.qa"),
+    url=None,
+    gitea_token=None,
+)
 amqp = AMQP(args)
 
-fake_method = namedtuple("Method", ["routing_key"])
-fake_job_done = fake_method("suse.openqa.job.done")
+
+class Method(NamedTuple):
```
Martchus: Maybe the name `Method` is a bit generic.


```suggestion
class FakeMethod(NamedTuple):
```

tests/test_approve.py
```diff
@@ -427,22 +462,25 @@ def f_osc_core(*_args: Any, **_kwds: Any) -> NoReturn:
...
+def test_osc_unknown_exception(caplog: LogCaptureFixture, monkeypatch: pytest.MonkeyPatch) -> None:
     caplog.set_level(logging.DEBUG, logger="bot.approver")
 
     def f_osc_core(*_args: Any, **_kwds: Any) -> NoReturn:
-        raise Exception("Fake OBS exception")
+        class ArbitraryObsException(Exception):
+            def __init__(self) -> None:
+                super().__init__("Fake OBS exception")
+
+        raise ArbitraryObsException
```
Martchus: This goes against the intention of the style check. Now one can still not catch this exception without a generic exception handler as `ArbitraryObsException` is probably not accessible from outside this function. (Or is it accessible? Then this would be ok of course.)

tests/test_amqp.py
okurz: Done

perlpunk (approved):

PR #284, opened 2025-11-16
PART 4: Add coverage improvements + full AMQP coverage - After #283+#252+#298

okurz:
* Increase 'amqp' coverage for 'handle_incident_updates_dashboard'
* Makefile: Show missing coverage lines directly
* Add coverage for amqp 'stop' on no connection
* Increase 'amqp' coverage for 'incident_value_error'
* Increase 'amqp' coverage for 'bad_build'
* Increase 'amqp' coverage for 'no_build'
* Increase 'amqp' coverage for 'bad_routing_key'
* Increase 'amqp' coverage for 'call'
* Increase 'amqp' coverage for 'init_no_url'
* Ensure coverage target already on pytest call
* Collect branch coverage as well
* codecov: Rely on default values for precision
* Ensure complete patch coverage as target

After:
* #283
* #252 
* #298

Before:
* #292

### DIFF
--- a/openqabot/approver.py
+++ b/openqabot/approver.py
@@ -346,7 +346,7 @@ class Approver:
             )
         except HTTPError as e:
             return _handle_http_error(e, inc)
-        except Exception:  # pylint: disable=broad-except
+        except Exception:
             log.exception("")
             return False
 
@@ -362,7 +362,7 @@ class Approver:
                 msg,
                 inc.scm_info,
             )
-        except Exception:  # pylint: disable=broad-except
+        except Exception:
             log.exception("")
             return False
         return True
--- a/openqabot/loader/config.py
+++ b/openqabot/loader/config.py
@@ -36,7 +36,7 @@ def load_metadata(
     for p in get_yml_list(path):
         try:
             data = loader.load(p)
-        except Exception:  # pylint: disable=broad-except
+        except Exception:
             log.exception("")
             continue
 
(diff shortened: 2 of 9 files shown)


openqabot/amqp.py
```diff
@@ -79,6 +81,22 @@ def on_message(
...
+        try:
+            AMQP.operation = "incident"
+            r = self.normalize_data(inc, job)
+        except KeyError:
+            return
+        # Post update about matching openQA job into dashboard database
+        if r["job_id"] == message["id"]:
+            self.post_result(r)
+
+    def fetch_openqa_results(self, inc: Data, message: dict[str, Any]) -> None:
```
d3flex: s/fetch_openqa_results/_fetch_openqa_results/

d3flex (commented):
As we do not have clear standards, here some notes on test_amqp.py
- Too many patches which makes me think if this actually test anything
- Some of the patches can be extracted in a fixture to avoid duplicate them
- To make it clear and readable, it could be a good idea to use one parameterized function

okurz:
> As we do not have clear standards, here some notes on test_amqp.py
>     * Too many patches which makes me think if this actually test anything

It does as a proper unit test mocking external dependencies. I have implemented separate tests for the higher level methods and the newly introduced methods in the amqp module.

>     * Some of the patches can be extracted in a fixture to avoid duplicate them

No, they are called differently as functions are mocked with differing return values or side-effects.

>     * To make it clear and readable, it could be a good idea to use one parameterized function

I would not know for what. Also in tests commonly it is preferred to be explicit at the potential cost for a bit of redundancy unlike in the production code that is tested.

Martchus (approved):

PR #288, opened 2025-11-17
Allow computing package list for increments from source report

Martchus:
* Based on https://gitlab.suse.de/maintenance/tools/-/commit/2a12921dbd7d56a7e74cc382f89ab6709a2d237b
* See https://progress.opensuse.org/issues/190542#note-47

---

Still a draft as I need to write tests, improve the handling of the temp files and catch up with the prototype on https://gitlab.suse.de/maintenance/tools/-/blob/master/report-diff-for-product-increment.pl which Marcus is still updating.

### DIFF
--- a/tests/test_incrementapprover.py
+++ b/tests/test_incrementapprover.py
@@ -178,6 +178,7 @@ def prepare_approver(
     diff_project_suffix: str = "none",
     test_env_var: str = "",
     config: IncrementConfig | None = None,
+    request_id: int | None = None,
 ) -> IncrementApprover:
     os.environ["CI_JOB_URL"] = test_env_var
     caplog.set_level(logging.DEBUG, logger="bot.increment_approver")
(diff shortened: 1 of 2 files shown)


tests/test_incrementapprover.py
```diff
@@ -317,6 +373,29 @@ def test_scheduling_extra_livepatching_builds_with_no_openqa_jobs(
...
+    args = {
+        "caplog": caplog,
+        "monkeypatch": monkeypatch,
+        "schedule": True,
+        "diff_project_suffix": "source-report",
+        "config": next(IncrementConfig.from_config_file(path)),
+    }
+    (errors, jobs) = run_approver(**args)
+    messages = [x[-1] for x in caplog.record_tuples]
+    assert "Comuting source report diff for request 42" in messages
```
d3flex: s/Comuting/computing/? no sure what is it meant to be?

Martchus:
It would be nice to merge this to get out of the rebuild loop. I have also just created https://gitlab.suse.de/qa-maintenance/metadata/-/merge_requests/1609 so we can configure using this new approach.

okurz (approved):
The logic in incrementapprover becomes more and more complicated quickly and I am not convinced this is sustainable. Just based on your statement that you would prefer this to be merged quickly I am approving based an skimming over the algorithms.

PR #290, opened 2025-11-17
Use lxml consistently

okurz:
We have lxml available where we care. The change to introduce defused
introduced a problem for the type checks. The easiest and most
consistent approach is if we just rely on lxml everywhere.


okurz:
> Did you test the uncovered code in `comments.py` manually?

Yes. Also I encountered a problem "ValueError: Unicode strings with encoding declaration are not supported. Please use bytes input or XML fragments" in repohash.py . Will look into this.

PR #292, opened 2025-11-17
PART 1: Add more test coverage

okurz:
Mostly test coverage improvement and other minor behavior improvements

* See individual commits

After:
* #284

Before:
* #300

### DIFF
--- a/openqabot/loader/config.py
+++ b/openqabot/loader/config.py
@@ -49,35 +49,33 @@ def load_metadata(
         if "product" not in data:
             log.debug("Skipping invalid config %s", p)
             continue
-
-        if settings:
-            for key in data:
-                if key == "incidents" and not incidents:
+        for key in data:
+            if key == "incidents" and not incidents:
+                ret.append(
+                    Incidents(
+                        data["product"],
+                        data.get("product_repo"),
+                        data.get("product_version"),
+                        settings,
+                        data[key],
+                        extrasettings,
+                    ),
+                )
+            elif key == "aggregate" and not aggregate:
+                try:
                     ret.append(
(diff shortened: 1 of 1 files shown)


openqabot/pc_helper.py
```diff
@@ -130,6 +130,6 @@ def is_newer(date1: str, date2: str) -> bool:
         if (region is not None) and (region != image["region"]):
             continue
         # Get latest one based on 'publishedon'
-        if recentimage is None or is_newer(image["publishedon"], recentimage["publishedon"]):
+        if recentimage is None or is_newer(image["publishedon"], recentimage["publishedon"]):  # pragma: no branch
```
Martchus: To me this looks like a branch. (Actually, there are multiple branches due to the `or`.) So why `no branch`?

perlpunk (approved):

tests/test_giteasync.py
```diff
@@ -125,6 +125,7 @@ def run_gitea_sync(
     *,
     no_build_results: bool = False,
     allow_failures: bool = True,
+    dry: bool = False,
```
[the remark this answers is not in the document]
okurz: That does not make sense. In https://github.com/openSUSE/qem-bot/pull/292/files#diff-88f9b0feff1cde6d6b67e6074f76e481483177dd0a64a89a6a871700e2f4fb5dR156 I call `run_gitea_sync` with `dry=True` so we need the argument

openqabot/pc_helper.py
Martchus: And what makes this branch special then (so we don't need to track branch coverage even though we might do that generally)?

PR #299, opened 2025-11-18
Use interpreter from environment to better support virtual environments

nicksinger:
This requests `qem-bot.py` to use the python interpreter configured in the current users environment. It is mainly useful for systems where not all requirements can be installed system-wide (e.g. on my system, I have to install `osc` via pip into an virtual environment). Without this change:

```shell
nicksinger@workstation git/qem-bot ‹master› » pip freeze | grep osc  # system python does not have the osc module available
nicksinger@workstation git/qem-bot ‹master› » ./qem-bot.py --debug --configs metadata/qem-bot -s metadata/qem-bot/singlearch.yml -t "${token}" --dry increment-approve --increment-config metadata/qem-bot  # traceback expected since venv not activated, using system python
Traceback (most recent call last):
  File "/home/sqozz/git/qem-bot/./qem-bot.py", line 7, in <module>
    main()
    ~~~~^^
  File "/home/sqozz/git/qem-bot/openqabot/main.py", line 32, in main
    sys.exit(cfg.func(cfg))
             ~~~~~~~~^^^^^
  File "/home/sqozz/git/qem-bot/openqabot/args.py", line 83, in do_increment_approve
    from .incrementapprover import IncrementApprover
  File "/home/sqozz/git/qem-bot/openqabot/incrementapprover.py", line 11, in <module>
    import osc.conf
ModuleNotFoundError: No module named 'osc'
nicksinger@workstation git/qem-bot ‹master› » . ./venv/bin/activate
(venv) nicksinger@workstation git/qem-bot ‹master› » pip freeze | grep osc  # proof that osc is installed in the now active environment
osc==1.22.0
(venv) nicksinger@workstation git/qem-bot ‹master› » ./qem-bot.py --debug --configs metadata/qem-bot -s metadata/qem-bot/singlearch.yml -t "${token}" --dry increment-approve --increment-config metadata/qem-bot  # invocation still fails because wrong interpreter is used
Traceback (most recent call last):
  File "/home/sqozz/git/qem-bot/./qem-bot.py", line 7, in <module>
    main()
    ~~~~^^
  File "/home/sqozz/git/qem-bot/openqabot/main.py", line 32, in main
    sys.exit(cfg.func(cfg))
             ~~~~~~~~^^^^^
  File "/home/sqozz/git/qem-bot/openqabot/args.py", line 83, in do_increment_approve
    from .incrementapprover import IncrementApprover
  File "/home/sqozz/git/qem-bot/openqabot/incrementapprover.py", line 11, in <module>
[...]

PR #316, opened 2025-11-26
feat: Use consistent 'python3' calls to not rely on shebang

okurz:
This improves portability of the project, to know more see https://github.com/os-autoinst/openqa_review/pull/153#discussion_r658592487 and https://github.com/openSUSE/qem-bot/pull/299#event-21022285328

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,4 +29,4 @@ test: only-test checkstyle
 
 .PHONY: test-all-commands-unstable
 test-all-commands-unstable:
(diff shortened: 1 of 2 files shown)


Wabri:
question(blocking): yes, but why?

nicksinger (approved):
This improves portability of the project by a lot but breaks a single use-case which seems rather important, see https://github.com/os-autoinst/openqa_review/pull/153#discussion_r658592487 and https://github.com/openSUSE/qem-bot/pull/299#event-21022285328

Wabri (approved):

PR #317, opened 2025-11-26
Proposal: Add minimal development containerized environemnt to run tests

Wabri:


okurz (changes requested):
As there is already https://github.com/openSUSE/qem-bot/blob/master/Containerfile this might be a bit confusing for users finding both. Also Containerfile is using a different approach. Here you spawn a container to run pip within. Why not just a Python virtualenv then?

Podmanfile.dev
```diff
@@ -0,0 +1,11 @@
+FROM python:3.12-slim
```
baierjan: Is there a reason to not use Tumbleweed/Leap as a base?

Wabri:
Closing due stale and we kinda already have a minimal development containerized environemnt

PR #318, opened 2025-11-27
PART 11: More tests and minor improvements

okurz:
* feat(gitea): Add log_debug for start of remote requests
* test: Add more tests for smeltsync to improve coverage
* test: Add test for RetryError in repohash.py
* style: Fix vulture and ruff warnings
* test: Add test for invalid yaml in incrementconfig
* refactor(tests): Add tests for incrementapprover 'get_regex_match' and 'find_request_on_obs'

Before:
* #319

### DIFF
--- a/openqabot/loader/gitea.py
+++ b/openqabot/loader/gitea.py
@@ -124,6 +124,7 @@ def compute_repo_url_for_job_setting(
 
 
 def get_open_prs(token: dict[str, str], repo: str, *, dry: bool, number: int | None) -> list[Any]:
+    log.debug("Loading open PRs from '%s'%s", repo, ", dry-run" if dry else "")
     if dry:
         return read_json("pulls")
     open_prs = []


Readme.md
```diff
@@ -166,6 +166,13 @@ way to get them is via pip:
 
     pip install -e ".[dev]"
 
+For local development, you can use `uv` to create a virtual environment and
+install dependencies:
+
+    uv venv
+    source .venv/bin/activate
+    uv pip install -e ".[dev]"
+
```
Wabri: where does `uv` come from?

openqabot/amqp.py
```diff
@@ -60,9 +60,9 @@ def stop(self) -> None:
 
     def on_message(
         self,
-        unused_channel: pika.channel.Channel,  # noqa: ARG002 Unused method argument
+        _: pika.channel.Channel,
         method: pika.spec.Basic.Deliver,
-        unused_properties: pika.spec.BasicProperties,  # noqa: ARG002 Unused method argument
+        __: pika.spec.BasicProperties,
```
d3flex: seems like double underscore here. Can you take a look?

okurz: That's on purpose to mark the unused method arguments. I can't use `_` two times

tests/test_incrementapprover.py
```diff
@@ -519,3 +519,21 @@ def test_get_regex_match_invalid_pattern(caplog: pytest.LogCaptureFixture, monke
     approver = prepare_approver(caplog, monkeypatch)
     approver._get_regex_match("[", "some string")  # noqa: SLF001
     assert "Pattern `[` did not compile successfully" in caplog.text
+
+
+def test_find_request_on_obs_with_request_id(caplog: pytest.LogCaptureFixture, monkeypatch: pytest.MonkeyPatch) -> None:
+    def fake_request_from_api(apiurl: str, reqid: str) -> osc.core.Request:
+        assert apiurl == OBS_URL
+        assert reqid == "43"
```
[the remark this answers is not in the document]
d3flex: https://github.com/openSUSE/qem-bot/pull/318/files#diff-f763aa73e4c05dd0729b511e4ec0d503523a3b7d21f3ce8b884dc28ad4005e7cR536

Readme.md
d3flex: this is unnecessary here. Everyone can use whatever wants to setup the env. I think a reason to add it, it could be if we had adapted it in our workflow and wanted to ensure that noone works with different packages. The instruction here, it has no really value.  It just another way but not related to the project. Can we remove it please?

openqabot/amqp.py
```diff
@@ -104,5 +104,5 @@ def handle_incident(self, inc_nr: int, message: dict[str, Any]) -> None:
         approve = Approver(self.args, inc_nr)
         approve()
 
-    def handle_aggregate(self, unused_build: str, unused_message: dict[str, Any]) -> None:  # noqa: ARG002 Unused method argument
```
Martchus: `_` is a terrible name for an argument. Since the arguments are unused you're removing almost the last bit of information of what these arguments are meant to be. (Technically one can still check the call-sites but I prefer if a function is self-explaining.)

Maybe there's something like `[maybe_unused]]` in C++?

If not, you can also simply use these arguments, e.g. to print a log message.

okurz: We can also keep the previous specific noqa exclusion comments in place. I still don't understand why we need to fill a method with unused arguments which does not do anything. Can I just delete the method and where it's called?

Martchus: You can also remove it if you like; it wasn't me who wrote this function. I think it was (and is) meant as a hint for future extension but if you don't like these kinds of structures in the code base we can remove it.

okurz: ok, removed

d3flex (approved):

PR #319, opened 2025-12-08
PART 12: More tests+use uv+use pytest-mock

okurz:
See https://github.com/openSUSE/qem-bot/pull/319/commits

After:
* #318
Before:
* #321

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ dev = [
 branch = true
 
 [tool.coverage.report]
-fail_under = 90
+fail_under = 94
 show_missing = true
 skip_covered = true
 


.gitignore
```diff
@@ -25,6 +25,7 @@ share/python-wheels/
 .installed.cfg
 *.egg
 MANIFEST
+uv.lock
```
mpagot: nop, I think it should be in git

TL;DR

For applications, `uv.lock` should be committed to version control for reproducible, cross-platform environments.
The official layout guide recommends this: https://docs.astral.sh/uv/concepts/projects/layout/#the-lockfile
`uv` automatically manages `uv.lock`; `uv sync` uses it to create identical environments for all users.
This behavior is detailed at: https://docs.astral.sh/uv/concepts/projects/sync/
For libraries, committing `uv.lock` is also the recommended practice for development consistency.
This provides a stable, reproducible environment for all contributors and for your primary CI tests.
To ensure broad compatibility, your CI should also test against different dependency resolutions.
Use `uv sync --resolution lowest` in CI to verify your declared lower dependency bounds.
Use `uv sync --upgrade` in a separate CI job to test against the latest compatible versions.
These settings are explained at: https://docs.astral.sh/uv/reference/settings/#resolution

mpagot:
Generally speaking my personal feeling it that this PR is too big, I'd prefer to have one only containing:
- .github/workflows/ci.yml
- pyproject.toml
- uv.lock

tests/test_incidents.py
```diff
@@ -442,3 +438,75 @@ def test_gitea_incidents() -> None:
...
+        "inReview": True,
+        "isActive": False,
+        "inReviewQAM": False,
+        "approved": True,
+        "embargoed": False,
+        "packages": ["foo"],
+        "channels": ["SUSE:Updates:SLE-Product-SLES:15-SP3:x86_64"],
+        "emu": False,
+        "type": "git",
+    }
```
Wabri: question: are all of this arguments needed for the test to pass?

tests/test_incidents.py
```diff
@@ -442,3 +438,75 @@ def test_gitea_incidents() -> None:
...
+        "approved": False,
+        "embargoed": False,
+        "packages": ["foo"],
+        "channels": ["SUSE:Updates:SLE-Product-SLES:15-SP3:x86_64"],
+        "emu": False,
+        "type": "smelt",
+    }
+    inc = Incident(inc_data)
+
+    test_config = {"FLAVOR": {"AAA": {"archs": ["x86_64"], "issues": {"OS_TEST_ISSUES": "SLE-Product-SLES:15-SP3"}}}}
```
Wabri: suggestion: maybe more readable like this?

```suggestion
    test_config = {
        "FLAVOR": {
            "AAA": {
                "archs": ["x86_64"],
                "issues": {
                    "OS_TEST_ISSUES": "SLE-Product-SLES:15-SP3"
                }
            }
        }
    }
```

Add more lines, but I would prefer readability over length.

Wabri (changes requested):

kalikiana (approved):

tests/test_incrementapprover.py
```diff
@@ -519,3 +519,21 @@ def test_get_regex_match_invalid_pattern(caplog: pytest.LogCaptureFixture, monke
...
+        req.state = type("state", (), {"to_xml": lambda: True})
+        req.reviews = [ReviewState("review", OBS_GROUP)]
+        req.to_str = lambda: "<request />"
+        return req
+
+    monkeypatch.setattr(osc.core.Request, "from_api", fake_request_from_api)
+    approver = prepare_approver(caplog, monkeypatch, request_id=43)
+    approver._find_request_on_obs("foo")  # noqa: SLF001
+    assert "Checking specified request 43" in caplog.text
+    assert "<request />" in caplog.text
```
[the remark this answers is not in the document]
okurz: The change you are commenting is actually an older, outdated diff which is already in master. I amended a commit in this PR and pushed the update and now the diff doesn't show anymore.

PR #321, opened 2025-12-08
PART 13: Use simpler mocking+more checks - After #319

okurz:
After:
* #319
Before:
* #327

### DIFF
--- a/tests/test_pc_helper.py
+++ b/tests/test_pc_helper.py
@@ -38,7 +38,7 @@ def test_pint_query_uses_cache(mocker: MockerFixture) -> None:
     get_mock.assert_called_once()
 
 
-def test_apply_publiccloud_pint_image(mocker: MockerFixture) -> None:
+def test_apply_publiccloud_pint_image_empty_settings(mocker: MockerFixture) -> None:
     mocker.patch("openqabot.pc_helper.pint_query", side_effect=lambda *_args, **_kwargs: {"images": []})
     mocker.patch(
         "openqabot.pc_helper.get_recent_pint_image",
@@ -53,6 +53,8 @@ def test_apply_publiccloud_pint_image(mocker: MockerFixture) -> None:
     assert "PUBLIC_CLOUD_PINT_FIELD" not in settings
     assert "PUBLIC_CLOUD_REGION" not in settings
 
+
+def test_apply_publiccloud_pint_image_simple_settings(mocker: MockerFixture) -> None:
     mocker.patch("openqabot.pc_helper.pint_query", side_effect=lambda *_args, **_kwargs: {"images": []})
     mocker.patch(
         "openqabot.pc_helper.get_recent_pint_image",
@@ -71,6 +73,8 @@ def test_apply_publiccloud_pint_image(mocker: MockerFixture) -> None:
     assert "PUBLIC_CLOUD_PINT_FIELD" not in settings
     assert "PUBLIC_CLOUD_REGION" not in settings
 
+
+def test_apply_publiccloud_pint_image_different_region(mocker: MockerFixture) -> None:
     mocker.patch("openqabot.pc_helper.pint_query", side_effect=lambda *_args, **_kwargs: {"images": []})
     mocker.patch(
         "openqabot.pc_helper.get_recent_pint_image",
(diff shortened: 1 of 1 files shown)


okurz:
Fixed the branch coverage regression in openqabot/main.py and added three commits

openqabot/commenter.py
```diff
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: MIT
 from __future__ import annotations
 
+from argparse import Namespace
```
Wabri: thought: I think this will not cause any issue on my current work on https://github.com/openSUSE/qem-bot/pull/322 but I'll rebase anyway 👍

tests/test_approve.py
```diff
@@ -817,57 +817,53 @@ def f_patch(*_args: Any, **_kwds: Any) -> NoReturn:
...
+@pytest.mark.parametrize(
+    ("older_jobs_data", "log_message"),
+    [
+        pytest.param(
+            {"build": "20240101XY", "result": "failed", "id": 123},
+            "Older usable jobs did not succeed",
+            id="no_suitable_older_jobs",
+        ),
+        pytest.param(
+            {"build": "invalid-date"},
```
Martchus: So again an invalid date? Why do we need that twice? If we need it twice it would make sense to generalize the name of this test function accordingly.

Makefile
```diff
@@ -17,6 +17,14 @@ tidy:
 	ruff format
 	ruff check --fix
 
+.PHONY: check-architecture
+check-architecture:
```
Martchus: What has the use of a certain decorator in our unit tests to do with the "architecture"?

Makefile
```diff
@@ -17,6 +17,14 @@ tidy:
...
+	@if git grep -nE '^\s*@(unittest\.mock\.|mock\.)?patch' tests/; then \
+		echo "❌ Error: @patch decorator detected. Avoid to prevent argument ordering bugs."; \
```
Martchus: Not sure whether I like using an ❌ like this. It looks a bit out of place.

openqabot/loader/incrementconfig.py
```diff
@@ -75,6 +75,9 @@ def from_config_file(file_path: Path) -> Iterator["IncrementConfig"]:
                 IncrementConfig.from_config_entry,
                 YAML(typ="safe").load(file_path).get("product_increments", []),
             )
+        except AttributeError:
+            log.debug("Ignoring file '%s' invalid for incrementconfig", file_path)
```
Martchus:
```suggestion
            log.debug("Ignoring file '%s' as it contains no valid increment config", file_path)
```

openqabot/amqp.py
```diff
@@ -66,18 +66,17 @@ def on_message(
...
-                log.debug("Received AMQP message: %s", pformat(message))
-                log.info("Job for incident %s done", inc_nr)
-                return self.handle_incident(inc_nr, message)
-            match = build_agg_regex.match(message["BUILD"])
-            if match:
-                build_nr = match.group(0)
-                log.debug("Received AMQP message: %s", pformat(message))
-                log.info("Aggregate build %s done", build_nr)
+        if method.routing_key != "suse.openqa.job.done" or "BUILD" not in message:
+            return None
```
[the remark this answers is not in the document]
okurz: this is weird but you again found a diff which is not part of this PR. Also github does not mention your comment as "outdated" so something fishy is going on but I can't retroactively change this diff here.

Makefile
okurz: I could write "design". Is that better?

tests/test_amqp.py
```diff
@@ -44,19 +44,20 @@ def test_init_no_url() -> None:
     amqp.stop()
 
 
-def test_call() -> None:
-    with patch("openqabot.amqp.pika"):
```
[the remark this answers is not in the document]
Martchus: I guess "Use pytest-mock consistently" is good enough. Actually, the statement that this is intentional and we will  not run around in circles here is also good enough.

Makefile
okurz: "conventions" sounds good. Done

okurz:
@Wabri please re-review. I tried to address your points. @Martchus all your points addressed except for the open discussion https://github.com/openSUSE/qem-bot/pull/321#discussion_r2614380486

Martchus (approved):

tests/test_openqabot.py
```diff
@@ -25,69 +25,37 @@ def test_no_args_prints_help() -> None:
...
     mock_args.token = "dummy_token"  # noqa: S105
     mock_args.func = MagicMock(return_value=0)
-    monkeypatch.setattr("openqabot.args.ArgumentParser.parse_args", MagicMock(return_value=mock_args))
-    mock_logger = MagicMock()
-    mock_logger.setLevel = MagicMock()
-    monkeypatch.setattr("openqabot.main.create_logger", MagicMock(return_value=mock_logger))
-    monkeypatch.setattr(sys, "exit", MagicMock(side_effect=SystemExit))
-
-    with contextlib.suppress(SystemExit):
+    mock_args.debug = True
```
Wabri: suggestion: is it worth to make `mock_args` a fixture?

okurz: so we have two test functions with "mock_args" but quite different so I don't see the benefit and I recommend against it.

okurz:
@Wabri I incorporated your suggestions while trying to not change too much which would invalidate the approval of others. I would appreciate your quick approval so that others can rebase on top of this older still open PR.

PR #324, opened 2025-12-12
Fix direction of comparison in source report based repo diff

Martchus:
* Swap "repo a" and "repo b" so that:
    * "repo a" is the *target* project of the request action (e.g. `SUSE:Products:SLE-Product-SLES:16.0:aarch64`)
    * "repo b" is the *source* project of the request action (e.g. `SUSE:SLFO:Products:SLES:16.0:TEST`)
    * Note that from the point of view of the request the *source* is the request report and the *target* is the regular report. For diffing however, we need to consider the regular report "version A" and the request report "version B".
* Avoid using `tgt_packages` and `src_packages` in `_compute_packages_of_request_from_source_report` as this naming leads to confusion
* Tested against request https://build.suse.de/request/show/396740 which is open at the time of making this request and I've got results which are consistent with the Perl prototype.
* See https://progress.opensuse.org/issues/190542#note-61

### DIFF
--- a/openqabot/incrementapprover.py
+++ b/openqabot/incrementapprover.py
@@ -151,12 +151,15 @@ class IncrementApprover:
     def _compute_packages_of_request_from_source_report(
         self, request: osc.core.Request | None
     ) -> tuple[defaultdict[str, set[Package]], int]:
-        tgt_packages = defaultdict(set)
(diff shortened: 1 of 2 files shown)


Martchus:
Thanks for merging this fast but the PI has been declined anyway.

PR #325, opened 2025-12-16
Revert "loader/gitea: Simplify add_packages_from_files"

Martchus:
This reverts commit e67fb259c0b91b063f2bcdc049d5521798cff196 as it changed `add_packages_from_files` to only consider one `_patchinfo` file which leads to missing packages.

---

@okurz Note that I first wanted to create a follow-up commit to re-establish what you were trying to achieve here. However, your version is not a simplification at all and much harder to read. So I'd just like to change this back to how it was before. (Your version needs twice as many lines of code, needs one more `if` and uses more complicated abstractions. How on earth is that "simpler"?)

### DIFF
--- a/tests/test_giteasync.py
+++ b/tests/test_giteasync.py
@@ -20,6 +20,7 @@ from openqabot.config import OBS_DOWNLOAD_URL, OBS_URL, QEM_DASHBOARD
 from openqabot.giteasync import GiteaSync
 from openqabot.loader.gitea import (
     add_build_results,
+    add_packages_from_files,
     compute_repo_url_for_job_setting,
     get_product_name,
     get_product_name_and_version_from_scmsync,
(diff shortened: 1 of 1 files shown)


openqabot/loader/gitea.py
```diff
@@ -437,16 +437,11 @@ def add_packages_from_patchinfo(
...
-        ),
-        None,
-    )
-    if patchinfo_file:
-        add_packages_from_patchinfo(incident, token, patchinfo_file["raw_url"], dry=dry)
+    for file_info in files:
+        file_name = file_info.get("filename", "").split("/")[-1]
+        raw_url = file_info.get("raw_url")
+        if file_name == "_patchinfo" and raw_url is not None:
+            add_packages_from_patchinfo(incident, token, raw_url, dry=dry)
```
okurz: I am fine to change that but I would prefer a less procedural and more functional style. You can write it differently but the "for file_info in files:" is for me a symptom that we should improve on. How about splitting the query from the action with a generator expression and processing that

```
patchinfo_urls = (f["raw_url"] for f in files if f.get("filename", "").split("/")[-1] == "_patchinfo" and f.get("raw_url"))
for url in patchinfo_urls:
    add_packages_from_patchinfo(incident, token, url, dry=dry)
```

okurz (changes requested):
> LGTM. I'd also recommend adding a code comment that there may be multiple `_patchinfo` files in the same update and all need to be processed.

I would prefer a test over a comment.

Can you add a test that is apparently missing for the "we need to process all files" part?

nicksinger (approved):
As this seems to fix an actual occurring problem, I'm approving - we had this code anyway, right? I think we can discuss "prettiness" of the code later and best in a small PR

openqabot/loader/gitea.py
Martchus: I'm really wondering why you hate good old for-loops so much. You say it isn't your preferred style but I'm wondering what actual problem you're trying to solve here. I'm also wondering why you like these generator expressions so much. They don't increase readability in my opinion (in fact they decrease readability). They can be nice to avoid intermediate arrays but I don't think this applies here.

kalikiana (approved):

PR #330, opened 2025-12-17
feat: Make number of retries overridable to ease development

Martchus:
During development it is usually better to get an error message immediately instead of having to wait several retries.

### DIFF
--- a/openqabot/utils.py
+++ b/openqabot/utils.py
@@ -81,10 +81,14 @@ def merge_dicts(dict1: dict[Any, Any], dict2: dict[Any, Any]) -> dict[Any, Any]:
     return copy
 
 
-def __retry(retries: int | None, backoff_factor: float) -> Session:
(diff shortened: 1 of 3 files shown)


openqabot/utils.py
```diff
@@ -80,10 +81,14 @@ def merge_dicts(dict1: dict[Any, Any], dict2: dict[Any, Any]) -> dict[Any, Any]:
     return copy
 
 
+def number_of_retries(fallback: int) -> int:
+    return 0 if "PYTEST_VERSION" in os.environ else int(os.environ.get("QEM_BOT_RETRIES", fallback))
```
okurz: How about specifying the default already here

```suggestion
def number_of_retries(fallback: int = 3) -> int:
    return 0 if "PYTEST_VERSION" in os.environ else int(os.environ.get("QEM_BOT_RETRIES", fallback))
```

kalikiana (approved):

PR #332, opened 2025-12-18
feat(gitea-sync): Support SL-Micro

Martchus:
Related ticket: https://progress.opensuse.org/issues/192523

---

A draft because tests are missing and we maybe want to make this configurable instead of just adding another replacement. However, we could also merge this in case we want to take time pressure for this feature. It would also allow us to gather more data from production sooner. We've also already [started working on tests](https://github.com/openSUSE/qem-bot/compare/master...Martchus:qem-bot:fix-sl-micro-tests?expand=1).


okurz (approved):

nicksinger (approved):

okurz:
rebased

PR #337, opened 2026-01-06
PART 20: Use "ty" by default - After #336

okurz:
After:
* #336

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,4 @@
 SOURCE_FILES ?= $(shell git ls-files "**.py")
-USE_TY_TYPECHECK ?= 0
 ISOLATE ?= 1
 
 ifeq ($(ISOLATE),1)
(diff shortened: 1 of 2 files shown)


openqabot/openqabot.py
```diff
@@ -51,9 +51,9 @@ def post_openqa(self, data: dict[str, Any]) -> None:
 
     def __call__(self) -> int:
         log.info("Entering bot main loop")
-        post: list[dict[str, Any]] = []
-        for worker in self.workers:
-            post += worker(self.incidents, self.token, self.ci, ignore_onetime=self.ignore_onetime)
+        post = [
```
aylen384: can we still have the type hint?

aylen384 (approved):

openqabot/openqabot.py
okurz: That should not be necessary anymore after this change. In before the variable was declared and initialized as an empty array and then processed later. Now it's an immediate assignment of the resulting value to a local value so IMHO we shouldn't need the type hint. We also don't commonly add type hints in other places for local variables and I guess that's a good compromise and common practice also enforced by the current set of rules.

PR #339, opened 2026-01-08
feat(repodiff): Allow computing diff of production SL Micro repo

Martchus:
The zstandard module does not seem to be capable of decompressing the XML files used in production. To support this we need to use pystd after all. This effectively reverts 5e2bf52fae60c40efd7d3e4a59844ee1c933b3ce.

### DIFF
--- a/openqabot/repodiff.py
+++ b/openqabot/repodiff.py
@@ -10,8 +10,8 @@ from logging import getLogger
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, NamedTuple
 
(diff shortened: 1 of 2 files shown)


okurz (approved):
> Not sure whether we can merge this as I created [5e2bf52](https://github.com/openSUSE/qem-bot/commit/5e2bf52fae60c40efd7d3e4a59844ee1c933b3ce) to still support Leap 15.6.

I guess we can as we bumped our requirements to newer versions of Python and other dependencies anyway.

PR #343, opened 2026-01-08
fix(loader/repohash): Handle missing repohash file gracefully

okurz:
As qem-bot tries to compute the repohash for all possible combinations
of products and architectures which do not exist we should handle such
cases gracefully.

This fixes the recently encountered case that nearly no gitea pull
requests are triggering openQA tests anymore.

Related progress issue: https://progress.opensuse.org/issues/194453

### DIFF
--- a/tests/test_repohash.py
+++ b/tests/test_repohash.py
@@ -136,7 +136,18 @@ def test_get_max_revision_retry_error(caplog: pytest.LogCaptureFixture) -> None:
     with pytest.raises(NoRepoFoundError):
         rp.get_max_revision(repos, arch, project)
 
-    assert "Incident skipped: RepoHash metadata not found at" in caplog.records[0].msg
+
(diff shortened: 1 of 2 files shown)


tests/test_repohash.py
```diff
@@ -136,7 +136,24 @@ def test_get_max_revision_retry_error(caplog: pytest.LogCaptureFixture) -> None:
...
+
+    responses.add(
+        responses.GET,
+        url="http://download.suse.de/ibs/SUSE:/Maintenance:/12345/SUSE_Updates_SLES_15SP3_x86_64/repodata/repomd.xml",
+        status=404,
+    )
+
+    ret = rp.get_max_revision(repos, arch, project)
+    assert ret == 0
+
```
Wabri:
```suggestion
```

Wabri (approved):

PR #348, opened 2026-01-12
PART 25: feat: rename incidents to submissions to be more generic - After #347

okurz:
Related issues:
* https://progress.opensuse.org/issues/194453
* https://progress.opensuse.org/issues/192904

After:
* #347
Before:
* #349

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -22,17 +22,17 @@ updates information about incidents and related openQA tests.
 
     positional arguments:
       {full-run,incidents-run,updates-run,smelt-sync,gitea-sync,inc-approve,inc-comment,inc-sync-results,aggr-sync-results,increment-approve,repo-diff,amqp}
-        full-run            Full schedule for Maintenance Incidents in openQA
-        incidents-run       Incidents only schedule for Maintenance Incidents in
-                            openQA
-        updates-run         updates only schedule for Maintenance Incidents in
-                            openQA
+        full-run            Full schedule for Maintenance Submissions in openQA
+        incidents-run       Submissions only schedule for Maintenance Submissions
+                            in openQA
+        updates-run         Aggregates only schedule for Maintenance Submissions
+                            in openQA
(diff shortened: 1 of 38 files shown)


Martchus (commented):
I'm not against this - although I'm also really wondering what we gain from that. Before we had one term ("incident") and now we have two terms ("incident" and "submission") for the same thing. That seems worse. Or do you think people will stop calling incidents "incidents" because we change the variable names and help text of qem-bot? How does SMELT call Gitea-based incidents? How are they referred to in Wiki pages?

okurz:
> > How does SMELT call Gitea-based incidents?
> 
> Apparently also "submissions": https://smelt.suse.de/slfo-beta/submissions So that's in accordance at least. And it means "submission" is already in use.

that's actually a coincidence. I wasn't aware. But keep in mind that also there submissions do not seem to be only maintenance updates. So I guess we can both (SMELT+qem-bot) use the more generic term "submissions" without necessarily implying that those are always the same.

> > How are they referred to in Wiki pages?
> 
> Looks like https://en.opensuse.org/openSUSE:Maintenance_update_process hasn't been updated yet for Leap 16. I'm also not sure what other Wiki/Confluence pages to consider.

And maybe it would just call it "updates" then. So https://github.com/openSUSE/qem-bot/pull/349 as a follow-up is further distinguishing between SMELT based incidents and gitea PRs hence needing the generalization of "submissions" first. The additional use case I was thinking of is https://progress.opensuse.org/issues/192427 "openQA tests for gitea-based SLFO product development "staging" pull requests". And also IBS requests for product increments which are also not "incidents". So in qem-bot terminology we would say: qem-bot can handle any kind of submissions. So far used are the following kind of submissions:
1. SMELT based maintenance incidents
2. openSUSE updates
3. gitea PRs (maintenance)
4. gitea PRs (product development staging)
5. product increment IBS requests

PR #349, opened 2026-01-12
PART 26-2: disambiguate different kind of submissions (smelt+gitea) - After #348

okurz:
Related issues:
* https://progress.opensuse.org/issues/194453
* https://progress.opensuse.org/issues/192904

After:
* #348
* #354
* #357
Before:
* #352

### DIFF
--- a/openqabot/amqp.py
+++ b/openqabot/amqp.py
@@ -13,6 +13,7 @@ import pika.channel
 import pika.spec
 
 from .approver import Approver
+from .config import DEFAULT_SUBMISSION_TYPE
 from .loader.qem import get_submission_settings_data
 from .syncres import SyncRes
 from .types.types import Data
@@ -23,7 +24,7 @@ if TYPE_CHECKING:
 
 
 log = getLogger("bot.amqp")
-build_sub_regex = re.compile(r":([^:]+):(\d+):.*")
+build_sub_regex = re.compile(r":(?:(?P<type>[^:]+):)?(?P<id>\d+)(?::.*)?")
 build_agg_regex = re.compile(r"\d{8}-\d+")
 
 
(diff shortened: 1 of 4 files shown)


tests/test_loader_gitea.py
```diff
@@ -39,7 +39,7 @@ def test_get_open_prs_metadata_error(mocker: MockerFixture, caplog: pytest.LogCa
     mocker.patch("openqabot.loader.gitea.get_json", side_effect=requests.RequestException("error"))
     res = gitea.get_open_prs({}, "repo", dry=False, number=124)
     assert res == []
-    assert "PR #124 ignored: Could not read PR metadata" in caplog.text
+    assert "PR git:124 ignored: Could not read PR metadata" in caplog.text
```
Martchus: Maybe the "PR " is not a bit redundant can should be removed.

openqabot/args.py
```diff
@@ -155,6 +155,13 @@ def get_parser() -> ArgumentParser:
         action="store_true",
         help="Ignore onetime and schedule those test runs",
     )
+    cmdfull.add_argument(
+        "-I",
+        "--submission",
+        required=False,
+        type=str,
+        help="Submission ID (to process only a single submission)",
+    )
```
Martchus: That's probably going to be very useful, indeed.

Martchus (approved):
Looks generally good. I only have a few remarks.

openqabot/args.py
okurz: Fixd the issue that @nicksinger mentioned

Readme.md
```diff
@@ -16,23 +16,31 @@ updates information about incidents and related openQA tests.
...
-                            openQA
-        updates-run         updates only schedule for Maintenance Incidents in
-                            openQA
+      {full-run,submissions-run,incidents-run,updates-run,smelt-sync,gitea-sync,sub-approve,inc-approve,sub-comment,inc-comment,sub-sync-results,inc-sync-results,aggr-sync-results,increment-approve,repo-diff,amqp}
+        full-run            Full schedule for Maintenance Submissions in openQA
+        submissions-run     Submissions only schedule for Maintenance Submissions
+                            in openQA
+        incidents-run       DEPRECATED: Submissions only schedule for Maintenance
+                            Submissions in openQA (use submissions-run)
+        updates-run         Aggregates only schedule for Maintenance Submissions
```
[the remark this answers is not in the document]
okurz: This one is a bit different though. So far this command is I guess much more "maintenance" specific then the submissions-run. In case of "submissions-run" we also handle gitea PRs which don't need to be maintenance updates and we don't schedule aggregates for those same as we likely wouldn't do for product development staging tests which can also be triggered by "submissions-run". So for now I would favor to keep "updates-run". Also "aggregates" is a full 3 characters longer … :)

okurz:
I split out parts into #354 and added additional minor features+fixes there as the "disambiguate different kind of submissions" is more invasive. So #354 first, then continue here, then #352

PR #373, opened 2026-02-03
Apply default configuration for IncrementConfig

grisu48:
Apply a defined default config to any `IncrementConfig` items, as this has been default behaviour for other configuration files.

See any older configuration files in the `metadata` repository for examples where this has been used in the past.

* Related ticket: https://progress.opensuse.org/issues/189024

### DIFF
--- a/tests/test_loader_incrementconfig.py
+++ b/tests/test_loader_incrementconfig.py
@@ -69,19 +69,17 @@ def test_config_parsing_from_args() -> None:
     assert config[0].settings == {}
 
 
-def test_config_parsing_from_args_with_path() -> None:
+@pytest.mark.parametrize(("config_index", "expected_distri"), [(0, "foo"), (1, "bar")])
+def test_config_parsing_from_args_with_path(config_index: int, expected_distri: str) -> None:
     path = Path("tests/fixtures/config-increment-approver")
     configs = IncrementConfig.from_args(
         Namespace(increment_config=path, distri="sle", version="16.0", flavor="Online-Increments")
     )
     assert len(configs) == 2
-    assert configs[0].distri == "foo"
-    assert configs[1].distri == "bar"
-
+    config = configs[config_index]
+    assert config.distri == expected_distri
     additional_settings = {
         "ADDITIONAL_SETTING1": "present",
         "ADDITIONAL_SETTING2": "also here",
     }
-    for config in configs:
-        assert additional_settings.items() <= config.settings.items()
-        assert additional_settings.items() <= config.settings.items()
+    assert additional_settings.items() <= config.settings.items()


tests/test_loader_incrementconfig.py
d3flex: likely you could use pytest.mark.parametrize here. it would simplify the test and remove redundant code

openqabot/loader/incrementconfig.py
```diff
@@ -76,19 +76,30 @@ def from_config_entry(entry: dict[str, Any]) -> "IncrementConfig":
...
+                    IncrementConfig.from_config_entry,
+                    yaml.get("product_increments", []),
+                )
             )
+            # Apply default settings to all items
+            if load_defaults:
+                defaults = yaml.get("settings", {})
+                for item in items:
+                    for setting in defaults:
+                        item.settings.setdefault(setting, defaults[setting])
```
okurz: This is just updating the complete dict so let's use a single comprehension call

```suggestion
                [item.settings.update({**defaults, **item.settings}) for item in items]
```

tests/test_loader_incrementconfig.py
```diff
@@ -69,9 +69,17 @@ def test_config_parsing_from_args() -> None:
 
 def test_config_parsing_from_args_with_path() -> None:
```
d3flex:
```suggestion
@pytest.mark.parametrize(
    ("config_index", "expected_distri"), [(0, "foo"), (1, "bar")])
def test_config_parsing_from_args_with_path(config_index: int, expected_distri: str) -> None:
    path = Path("tests/fixtures/config-increment-approver")
    configs = IncrementConfig.from_args(
        Namespace(increment_config=path, distri="sle", version="16.0", flavor="Online-Increments")
    )
    assert len(configs) == 2

    config = configs[config_index]
    assert config.distri == expected_distri

    additional_settings = {
        "ADDITIONAL_SETTING1": "present",
        "ADDITIONAL_SETTING2": "also here",
    }
    assert additional_settings.items() <= config.settings.items()
```

Note that I havent tested myself

d3flex: I see. The problem is that if someone not ask for those improvements, it turns to either be neglected and to anchor on a less favorable style, as the next committer will see that this, claim the same and argue that that is not "how others doing" or "it is not important request". I ask the changes because it would be easier to merge without another round of reviews. 

I already provide the code. did you tried it out? or would you like me to add a commit for it, so we do not keep this PR open anymore?

grisu48:
> I push a change and I hope you dont mind, as you argue that you need this urgently

Agreed with this proposal, thank you!

openqabot/loader/incrementconfig.py
```diff
@@ -89,20 +89,29 @@ def from_config_entry(entry: dict[str, Any]) -> IncrementConfig:
...
+            items = list(
+                map(
+                    IncrementConfig.from_config_entry,
+                    yaml.get("product_increments", []),
+                )
             )
+            # Apply default settings to all items
+            if load_defaults:
+                defaults = yaml.get("settings", {})
+                [item.settings.update({**defaults, **item.settings}) for item in items]
```
Martchus: Won't this construct an array which we then don't actually use? Or is this some Python-magic? I'd prefer a good old for loop here.

PR #374, opened 2026-02-03
feat(approver): distinguish between failed, waiting, and stopped jobs

okurz:
Introduce JobStatus enum to provide more accurate feedback when a
submission is not approved. This allows the bot to distinguish
between actual test failures, tests still in progress, and tests
that were stopped.

Detailed logging has been added to identify specific unfinished or
stopped jobs, and tests have been updated to verify this behavior.

Before:
* #368

### DIFF
--- a/openqabot/approver.py
+++ b/openqabot/approver.py
@@ -6,6 +6,7 @@ import re
 import string
 from argparse import Namespace
 from datetime import datetime, timedelta
+from enum import IntEnum, unique
 from functools import lru_cache
 from http import HTTPStatus
 from logging import getLogger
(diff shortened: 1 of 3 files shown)


openqabot/approver.py
```diff
@@ -277,9 +303,9 @@ def mark_jobs_as_acceptable_for_submission(self, job_results: list[dict], sub: i
                 job_result[f"acceptable_for_{sub}"] = True
                 self.mark_job_as_acceptable_for_submission(job_id, sub)
 
-    def is_job_acceptable(self, sub: int, api: str, job_result: dict) -> bool:
+    def is_job_acceptable(self, sub: int, api: str, job_result: dict) -> JobStatus:
         if self.is_job_passing(job_result):
```
asmorodskyi: With introduction of JobStatus enum this function name became confusing . because job result the same as job status can be PASSED .

In openQA we also have Job state and result but in that case there is very clear distinction .

PR #420, opened 2026-03-04
WIP: Update note of scheduled product when handling increment approval

Martchus:
Related ticket: https://progress.opensuse.org/issues/193843

---

This is still WIP because tests are still failing as I need to add mocking for the new API call.


asmorodskyi (changes requested):
see comments inline

okurz:
I guess we have ditched this concept. I recommend to close this PR

PR #433, opened 2026-03-11
[WIP][RFC] feat(types): harden product version with validated ProductVersion type

okurz:
Motivation:
Product version extraction used plain strings and empty strings for missing
values, causing ambiguity and potential errors. PR #429 feedback requested
stricter definitions and better type safety in
https://github.com/openSUSE/qem-bot/pull/429#pullrequestreview-3920665752

Design Choices:
- Introduced ProductVersion string subclass in types.py with regex-based
  validation in __new__.
- Refactored gitea.py to return ProductVersion | None instead of "".
- Moved VERSION_EXTRACT_REGEX to types.py to centralize logic.
- Updated extraction logic to use fullmatch for stricter validation.
- Updated dependent code to use explicit is not None checks.

User Benefits:
- Improved type safety and runtime validation of product versions.
- Clearer API contracts with Optional types.
- Centralized version parsing logic improves maintainability.

### DIFF
--- a/openqabot/loader/gitea.py
+++ b/openqabot/loader/gitea.py
@@ -32,7 +32,19 @@ from openqabot.utils import retry10 as retried_requests
 if TYPE_CHECKING:
     from collections.abc import Iterator
 
+<<<<<<< HEAD
     from openqabot.types.types import Repos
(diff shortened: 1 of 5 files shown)


okurz:
not following up with this non-important branch

PR #478, opened 2026-04-10
fix: prevent duplicate approval comments on Gitea PRs

okurz:
Motivation:
Gitea PRs were receiving multiple duplicate approval comments from the bot
whenever the gitea-trigger ran and tests were already passed.

Design Choices:
- Implement a check in approve_pr to fetch existing reviews/comments.
- Verify both official Gitea APPROVED reviews and bot-specific comments.
- Match against the specific Tested commit ID to allow re-approval if the
  commit changes.

Benefits:
- Reduces spam on Gitea PRs.
- Consistent behavior across periodic trigger runs.

Related issue: https://progress.opensuse.org/issues/197906

### DIFF
--- a/openqabot/loader/gitea.py
+++ b/openqabot/loader/gitea.py
@@ -218,6 +218,19 @@ def get_open_prs(token: dict[str, str], repo: str, *, number: int | None) -> lis
         return []
 
 
+def _approval_identifiers(bot_user: str, commit_id: str, *, approve: bool = True) -> tuple[str, str]:
+    action = "approved" if approve else "decline"
+    return f"@{bot_user}: {action}", f"Tested commit: {commit_id}"
+
+
+def _is_bot_approval_comment(comment: dict[str, Any], bot_user: str, commit_id: str) -> bool:
+    """Check if a comment is an authentic approval from the bot."""
+    body = comment.get("body", "")
+    is_author = comment.get("user", {}).get("login") == bot_user
+    review_cmd, commit_str = _approval_identifiers(bot_user, commit_id)
+    return is_author and review_cmd in body and commit_str in body
+
+
 def review_pr(  # noqa: PLR0913
     token: dict[str, str],
     repo_name: str,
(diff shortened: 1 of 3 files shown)


Martchus (approved):

openqabot/loader/gitea.py
```diff
@@ -247,6 +260,21 @@ def review_pr(  # noqa: PLR0913
 def approve_pr(token: dict[str, str], repo_name: str, pr_number: int, commit_id: str, msg: str) -> bool:
     """Approve a PR on Gitea using its repository name and commit ID."""
     try:
+        bot_user = config.settings.git_review_bot_user
```
asmorodskyi, line 263: **question(non-blocking)** : just for the clarity - we still don't know why decision what to approve is based on the fact if this user is defined ?

asmorodskyi (approved):
LGTM

openqabot/loader/gitea.py
okurz: Sorry, can't parse your sentence. Please rephrase

asmorodskyi: below this line you have if/else where we deciding what to process comments of PR or some "reviews" ( no idea what is this ) . And this decision made based on a fact if `config.settings.git_review_bot_user` defined or not . If I would need refactor this function I would need to understand WHY this is done this way . Would be nice to answer this question sooner than later ...

asmorodskyi: 1. what else is changed when I define/not define `git_review_bot_user` ? 
2. In which commands it is used/not used ?

asmorodskyi: P.S. ofc. this questions not directly related to this PR and also you not obligated to find me answers . I am asking just in case you know them ...

okurz: @Martchus can you help to answer the question as I assume you originally introduced this?

PR #480, opened 2026-04-13
perf: filter directory listing on server side using P=*spdx.json

okurz:
Motivation:
To reduce the amount of data transferred and noise in logs, use the P=
query parameter to filter the directory listing to only relevant .spdx.json
files at the IBS/OBS source.

Design Choices:
- Added ?P=*spdx.json to the directory listing URL in load_build_info.
- Updated test helper to reflect the new URL format.

Benefits:
- Faster artifact discovery.
- Reduced network overhead.

### DIFF
--- a/openqabot/loader/buildinfo.py
+++ b/openqabot/loader/buildinfo.py
@@ -30,9 +30,10 @@ def load_build_info(
     """Determine build information from the project's repository listing."""
     build_project_url = config.build_project_url()
     sub_path = config.build_listing_sub_path
-    url = f"{build_project_url}/{sub_path}/?P=*spdx.json&jsontable=1"
-    log.debug("Checking for '%s' files on %s", build_regex, url)
-    rows = retried_requests.get(url).json().get("data", [])
+    url = f"{build_project_url}/{sub_path}/"
+    params = {"P": "*spdx.json", "jsontable": 1}
+    log.debug("Checking for '%s' files on %s with params %s", build_regex, url, params)
+    rows = retried_requests.get(url, params=params).json().get("data", [])
 
     def get_build_info_from_row(row: dict[str, Any]) -> BuildInfo | None:
         name = row.get("name", "")
--- a/openqabot/loader/buildinfo.py
+++ b/openqabot/loader/buildinfo.py
@@ -30,7 +30,7 @@ def load_build_info(
     """Determine build information from the project's repository listing."""
     build_project_url = config.build_project_url()
     sub_path = config.build_listing_sub_path
-    url = f"{build_project_url}/{sub_path}/?jsontable=1"
+    url = f"{build_project_url}/{sub_path}/?P=*spdx.json&jsontable=1"
     log.debug("Checking for '%s' files on %s", build_regex, url)
     rows = retried_requests.get(url).json().get("data", [])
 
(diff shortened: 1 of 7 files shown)


openqabot/loader/buildinfo.py
```diff
@@ -30,7 +30,7 @@ def load_build_info(
     """Determine build information from the project's repository listing."""
     build_project_url = config.build_project_url()
     sub_path = config.build_listing_sub_path
-    url = f"{build_project_url}/{sub_path}/?jsontable=1"
+    url = f"{build_project_url}/{sub_path}/?P=*spdx.json&jsontable=1"
```
nicksinger: building a URL like that seems prone to failures. IIRC `requests` supports buildings params via dicts but I assume we just pass the value here to something deeper which doesn't support this or could we use that mechanism somehow?

nicksinger (approved):
Approving despite the very manual way to construct a parametrized URL because we already did it like this in before.

openqabot/loader/buildinfo.py
kalikiana: I tested it with dist.suse.de, tho, and it does filter the listing as expected.

It might be nice to add a comment here referring to OBS since this is not very obvious.

okurz: will look into this

okurz: > I tested it with dist.suse.de, tho, and it does filter the listing as expected.
> 
> It might be nice to add a comment here referring to OBS since this is not very obvious.

But dist.suse.de and download.suse.de are not OBS per se. Not sure what you mean

okurz: done. Added a commit to refactor this. Other occurences of manual URL building have been adressed in #482

kalikiana: You used the term "OBS source" in your description. I don't know where this is documented and simply guessed based on job settings.

asmorodskyi (changes requested):
I would like to get more clear info from where this `P=` parameter comes from . and where it actually supported . Also I think we need some clear comments in code for future us about that

openqabot/loader/buildinfo.py
```diff
@@ -30,9 +30,10 @@ def load_build_info(
     """Determine build information from the project's repository listing."""
     build_project_url = config.build_project_url()
     sub_path = config.build_listing_sub_path
-    url = f"{build_project_url}/{sub_path}/?jsontable=1"
-    log.debug("Checking for '%s' files on %s", build_regex, url)
-    rows = retried_requests.get(url).json().get("data", [])
+    url = f"{build_project_url}/{sub_path}/"
+    params = {"P": "*spdx.json", "jsontable": 1}
```
Martchus: This should be configurable in the same way the regex pattern is.

asmorodskyi:
LGTM for me . keeping request changes until @kalikiana  will let me to remove it

okurz:
> I would like to get more clear info from where this `P=` parameter comes from . and where it actually supported . Also I think we need some clear comments in code for future us about that

I amended the last commit with a documented helper function to prevent any duplicating comments.

openqabot/loader/buildinfo.py
okurz: done, see https://github.com/openSUSE/qem-bot/pull/480/changes#diff-81268c2da9eaa6620c609f559e999635efe3de60658f00afab8b271556fb85a7R48

openqabot/loader/buildinfo.py
okurz: I extracted the query parameter building into a helper with according explanation. I hope that clarifies things.

kalikiana (approved):

asmorodskyi (approved):
LGTM

PR #484, opened 2026-04-13
Fix repeated comments on gitea-trigger approvals

okurz:
* fix: handle bot author mismatch in Gitea approval deduplication
* fix: avoid repeated Gitea PR comments with pagination and matching

Verified manually with a wet-run and a read-only gitea token using the command

```
./qem-bot.py --gitea-token … --debug gitea-trigger --pr-label "staging/In Progress"
```

which now yields log messages like

```
Evaluating PR 3225 for openQA triggering
openQA tests for PR 3225 (build PR-3225-3.19:SLES-16.1) are already covered
PR 3225 already approved via comment for commit …
```

Related issue: https://progress.opensuse.org/issues/197906

### DIFF
--- a/openqabot/commenter.py
+++ b/openqabot/commenter.py
@@ -164,13 +164,13 @@ class Commenter:
         # Add a marker so we can find our own comments later
         msg = add_marker(msg, "openqa", {"state": state})
 
-        comments = gitea.get_json_list(gitea.comments_url(repo, sub.id), self.gitea_token)
+        comments = gitea.iter_gitea_items(gitea.comments_url(repo, sub.id), self.gitea_token)
         formatted = {str(c["id"]): {"id": c["id"], "comment": c["body"]} for c in comments}
-        comment, _ = self.commentapi.comment_find(formatted, "openqa", {"state": state})
+        comment, info = self.commentapi.comment_find(formatted, "openqa")
 
         # To prevent spam, assume same state/result
         # and number of lines in message is a duplicate message
-        if comment and comment["comment"].count("\n") == msg.count("\n"):
+        if comment and info and info.get("state") == state and comment["comment"].count("\n") == msg.count("\n"):
             log.debug("Comment skipped: Previous comment is too similar")
             return
 
--- a/openqabot/loader/gitea.py
+++ b/openqabot/loader/gitea.py
@@ -153,19 +153,19 @@ def read_xml(name: str) -> etree.ElementTree:
 
 def reviews_url(repo_name: str, number: int) -> str:
     """Construct the URL for PR reviews."""
-    # https://docs.gitea.com/api/1.20/#tag/repository/operation/repolistPullReviews
+    # https://docs.gitea.com/api/1.25/#tag/repository/operation/repolistPullReviews
     return f"repos/{repo_name}/pulls/{number}/reviews"
 
 
 def changed_files_url(repo_name: str, number: int) -> str:
     """Construct the URL for PR changed files."""
-    # https://docs.gitea.com/api/1.20/#tag/repository/operation/repoGetPullRequestFiles
+    # https://docs.gitea.com/api/1.25/#tag/repository/operation/repoGetPullRequestFiles
     return f"repos/{repo_name}/pulls/{number}/files"
 
 
 def comments_url(repo_name: str, number: int) -> str:
     """Construct the URL for PR comments."""
-    # https://docs.gitea.com/api/1.20/#tag/issue/operation/issueCreateComment
+    # https://docs.gitea.com/api/1.25/#tag/issue/operation/issueCreateComment
     return f"repos/{repo_name}/issues/{number}/comments"
 
 
@@ -210,7 +210,7 @@ def get_open_prs(token: dict[str, str], repo: str, *, number: int | None) -> lis
         return [cast("dict[str, Any]", pr)]
 
     try:
-        # https://docs.gitea.com/api/1.20/#tag/repository/operation/repolistPullRequests
+        # https://docs.gitea.com/api/1.25/#tag/repository/operation/repolistPullRequests
         return list(iter_gitea_items(f"repos/{repo}/pulls?state=open", token))
     except (requests.exceptions.RequestException, TypeError):
         log.exception("Gitea API error: Could not fetch open PRs from %s", repo)
--- a/openqabot/loader/gitea.py
+++ b/openqabot/loader/gitea.py
@@ -80,24 +80,33 @@ def get_json(
     return response.json()
 
 
-def get_json_list(
-    query: str,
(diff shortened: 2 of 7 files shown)


openqabot/loader/gitea.py
```diff
@@ -189,32 +203,11 @@ def get_open_prs(token: dict[str, str], repo: str, *, number: int | None) -> lis
...
-            page += 1
-
     try:
-        return [pr for page in iter_pr_pages() for pr in page]
-    except requests.exceptions.JSONDecodeError:
-        log.exception("Gitea API error: Invalid JSON received for open PRs")
-        return []
-    except requests.exceptions.RequestException:
-        log.exception("Gitea API error: Could not fetch open PRs")
+        # https://docs.gitea.com/api/1.20/#tag/repository/operation/repolistPullRequests
```
asmorodskyi:
```suggestion
        # https://docs.gitea.com/api/1.25/#tag/repository/operation/repolistPullRequests
```

asmorodskyi (changes requested):
let's update URL

openqabot/commenter.py
```diff
@@ -168,9 +168,13 @@ def gitea_comment(self, sub: CommentableProtocol, msg: str, state: str) -> None:
         formatted = {str(c["id"]): {"id": c["id"], "comment": c["body"]} for c in comments}
         comment, _ = self.commentapi.comment_find(formatted, "openqa", {"state": state})
 
+        if comment is None:
```
asmorodskyi: **question(blocking):** why some comments have state and others don't ? Can we have single format to not search two times ?

openqabot/loader/gitea.py
```diff
@@ -226,7 +219,8 @@ def _approval_identifiers(bot_user: str, commit_id: str, *, approve: bool = True
 def _is_bot_approval_comment(comment: dict[str, Any], bot_user: str, commit_id: str) -> bool:
     """Check if a comment is an authentic approval from the bot."""
     body = comment.get("body", "")
-    is_author = comment.get("user", {}).get("login") == bot_user
+    allowed_authors = {bot_user, config.settings.obs_group}
```
asmorodskyi, line 228: **question(non-blocking)**: "obs_group" is author ?

tests/test_giteasync.py
```diff
@@ -174,7 +174,7 @@ def test_sync_with_product_repo(mocker: MockerFixture, caplog: pytest.LogCapture
     assert "Loaded 7 active PRs from products/SLFO" in caplog.messages
     assert "Fetching info for PR git:131 from Gitea" in caplog.messages
     assert "Syncing Gitea PRs to QEM Dashboard: Considering 1 submissions" in caplog.messages
-    assert len(responses.calls) == 25
+    assert len(responses.calls) == 24
```
Wabri, line 177: thought: For me this is a magic number, what is the meaning of 24?

tests/test_loader_gitea_helpers.py
```diff
@@ -157,18 +157,53 @@ def test_add_channel_for_build_result_local() -> None:
...
+    # First page response
+    mock_response1 = mocker.Mock()
+    mock_response1.json.return_value = [{"id": 1}]
+    mock_response1.links = {"next": {"url": "https://gitea.com/api/v1/some/query?page=2"}}
+
+    # Second page response
+    mock_response2 = mocker.Mock()
+    mock_response2.json.return_value = [{"id": 2}]
+    mock_response2.links = {}
+
```
Wabri:
```suggestion
```

Wabri (dismissed):

openqabot/loader/gitea.py
okurz: please stop requesting changes for code I did not change. This comment existed in before in https://github.com/openSUSE/qem-bot/pull/484/files#diff-9e182f20a85816dd88b523f1b5ee7bda5ee9f4933b319833e8ae8fb2b6dd80bfL201 . I think it would be way quicker if you just create your own PR on top of mine to bring in such unrelated changes.

tests/test_giteasync.py
okurz: same as the meaning of 25 in before. As the left-hand of the operator says this is the number of expected response calls which I managed to reduce by one

openqabot/loader/gitea.py
asmorodskyi: ok

tests/test_giteasync.py
Wabri: If you say so, ok!

openqabot/commenter.py
asmorodskyi: @okurz  just for the clarity currently I am still blocking this PR not because you don't want update URL ( will do this in separate PR as you suggested ) but because of this question

openqabot/loader/gitea.py
```diff
@@ -74,19 +74,33 @@ def get_json(query: str, token: dict[str, str], host: str | None = None) -> Json
...
 
 
-def get_json_list(query: str, token: dict[str, str], host: str | None = None) -> list[Any]:
-    """Fetch a list of JSON data from Gitea API."""
-    res = get_json(query, token, host)
-    if not isinstance(res, list):
-        msg = f"Gitea API returned {type(res).__name__} instead of list for query: {query}"
-        raise TypeError(msg)
-    return res
+def get_json_list(query: str, token: dict[str, str], host: str | None = None) -> Iterator[Any]:
```
d3flex: nits: `get_json_list` maybe was making sense as a shallow function. but now the name is far from what it represents. As always, it is open what is acceptable and how you would like to rename it. Even the ` iter_pr_pages` which was encapsulate part f this refactoring, sounds more helpful

tests/test_loader_gitea_helpers.py
okurz: done

openqabot/loader/gitea.py
okurz: this was done to make this consistent and more resilient but I am not sure if this is always needed. I removed the according change for now. Might bring it up in a follow-up

openqabot/commenter.py
okurz: I did update URLs in a separate commit in this PR now. Also, I changed this code now so that we don't search twice.

openqabot/loader/gitea.py
okurz: True. I renamed to `iter_gitea_items` as this is now a generalized re-used API result pagination handler.

okurz:
Updated to incorporate all review suggestions

asmorodskyi (approved):
LGTM

d3flex (approved):

Martchus (approved):

PR #485, opened 2026-04-14
Add codeowners to make sure that I notified about all PRs

asmorodskyi:


okurz (changes requested):
I don't see the purpose and I think it's wrong. As a QE infra member you need to subscribe to the complete project so you *will* be notified about changes. If you ignore those notifications then any code owners file will not fix that

asmorodskyi:
> I don't see the purpose and I think it's wrong. As a QE infra member you need to subscribe to the complete project so you _will_ be notified about changes. If you ignore those notifications then any code owners file will not fix that

1. can clearly state what **problems** this file will introduce to others so you blocking it ?
2. I don't want to use github notifications because they are too spammy and I am not getting emails like "new PR opened in qem-bot"  and I would like to have them

okurz (approved):

.github/CODEOWNERS
```diff
@@ -0,0 +1,9 @@
+# GitHub CODEOWNERS https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
+# This file is used as a workaround to ensure that users defined here are
+# automatically notified and assigned as a reviewer to all
+# Pull Requests created in this repository.
+#
+# While CODEOWNERS is typically for defining domain expertise,
+# it is used here globally to automate the PR notification process.
+
+@asmorodskyi
```
kalikiana, line 9: Maybe I don't completely understand this. Why is this just you and not the whole team? Or the @openSUSE/qa-tools-team even?

okurz: because as the comment in the file explains this is only a workaround. I wanted to point out to @asmorodskyi that this file is confusing in *exactly* the way you showed but asmorodskyi just stated to "read the comment". I gave up and approved. Now it's your turn to decide if you also just wanna give up or insist on changes ;)

asmorodskyi: please read comments in that file . it should give you answer . If you can not get answer to your question from comment above this means we need to improve comment, so I on purpose won't answer and let you find answer in comments :)

asmorodskyi: ok PR updated .is it more clear now who owns project ?

asmorodskyi: I have updated comments with bold capslock to make it even more visible :smile:

CODEOWNERS
```diff
@@ -0,0 +1 @@
+Project owned by members of https://github.com/orgs/openSUSE/teams/qa-tools-team
```
d3flex, line 1: is it just me seeing "This CODEOWNERS file contains errors". Once i open it reports a valid syntax but according to https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax this line is not valid

.github/CODEOWNERS
kalikiana: My question is still the same. Why shouldn't the whole team be listed and get notifications for PR's? And why is this needed if I'm already able to see those new pull requests?

asmorodskyi: > My question is still the same. Why shouldn't the whole team be listed and get notifications for PR's? 


I don't mind whole team to be least but I have concerns that whole team wants this :) 


> And why is this needed if I'm already able to see those new pull requests?

I don't get **emails** when new PR is created . do you have such emails ?

CODEOWNERS
asmorodskyi: I will fix this

.github/CODEOWNERS
baierjan: > I don't get emails when new PR is created . do you have such emails ?

Yes, I **do** get email for a new PR.

On a repository I am interested in:
20260416_21h35m22s_grim

In the account settings (Notifications section):
20260416_21h36m01s_grim

okurz: > I don't get **emails** when new PR is created . do you have such emails ?

I also get emails

okurz (changes requested):
I am revising my earlier approval because it now looks that @asmorodskyi simply wants to have it because they never configured emails according to https://github.com/openSUSE/qem-bot/pull/485#discussion_r3095993860

asmorodskyi:
ok in this case I will close this but I really need to figure out why it is not working for me ...

CODEOWNERS
asmorodskyi: sorry I won't fix this anymore :D

.github/CODEOWNERS
asmorodskyi: so you are saying that when I have this : 

image

I am NOT getting PR creation notification and when I have what you have than I do ? :thinking: 

can someone forward to my mailbox one random email showing me example of notification about PR creation ?

asmorodskyi: just for clarity , second screenshot is exactly the same for me and I haven't changed it

baierjan: I would guess that all activity should be good enough, if you also have the second part in the settings (customize email updates).

asmorodskyi: > I would guess that all activity should be good enough

but it is not enough because I am not getting this emails :thinking:

asmorodskyi: nevermind , I am giving up and will try to switch to GH Notifications instead and just disable mail notifications

PR #488, opened 2026-04-15
feat: add devel flag to increment-approve and remove devel job filtering

asmorodskyi:
- Add --devel CLI option and QEM_BOT_DEVEL environment variable to the increment-approve command.
- Introduce devel field to IncrementConfig and ApprovalStatus to track development mode.
- Remove automatic filtering of openQA jobs in development groups by deleting _filter_results and _filter_jobs from IncrementApprover.
- Update handle_approval to skip OBS request approval when the devel flag is enabled.
- Align the test suite by removing tests for deleted filtering logic and updating ApprovalStatus constructor calls.
- Add test_handle_approval_devel_skip to verify approval skipping in development mode.

### DIFF
--- a/openqabot/args.py
+++ b/openqabot/args.py
@@ -528,6 +528,10 @@ def increment_approve(  # noqa: PLR0913
             help="Monitor and schedule only products with the specified DISTRI parameter",
         ),
     ] = "sle",
+    devel: Annotated[
+        bool,
+        typer.Option("--devel", envvar="QEM_BOT_DEVEL", help="Include development openQA jobs"),
+    ] = False,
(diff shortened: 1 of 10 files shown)


okurz (changes requested):
Please describe the motivation. I only see what changes you did but not the problem you try to fix and how you want to solve it.

The past days working on KLP I also added a "--devel-filter/no-devel-filter" arg which is maybe what you had in mind. I didn't have a pull request for that yet. Created https://github.com/openSUSE/qem-bot/pull/489 now with also a potential fix for the devel job filtering not considered for scheduling evaluation.

PR #489, opened 2026-04-15
fix: prevent constant retriggering of development job groups

okurz:
Motivation:
When development job groups are filtered out, the increment approver
previously saw no jobs and would re-schedule them if the --schedule flag
was active. This led to unnecessary duplicate jobs.

Design Choices:
- Modified `process_build_info` to keep track of raw (unfiltered) results.
- Updated `_handle_not_ready_jobs` to distinguish between "no jobs at all"
  and "jobs exist but were filtered out".
- If jobs exist but are filtered, a message "No jobs for evaluation
  (filtered)" is added to the reasons to disapprove, and no new jobs are
  scheduled.
- Refactored `_handle_not_ready_jobs` to reduce argument count and comply
  with linting rules.

Benefits:
Avoids noisy and redundant job triggering in openQA when developers or
test maintainers move jobs to development groups.

### DIFF
--- a/openqabot/incrementapprover.py
+++ b/openqabot/incrementapprover.py
@@ -430,16 +430,18 @@ class IncrementApprover:
     ) -> int:
         """Handle cases where openQA jobs are not ready or missing."""
         info_str = build_info.format_multi_build(params)
-        if openqa_jobs_ready is None:
-            if jobs_were_filtered:
-                approval_status.reasons_to_disapprove.append(f"No jobs for evaluation (filtered) for {info_str}")
-                return 0
+        if openqa_jobs_ready is False:
+            approval_status.reasons_to_disapprove.append(f"Not all jobs ready for {info_str}")
+            return 0
 
-            approval_status.reasons_to_disapprove.append(f"No jobs scheduled for {info_str}")
-            return self.schedule_openqa_jobs(build_info, params) if self.args.schedule else 0
+        if jobs_were_filtered:
+            approval_status.reasons_to_disapprove.append(
+                f"No jobs left for evaluation (all were filtered) for {info_str}"
+            )
+            return 0
 
-        approval_status.reasons_to_disapprove.append(f"Not all jobs ready for {info_str}")
-        return 0
+        approval_status.reasons_to_disapprove.append(f"No jobs scheduled for {info_str}")
+        return self.schedule_openqa_jobs(build_info, params) if self.args.schedule else 0
 
     def process_build_info(
         self,
--- a/openqabot/incrementapprover.py
+++ b/openqabot/incrementapprover.py
@@ -423,17 +423,22 @@ class IncrementApprover:
         self,
         build_info: BuildInfo,
         params: ScheduleParams,
-        info_str: str,
         *,
         openqa_jobs_ready: bool | None,
         approval_status: ApprovalStatus,
+        jobs_were_filtered: bool = False,
     ) -> int:
         """Handle cases where openQA jobs are not ready or missing."""
+        info_str = build_info.format_multi_build(params)
         if openqa_jobs_ready is None:
-            approval_status.reasons_to_disapprove.append("No jobs scheduled for " + info_str)
+            if jobs_were_filtered:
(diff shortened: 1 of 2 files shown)


Wabri (dismissed):
Can you please fix the syntax errors? https://github.com/openSUSE/qem-bot/actions/runs/24443159455/job/71412838233?pr=489#step:3:12

okurz:
> Can you please fix the syntax errors? https://github.com/openSUSE/qem-bot/actions/runs/24443159455/job/71412838233?pr=489#step:3:12

fixed in #490 , rebased on top

okurz:
Let me look into better naming and refactoring to make this easier to understand

EDIT: DONE

openqabot/incrementapprover.py
```diff
@@ -104,7 +104,7 @@ def get_regex_match(pattern: str, string: str) -> re.Match | None:
 
     def _filter_jobs(self, jobs: dict[str, dict[str, Any]]) -> dict[str, dict[str, Any]]:
         """Filter jobs within a state, removing those in devel groups."""
-        if not config.settings.devel_filter:
+        if not config.settings.devel_filter or config.settings.allow_development_groups:
```
asmorodskyi: **question(blocking)** : can you elaborate on this ? how this setting correlate with `devel_filter` they looks very familiar

asmorodskyi (changes requested):

openqabot/incrementapprover.py
```diff
@@ -426,17 +426,22 @@ def _handle_not_ready_jobs(
...
         *,
         openqa_jobs_ready: bool | None,
         approval_status: ApprovalStatus,
+        jobs_were_filtered: bool = False,
     ) -> int:
         """Handle cases where openQA jobs are not ready or missing."""
+        info_str = build_info.format_multi_build(params)
         if openqa_jobs_ready is None:
-            approval_status.reasons_to_disapprove.append("No jobs scheduled for " + info_str)
+            if jobs_were_filtered:
```
asmorodskyi: **suggestion (non-blocking)** : why it is actually under `if openqa_jobs_ready is None` ? I think we can/should put it on same level with `openqa_jobs_ready`

openqabot/incrementapprover.py
okurz: actually as I misunderstood your PR I will leave the "devel-filter" part for later. I removed the first commit from this branch and this change. So this PR now solely focusses on the fix-part

openqabot/incrementapprover.py
```diff
@@ -426,17 +426,22 @@ def _handle_not_ready_jobs(
         self,
         build_info: BuildInfo,
         params: ScheduleParams,
-        info_str: str,
         *,
         openqa_jobs_ready: bool | None,
         approval_status: ApprovalStatus,
+        jobs_were_filtered: bool = False,
     ) -> int:
         """Handle cases where openQA jobs are not ready or missing."""
```
asmorodskyi: **suggestion(blocking)**:  let's improve documentation to ease further maintenance of this code :

```
"""Manage disapproval reasons and optional rescheduling for incomplete openQA results.

        Args:
            build_info (BuildInfo): Metadata regarding the product and build version.
            params (ScheduleParams): The parameters required to identify or schedule openQA jobs.
            openqa_jobs_ready (bool | None): Job state (None for missing, False for pending).
            approval_status (ApprovalStatus): The status container updated with disapproval reasons.
            jobs_were_filtered: (bool) : controls if some jobs were filtered which would mean that we should exclude them from approval status 
                                               

        Returns:
            int: The number of failed job postings if rescheduling is performed; otherwise 0.

        """
```

asmorodskyi: feel free to come up with different statements but this function need better doc than just to be happy that we passing ruff checks ...

openqabot/incrementapprover.py
okurz: done

asmorodskyi (approved):
LGTM

openqabot/incrementapprover.py
okurz: fine. added

Martchus (approved):

PR #490, opened 2026-04-15
fix: resolve ty type check errors revealed by new ty version

okurz:
Design Choices:
- Applied '# ty: ignore[invalid-assignment]' to 'osc.core.Request().state'
  assignments in 'tests/helpers.py' and also in
  'tests/test_incrementapprover_scenarios.py'. This avoids AttributeError
  when 'to_xml()' is called on real Request objects during debug logging,
  as our fake Action objects lack this method.
- Replaced 'bot.openqa = True' with a proper mock of 'OpenQAInterface'
  in 'tests/test_openqabot_simple.py' to satisfy strict type checking
  while maintaining boolean truthiness for conditional logic.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -151,7 +151,7 @@ def fake_get_request_list(url: str, project: str, **_kwargs: Any) -> list[osc.co
     assert "OBS:PROJECT" in project
     req = osc.core.Request()
     req.reqid = 42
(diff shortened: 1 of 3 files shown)


Wabri (approved):

asmorodskyi (approved):
LGTM

PR #491, opened 2026-04-15
ci: prevent duplicate call to ruff with inconsistent version

okurz:
Since be354b16 the very qick "checkstyle" including ruff is called
implicitly as part of the "make test-with-coverage" Makefile target. But
the "ruff-action" in GitHub actions executes ruff using the latest
upstream ruff version leading to inconsistent results.


asmorodskyi (approved):
LGTM

nicksinger (approved):

PR #494, opened 2026-04-16
fix: use env in shebang for virtual environment support

asmorodskyi:
Changes the shebang in qem-bot.py
from /usr/bin/python3 to /usr/bin/env python3.
This allows the script to use the Python interpreter
from an active virtual environment.

### DIFF
--- a/qem-bot.py
+++ b/qem-bot.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 # Copyright SUSE LLC
 # SPDX-License-Identifier: MIT
 """Main script to run the bot."""


okurz (changes requested):
please see #299 first. you may wanna update your commit message accordingly

asmorodskyi:
> please see #299 first. you may wanna update your commit message accordingly

maybe instead of throwing URL you will write one/two sentence what you trying to say with it ? I have read through all conversations where different people expressing different opinions about topic . If I get it correct only reason not to use env is that you can not kill running process because path replaced with something . Did I get it right ? Anything else what I have missed ?

Wabri:
> > please see #299 first. you may wanna update your commit message accordingly
> 
> maybe instead of throwing URL you will write one/two sentence what you trying to say with it ? I have read through all conversations where different people expressing different opinions about topic . If I get it correct only reason not to use env is that you can not kill running process because path replaced with something . Did I get it right ? Anything else what I have missed ?

Wrote by @perlpunk:
(replace perl with python) I can only say that for CPAN modules (or rather any scripts that are installed with them) they should have #!/usr/bin/perl as their shebang.

The reason is that they are expected to work on the version of perl they were installed with.

If you have some other version /some-debugging-perl in your path and are executing an installed perl script, for example at /usr/bin/mojo, you want to run it with /usr/bin/perl and not your temporary perl.
Or if you want, you can always call it explicitly with perl.

I think it's good to be able to rely on the fact that every script uses /usr/bin/perl, and if for some reason you need to run perl from a different path, do it explicitly with perl script.pl.

If you use /usr/bin/env everywhere, then adding a path to a different python will change it for every script that uses env, and for some that could be the wrong python.

okurz:
> > please see #299 first. you may wanna update your commit message accordingly
> 
> maybe instead of throwing URL you will write one/two sentence what you trying to say with it ? I have read through all conversations where different people expressing different opinions about topic . If I get it correct only reason not to use env is that you can not kill running process because path replaced with something . Did I get it right ? Anything else what I have missed ?

What you mention refers to https://github.com/os-autoinst/openqa_review/pull/153#discussion_r658592487 . Next to that there are the following arguments:
1. a security aspect: with "env" a user might be lured into calling a different interpreter
2. https://github.com/openSUSE/qem-bot/pull/299#discussion_r2538753717 states that packages are expected to work on the version of the interpreter they were installed with.
3. Please also consider the motivation of https://github.com/os-autoinst/openqa_review/commit/b1275f79f21214d34cff08b3ba9e0d3a0e1fa836 so that we are not going back and forth
4. Issue #299 was closed without the change being accepted. The PR doesn't acknowledge why this time should be different.

asmorodskyi:
> Wrote by @perlpunk: (replace perl with python) I can only say that for CPAN modules (or rather any scripts that are installed with them) they should have #!/usr/bin/perl as their shebang.
> 
> The reason is that they are expected to work on the version of perl they were installed with.
> 
> If you have some other version /some-debugging-perl in your path and are executing an installed perl script, for example at /usr/bin/mojo, you want to run it with /usr/bin/perl and not your temporary perl. Or if you want, you can always call it explicitly with perl.
> 
> I think it's good to be able to rely on the fact that every script uses /usr/bin/perl, and if for some reason you need to run perl from a different path, do it explicitly with perl script.pl.
> 
> If you use /usr/bin/env everywhere, then adding a path to a different python will change it for every script that uses env, and for some that could be the wrong python.


**(replace perl with python)**  - I would interrupt here . Comparing Python's virtual environments to Perl's CPAN is a bit like comparing a toolbox to a hardware store. While they are related in the workflow of managing code, they serve fundamentally different roles.

In Python terms, CPAN is the equivalent of PyPI (the Python Package Index). Python’s virtual environments (venv) are more equivalent to Perl tools like local::lib, perlbrew, or Carton.

perlpunk:
> In Python terms, CPAN is the equivalent of PyPI (the Python Package Index). Python’s virtual environments (venv) are more equivalent to Perl tools like local::lib, perlbrew, or Carton.

And the same applies there. In perlbrew you have the same problem. If you have a perl program installed in the system with `/usr/bin/perl`, let's say the `ack` script, you can always rely on it.
If it had `/usr/bin/env perl` as a shebang and you are in a perlbrew environment, and you all `ack`, then it will call your perlbrew perl, where you don't have the `App::Ack` library installed, so it will fail.
I was talking about CPAN, but it was also about rpm packages - we enforce `/usr/bin/perl` shebangs for packaged perl scripts for the reason above.

Of course things might be different with qem-bot.py, because it is actually not installed; but we use the git version directly.

PR #495, opened 2026-04-16
perf(increment-approve): optimize OBS source report processing + development group filtering

okurz:

### DIFF
--- a/openqabot/loader/sourcereport.py
+++ b/openqabot/loader/sourcereport.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 import tempfile
 from collections import defaultdict
+from concurrent.futures import ThreadPoolExecutor
 from functools import lru_cache
 from logging import getLogger
 from typing import Any
@@ -100,18 +101,30 @@ def compute_packages_of_request_from_source_report(
     request: osc.core.Request,
 ) -> tuple[defaultdict[str, set[Package]], int]:
     """Compute the package diff of a request based on source reports."""
+    # target projects (e.g. `SUSE:Products:SLE-Product-SLES:16.0:aarch64`)
     repo_a: defaultdict[str, set[Package]] = defaultdict(set)
+    # source projects (e.g. `SUSE:SLFO:Products:SLES:16.0:TEST`)
     repo_b: defaultdict[str, set[Package]] = defaultdict(set)
 
-    for action in request.actions:
-        log.debug("Checking action '%s' -> '%s' of request %s", action.src_project, action.tgt_project, request.reqid)
-        # add packages for target project (e.g. `SUSE:Products:SLE-Product-SLES:16.0:aarch64`), that is repo "A"
-        load_packages_from_source_report(
-            action, OBSBinary(action.tgt_project, action.src_package, "images", "local"), repo_a
-        )
-        # add packages for source project (e.g. `SUSE:SLFO:Products:SLES:16.0:TEST`), that is repo "B"
-        load_packages_from_source_report(
-            action, OBSBinary(action.src_project, action.src_package, "product", "local"), repo_b
-        )
+    def worker(action: Any, binary: OBSBinary) -> defaultdict[str, set[Package]]:  # noqa: ANN401
+        packages: defaultdict[str, set[Package]] = defaultdict(set)
+        load_packages_from_source_report(action, binary, packages)
+        return packages
+
+    tasks = [
+        *((a, OBSBinary(a.tgt_project, a.src_package, "images", "local")) for a in request.actions),
+        *((a, OBSBinary(a.src_project, a.src_package, "product", "local")) for a in request.actions),
+    ]
+
+    with ThreadPoolExecutor() as executor:
+        # Use executor.map to ensure exceptions are raised if they occur in threads
+        results = list(executor.map(lambda p: worker(*p), tasks))
+
+    # results contains first all "target" projects, then all "source" projects
+    num_actions = len(request.actions)
+    for i, res in enumerate(results):
+        target_repo = repo_a if i < num_actions else repo_b
+        for arch, pks in res.items():
+            target_repo[arch].update(pks)
 
     return RepoDiff.compute_diff_for_packages("product repo", repo_a, "TEST repo", repo_b)
--- a/openqabot/loader/sourcereport.py
+++ b/openqabot/loader/sourcereport.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 import tempfile
 from collections import defaultdict
+from functools import lru_cache
 from logging import getLogger
 from typing import Any
 from urllib.error import HTTPError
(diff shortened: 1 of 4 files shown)


asmorodskyi (dismissed):
I just would like to raise a flag that multi-thread is quite expensive thing which brings it's own set of problems to a code . And we should always think hard to find other ways how to achieve same goal . From I understand only reason why we switching to multi-threading is because @okurz  find it's annoying to wait 10 minutes for single run **during development phase** which I find quite questionable reasoning .  multi-threading for example makes debug'ing of this code challenging .

My blocking does not mean hard block but I would like raise discussion and gather more opinions before we actually dive into this river ...

okurz:
> I just would like to raise a flag that multi-thread is quite expensive thing which brings it's own set of problems to a code

multi-threading is not expensive. Here the code is naturally mapping out what we want to do: Process *all* datapoints, regardless of the order or any other dependency.

By the way, the two commits mostly optimized by caching and a more efficient lookup. The multi-threading is only secondary and is following the concept of futures that are already used in qem-bot in other locations

openqabot/loader/sourcereport.py
```diff
@@ -95,14 +103,26 @@ def compute_packages_of_request_from_source_report(
     """Compute the package diff of a request based on source reports."""
     repo_a: defaultdict[str, set[Package]] = defaultdict(set)
     repo_b: defaultdict[str, set[Package]] = defaultdict(set)
-    for action in request.actions:
-        log.debug("Checking action '%s' -> '%s' of request %s", action.src_project, action.tgt_project, request.id)
-        # add packages for target project (e.g. `SUSE:Products:SLE-Product-SLES:16.0:aarch64`), that is repo "A"
```
Martchus, line 100: I'd like to preserve these comments which state what A and B mean in this specific context. The example is SUSE-specific but nevertheless also useful in practice to know what kind of repos are expected to be compared here.

Martchus (changes requested):

asmorodskyi:
> multi-threading is not expensive. Here the code is naturally mapping out what we want to do: Process _all_ datapoints, regardless of the order or any other dependency.


disagree .  By introducing multithreading (MT), we trade a stable, linear execution model for non-deterministic failure modes (like race conditions or hung threads) that are significantly harder to debug in CI/CD. Even local debug'ing is becoming big problem with multi-threading. Also "classical" MT problems like dead-locks and race conditions does not look like related to our current code but still should be taken into consideration as potential risk 

> By the way, the two commits mostly optimized by caching and a more efficient lookup. 

and have you checked performance only with this commits ? w/o introduction of MT ? maybe it is enough ?

> The multi-threading is only secondary and is following the concept of futures that are already used in qem-bot in other locations

and it is in my TODO list to try to remove it from there too :) so does not sound like right argument

okurz:
> > multi-threading is not expensive. Here the code is naturally mapping out what we want to do: Process _all_ datapoints, regardless of the order or any other dependency.
> 
> disagree . By introducing multithreading (MT), we trade a stable, linear execution model for non-deterministic failure modes (like race conditions or hung threads) that are significantly harder to debug in CI/CD. Even local debug'ing is becoming big problem with multi-threading. Also "classical" MT problems like dead-locks and race conditions does not look like related to our current code but still should be taken into consideration as potential risk
> 
> > By the way, the two commits mostly optimized by caching and a more efficient lookup.
> 
> and have you checked performance only with this commits ? w/o introduction of MT ? maybe it is enough ?

Yes, I have now with multiple measurements (which took me 1.5h just to convince you). Each row shows the result of the commit and all previous ones included

| Commit | Description | Mean Runtime (ms) | Speedup | Improvement | CV (Stability) |
| :--- | :--- | :--- | :--- | :--- | :--- |
| 9e669b7a | origin/master | 164878 ± 35065.24 | 1.000x | 0.00% | 21.27% |
| 2e51c50e | optimize development group filtering | 96347 ± 3441.57 | 1.711x | 41.56% | 3.57% |
| 1b4ac53d | optimize OBS source report processing | 108278 ± 9709.53 | 1.523x | 34.33% | 8.97% |
| b1bbdbc6 | multi-thread dev group filtering | 100747 ± 6975.89 | 1.637x | 38.90% | 6.92% |
| 191830d0 | multi-thread OBS source report processing | 52488 ± 2986.74 | 3.141x | 68.17% | 5.69% |

here it looks like "optimize OBS source report processing" actually caused a slow-down vs. it's base "optimize development group filtering" which I assume is an outlier. But the following multi-thread commit show again a visible improvement. It should be obvious that the biggest effect comes from "multi-thread OBS source report processing".

The benchmarks should answer your question directly. With only the optimization/caching commits (no MT), runtime is ~96s. With MT for OBS source report processing, it drops to ~52s — a further 46% reduction that caching cannot achieve because the bottleneck shifts to I/O-bound parallel HTTP fetches.

Regarding MT risks: Same as in existing code I use concurrent.futures.ThreadPoolExecutor with no shared mutable state: Each work unit is independent. There are no locks, no shared data structures to mutate, so deadlocks and race conditions are structurally impossible. This is the same pattern already used in qem-bot (and the standard Python pattern for I/O-bound parallelism).
>
> The "optimize OBS source report processing" commit showing ~108s (vs 96s for the previous) is within measurement noise given its 9% CV — the important signal is the final result.
> > The multi-threading is only secondary and is following the concept of futures that are already used in qem-bot in other locations
> 
> and it is in my TODO list to try to remove it from there too :) so does not sound like right argument

This is ridiculous.  If you plan to remove existing futures usage, that should be a separate discussion - removing I/O parallelism from an I/O-bound tool would be a regression.

openqabot/loader/sourcereport.py
okurz: done

nicksinger: Could you add it to 4288049a466389a142d402b9f2da1cecd0c02121 (where the actual change was done) and not 26a6c1889f73b12288d615647dc0df8394e4dd0b please?

okurz: done. Still did that where repo_a and repo_b are defined. I hope that's what you had in mind

Martchus (approved):

openqabot/loader/sourcereport.py
```diff
@@ -93,16 +101,30 @@ def compute_packages_of_request_from_source_report(
...
+        return packages
+
+    tasks = [
+        *((a, OBSBinary(a.tgt_project, a.src_package, "images", "local")) for a in request.actions),
+        *((a, OBSBinary(a.src_project, a.src_package, "product", "local")) for a in request.actions),
+    ]
+
+    with ThreadPoolExecutor() as executor:
+        # Use executor.map to ensure exceptions are raised if they occur in threads
+        results = list(executor.map(lambda p: worker(*p), tasks))
```
Martchus, line 121: By the way: Here `load_packages_from_source_report` doesn't modify shared/global state (as far as I can tell, I haven't checked the code of the osc module and possibly other dependencies) and we actually use the result of the `map` call instead of modifying a shared list. So I guess this piece of code is good to run in parallel. I'm just mentioning this because you should probably do something similar in the other PR as well.

kalikiana (approved):

PR #496, opened 2026-04-16
feat: use REGEX query parameter for server-side filtering

okurz:
Motivation:
To provide more powerful and precise server-side filtering of directory
listings on IBS/OBS, switch from the globbing 'P' parameter to the 'REGEX'
parameter. This allows using the fully configured build regex for filtering
at the source, reducing data transfer and noise.

Design Choices:
- Updated get_obs_filter_params to use 'REGEX' key instead of 'P'.
- Changed build_filter default to None to allow leveraging build_regex as the
  default server-side filter.
- Updated load_build_info to use build_regex if no explicit build_filter is
  provided.
- Updated repodiff.py to use a proper regex for primary XML discovery.
- Updated test helpers to reflect the new URL format and parameters.

Benefits:
- More efficient artifact discovery by filtering out irrelevant files
  (e.g. -Debug, -Source) directly on the server.
- Better consistency by leveraging existing regex configurations for
  server-side filtering.

### DIFF
--- a/openqabot/config.py
+++ b/openqabot/config.py
@@ -45,7 +45,6 @@ class Settings(BaseSettings):
     singlearch: Path = Field(default=Path("/etc/openqabot/singlearch.yml"), alias="QEM_BOT_SINGLEARCH")
     retry: int = Field(default=2, alias="QEM_BOT_RETRY")
     approve_comment: bool = Field(default=False, alias="QEM_BOT_APPROVE_COMMENT")
-    build_filter: str | None = Field(default=None, alias="QEM_BOT_BUILD_FILTER")
 
     # App-specific settings
     qem_dashboard_url: str = Field(default="http://dashboard.qam.suse.de/", alias="QEM_DASHBOARD_URL")
--- a/openqabot/config.py
+++ b/openqabot/config.py
@@ -45,7 +45,7 @@ class Settings(BaseSettings):
     singlearch: Path = Field(default=Path("/etc/openqabot/singlearch.yml"), alias="QEM_BOT_SINGLEARCH")
     retry: int = Field(default=2, alias="QEM_BOT_RETRY")
     approve_comment: bool = Field(default=False, alias="QEM_BOT_APPROVE_COMMENT")
-    build_filter: str = Field(default="*spdx.json", alias="QEM_BOT_BUILD_FILTER")
+    build_filter: str | None = Field(default=None, alias="QEM_BOT_BUILD_FILTER")
 
     # App-specific settings
     qem_dashboard_url: str = Field(default="http://dashboard.qam.suse.de/", alias="QEM_DASHBOARD_URL")
--- a/openqabot/loader/buildinfo.py
+++ b/openqabot/loader/buildinfo.py
@@ -32,7 +32,7 @@ def load_build_info(
     build_project_url = config.build_project_url()
     sub_path = config.build_listing_sub_path
     url = f"{build_project_url}/{sub_path}/"
-    params = get_obs_filter_params(config.build_filter)
+    params = get_obs_filter_params(config.build_filter or build_regex)
     log.debug("Checking for '%s' files on %s with params %s", build_regex, url, params)
     rows = retried_requests.get(url, params=params).json().get("data", [])
 
--- a/openqabot/loader/buildinfo.py
+++ b/openqabot/loader/buildinfo.py
@@ -32,7 +32,7 @@ def load_build_info(
     build_project_url = config.build_project_url()
     sub_path = config.build_listing_sub_path
     url = f"{build_project_url}/{sub_path}/"
-    params = get_obs_filter_params(config.build_filter or build_regex)
+    params = get_obs_filter_params(build_regex)
     log.debug("Checking for '%s' files on %s with params %s", build_regex, url, params)
     rows = retried_requests.get(url, params=params).json().get("data", [])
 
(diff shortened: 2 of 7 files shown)


openqabot/loader/buildinfo.py
```diff
@@ -32,7 +32,7 @@ def load_build_info(
     build_project_url = config.build_project_url()
     sub_path = config.build_listing_sub_path
     url = f"{build_project_url}/{sub_path}/"
-    params = get_obs_filter_params(config.build_filter)
+    params = get_obs_filter_params(config.build_filter or build_regex)
```
asmorodskyi: **suggestion(blocking)** : I find this confusing. keep in mind that in a fact build_regex and build_filter both fields of **same class** . what if we would just move `get_obs_filter_params` into IncrementConfig class and let this class to decide what should be filter param ?

asmorodskyi (changes requested):

openqabot/utils.py
```diff
@@ -109,15 +109,15 @@ def merge_dicts(dict1: dict[Any, Any], dict2: dict[Any, Any]) -> dict[Any, Any]:
 
 
 def get_obs_filter_params(pattern: str) -> dict[str, Any]:
-    """Reduce data transfer by evaluating the 'P' parameter at the source via 'jsontable'.
+    """Reduce data transfer by evaluating the 'REGEX' parameter at the source via 'jsontable'.
 
     References:
         * https://github.com/openSUSE/MirrorCache/blob/207d61237c0597f8f4ff9d7ad12c4f9cb5d5cd1f/lib/MirrorCache/Datamodule.pm#L311
```
asmorodskyi:
```suggestion
        * https://github.com/openSUSE/MirrorCache/blob/207d61237c0597f8f4ff9d7ad12c4f9cb5d5cd1f/lib/MirrorCache/Datamodule.pm#L309
```

openqabot/utils.py
```diff
@@ -109,15 +109,15 @@ def merge_dicts(dict1: dict[Any, Any], dict2: dict[Any, Any]) -> dict[Any, Any]:
...
         * https://github.com/openSUSE/MirrorCache/issues/349
```
asmorodskyi, line 116: I followed URL and can not understand how it is related here 🤔

openqabot/config.py
```diff
@@ -45,7 +45,7 @@ class Settings(BaseSettings):
     singlearch: Path = Field(default=Path("/etc/openqabot/singlearch.yml"), alias="QEM_BOT_SINGLEARCH")
     retry: int = Field(default=2, alias="QEM_BOT_RETRY")
     approve_comment: bool = Field(default=False, alias="QEM_BOT_APPROVE_COMMENT")
-    build_filter: str = Field(default="*spdx.json", alias="QEM_BOT_BUILD_FILTER")
+    build_filter: str | None = Field(default=None, alias="QEM_BOT_BUILD_FILTER")
```
asmorodskyi: **question(non-blocking)** - why do you think we should not have default value anymore ?
**suggestion(non-blocking)** - I think would make more sense if this parameter would store not just value but key/value ( e.g.  `REGEX=<regex>` ) 
**suggestion(non-blocking)** - I think what we store here has nothing to do with "build_filter" it sounds more like "filter_param" .

openqabot/utils.py
okurz: It shows how the `?P=` parameter can be used

openqabot/config.py
okurz: Because by using the regex that is already defined we don't need to replicate this value here. That is what is stated in the commit message with
> Changed build_filter default to None to allow leveraging build_regex as the default server-side filter.

asmorodskyi: so we have configuration parameter which meant to  be used for server-side filtering . you can configure it via cmd parameter or as environment variable . But if you not define it anywhere deep down in code we would just use some other unrelated variable to define it . I find this flow very confusing ...

okurz: > some other unrelated variable to define it

no, it's not unrelated. That is actually what we want to filter for. I optimized the lookup by using the regex we would use for filtering already earlier already in the remote request. So instead of:

```
1. request the server to list all files
2. transfer the list of all files from the remote server to local
3. filter by regex locally
```

we do

```
1. request to server to list files filtered by regex
2. transfer the list of filtered files from the remote server to local
3. process the regex filtered list locally
```

which has two benefits:
1. Less data to transfer hence performance optimization
2. Less noisy log output of "all files" from which we later (silently) filter anyway. This eases development and was actually my motivation why I started this change

asmorodskyi: 90% of your comment is about why it is good to do server side filtering but I never question that. Yes it is better to do filtering server side . topic closed :) 

My concerns about how you store and use **the value** for filtering . Currently codewise we have identified  `build_filter` to be variable which store value for the filter which ( for me at least ) mean that when this variable has value None filter is not defined , but in reality instead in such a case we using different variable which might be related but holds different purpose. 

Maybe we should just drop `build_filter`   and use `build_regex` to do server-side filtering ?

okurz: > Maybe we should just drop build_filter and use build_regex to do server-side filtering ?

yes, thought about that as well. I did that now.

openqabot/loader/buildinfo.py
okurz: I dropped "build_filter" completely now, see https://github.com/openSUSE/qem-bot/pull/496#discussion_r3102961346

openqabot/utils.py
okurz: done

asmorodskyi (approved):
LGTM

Martchus (approved):

PR #498, opened 2026-04-16
fix: implement cooldown for submission rescheduling

okurz:
## Summary
- Introduced a configurable `schedule_cooldown` setting (defaulting to 2 hours) to stabilize the openQA job queue.
- Updated `is_scheduled_job` logic to prevent redundant rescheduling when `REPOHASH` changes within the cooldown window.
- Refactored logic to be more concise and improved log readability.

Related issue: https://progress.opensuse.org/issues/199793


okurz:
not following up with this non-important branch

PR #499, opened 2026-04-17
refactor: robust QEM Dashboard URL construction

okurz:
Motivation:
Missing a trailing slash in the QEM_DASHBOARD_URL environment variable
previously led to invalid URL construction through simple string
concatenation (e.g., 'http://hostapi/...' instead of 'http://host/api/...').

Design Choices:
- Added a Pydantic field validator in the Settings class to ensure that
  qem_dashboard_url always ends with a trailing slash upon initialization.
- Implemented a centralized 'dashboard_url' helper method in the Settings
  class that uses 'urllib.parse.urljoin' for robust path joining.
- Refactored core modules (dashboard.py, approver.py, etc.) to use this
  centralized helper instead of manual f-string concatenation.
- Updated relevant test suites to use the new helper, ensuring
  consistency between implementation and verification logic.
- Cleaned up redundant imports of 'urljoin' in specialized modules by
  encapsulating it within the configuration layer.

Benefits:
- Eliminates a common source of configuration-driven bugs.
- Improves code maintainability through better encapsulation of URL
  building logic.
- Ensures all parts of the application handle the dashboard base URL
  consistently, even when overridden by users.

### DIFF
--- a/openqabot/approver.py
+++ b/openqabot/approver.py
@@ -415,7 +415,7 @@ class Approver:
 
     def approve(self, sub: SubReq) -> bool:
         """Approve a submission in OBS or Gitea."""
-        msg = f"Request accepted for '{config.settings.obs_group}' based on data in {config.settings.qem_dashboard_url}"
+        msg = f"Request accepted for '{config.settings.obs_group}' based on data in {config.settings.dashboard_url()}"
         log.info("Approving %s", ms2str(sub))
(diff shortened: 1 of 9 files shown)


asmorodskyi (approved):
LGTM

kalikiana (approved):

Martchus (approved):

PR #500, opened 2026-04-17
feat: introduce unified sync command and grouped syncs

okurz:
Motivation:
Each command execution in production has a high container startup overhead.
Combining the most frequent sync operations into a single command reduces this
overhead significantly.

Design Choices:
- Introduced a new "sync" command that runs both SMELT and Gitea syncs.
- Moved individual "smelt-sync" and "gitea-sync" commands into the "advanced"
  sub-command group.
- Added hidden, deprecated top-level aliases for "smelt-sync" and "gitea-sync"
  to maintain backward compatibility.
- Unified sync accepts all Gitea-related options.

Benefits:
- Reduced production runtime by approximately 30 seconds per execution.
- Improved CLI organization while maintaining compatibility with existing
  automation pipelines.

### DIFF
--- a/openqabot/args.py
+++ b/openqabot/args.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 import logging
 import sys
+from dataclasses import asdict, dataclass
 from pathlib import Path
 from types import SimpleNamespace
 from typing import Annotated, Any
@@ -37,7 +38,16 @@ app = typer.Typer(
     no_args_is_help=True,
     add_completion=False,
 )
+advanced_app = typer.Typer(
+    name="advanced",
+    help="Advanced commands, e.g. for debugging.",
+    no_args_is_help=True,
+    add_completion=False,
+)
+app.add_typer(advanced_app)
 log = logging.getLogger("bot")
+
+DEFAULT_GITEA_PROJECT = "products/SLFO"
 
 pr_number_arg = Annotated[
     int | None,
@@ -52,6 +62,32 @@ gitea_project_arg = Annotated[
         "--gitea-project",
         envvar="GITEA_PROJECT",
         help="Project in Gitea to check for PRs. Project defined by {owner}/{repo}",
+    ),
+]
+allow_build_failures_option = Annotated[
+    bool,
+    typer.Option("--allow-build-failures", help="Sync data from PRs despite failing packages"),
+]
+consider_unrequested_prs_option = Annotated[
+    bool,
+    typer.Option(
+        "--consider-unrequested-prs",
+        help=f"Consider PRs where no review from team {config_module.settings.obs_group} was requested as well",
+    ),
+]
+amqp_option = Annotated[
+    bool,
+    typer.Option(
+        "--amqp",
+        help="After initial sync listen for new PRs via AMQP and submit them to QEM dashboard immediately",
+    ),
+]
+amqp_url_option = Annotated[str | None, typer.Option("--amqp-url", help="the URL of the AMQP server")]
+skip_initial_sync_option = Annotated[
+    bool,
+    typer.Option(
+        "--amqp-only",
+        help="Skip initial sync before handling AMQP events for new PRs",
     ),
 ]
 
@@ -158,6 +194,35 @@ def _require_gitea_token(args: SimpleNamespace) -> None:
             "Error: Missing option '--gitea-token' / '-g' or environment variable QEM_BOT_GITEA_TOKEN.", err=True
         )
         raise typer.Exit(1)
+
+
+def _require_sync_tokens(args: SimpleNamespace) -> None:
+    """Enforce both tokens needed by any command syncing Gitea into the dashboard."""
+    _require_token(args)
+    _require_gitea_token(args)
+
+
+@dataclass(frozen=True)
+class GiteaSyncOptions:
+    """Options of a Gitea sync run, shared by the "sync" and "advanced gitea-sync" commands."""
+
+    gitea_project: str = DEFAULT_GITEA_PROJECT
+    allow_build_failures: bool = False
+    consider_unrequested_prs: bool = False
+    pr_number: int | None = None
+    amqp: bool = False
+    amqp_url: str | None = None
+    skip_initial_sync: bool = False
+
+
+def _run_gitea_sync(args: SimpleNamespace, options: GiteaSyncOptions) -> int:
+    """Apply Gitea sync options onto the shared context and run the sync."""
+    for key, value in asdict(options).items():
+        setattr(args, key, value)
+    # Default from settings (which was already loaded in main callback)
+    if options.amqp_url is None:
+        args.amqp_url = config_module.settings.amqp_url
+    return GiteaSync(args)()
 
 
 @app.callback()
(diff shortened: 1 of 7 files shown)


openqabot/args.py
```diff
@@ -329,7 +336,68 @@ def smelt_sync(ctx: typer.Context) -> None:
...
+
+    # Gitea sync
+    args.gitea_repo = gitea_repo
+    args.allow_build_failures = allow_build_failures
+    args.consider_unrequested_prs = consider_unrequested_prs
+    args.pr_number = pr_number
+    args.amqp = amqp
+    # Default from settings (which was already loaded in main callback)
+    args.amqp_url = amqp_url if amqp_url is not None else config_module.settings.amqp_url
+    args.skip_initial_sync = skip_initial_sync
```
Martchus: If these assignments are C&P-ed from the individual command we should probably avoid the duplication. Otherwise we might introduce inconsistencies over time.

openqabot/args.py
```diff
@@ -37,6 +37,13 @@
     no_args_is_help=True,
     add_completion=False,
 )
+advanced_app = typer.Typer(
```
asmorodskyi, line 40: I don't understand what does it mean

okurz: this is the typer "app" instance that we extend the generic instance with to cover "more advanced and low-level commands"

asmorodskyi: and what are **more advanced and low-level commands** means ?

openqabot/args.py
```diff
@@ -336,7 +351,68 @@ def smelt_sync(ctx: typer.Context) -> None:
     sys.exit(syncer())
 
 
-@app.command("gitea-sync")
+@app.command("smelt-sync", hidden=True)
```
asmorodskyi, line 412: **question(blocking)**: I don't understand why do we need to over complicated this so much . the flow which you using here when command is removed in several iteration make sense for applications with huge user base and complex dependencies . Here we have script which has ONE production environment and ONE team using it . When we decide to delete something why can't we just delete it ?

asmorodskyi (changes requested):

okurz:
I will first try with multi-command executions #632

openqabot/args.py
okurz: Agreed, removed. Both hidden aliases are gone; `smelt-sync` and `gitea-sync` now exist only under `advanced`.

To make the reasoning explicit: the shims never protected anything. Production has to switch to `sync` anyway — reducing the per-invocation container startup is the entire point of the PR — so there was no window where the old top-level names were still needed. They only existed because the commands moved into `advanced` in the same change, which is self-inflicted complexity.

They also cost more than the line count suggests: the 8-option Gitea signature was declared three times, and it had already drifted twice within this branch (see the reply to Martchus below).

openqabot/args.py
okurz: You were right, and the duplication had already produced two bugs in this branch:

1. `sync` set `--gitea-repo` while `advanced gitea-sync` set `--gitea-project` — different flag names bound to the same `GITEA_PROJECT` env var.
2. `sync` only called `_require_token`, missing `_require_gitea_token`. It would run a full SMELT sync and then fail deep inside `GiteaSync` instead of failing fast.

Deduplicated in three steps, each removing one drift vector:

* Option definitions are now module-level `Annotated` aliases (matching the existing `pr_number_arg` / `comment_option` style), so flag name, env var, help text and type exist once. `gitea_repo_arg` is deleted.
* A frozen `GiteaSyncOptions` dataclass plus `_run_gitea_sync()` holds the body; the per-field assignments are gone (`asdict` loop), so adding an option means touching one dataclass.
* `_require_sync_tokens()` holds both guards, so bug 2 cannot recur. It stays out of `_run_gitea_sync` on purpose: `sync` must reject a missing Gitea token *before* spending a SMELT sync.

`sync` also no longer exposes `--pr-number` / `--amqp*`. Those are debugging-only (`--pr-number`'s own help text says so) and belong on `advanced gitea-sync`, not the production entry point — which shrinks `sync` from 7 options to 3.

Considered and rejected: hoisting the options onto `app.callback()`. That would be literally zero duplication, but `--gitea-project` and `--pr-number` would then be accepted by `sub-approve`, `amqp` and everything else, and would have to precede the subcommand. Not worth it to remove a list of parameter names. What remains duplicated is the parameter list in the two command signatures, which Typer forces.

openqabot/args.py
okurz: Fair — the wording was the problem. `advanced` means commands a human operator runs by hand for debugging, never invoked by the regular CI schedule: `smelt-sync`, `gitea-sync`, `sub-comment`, `repo-diff`. The scheduled pipeline only uses the top-level commands.

I put that sentence into the group help itself instead of leaving it as "advanced and low-level", so the distinction is visible in `--help` rather than needing this thread.

openqabot/args.py
okurz: Right. Typer requires parameters in the function signature for CLI introspection. To ensure consistency, the parameters are now defined as reusable Annotated types (e.g., allow_build_failures_option) and variable assignments are centralized via the GiteaSyncOptions dataclass. We can also consider to remove some potentially redundant individual commands after a transition period

openqabot/args.py
okurz: The help string and Readme.md have been updated to explicitly state that those are commands invoked manually by operators for debugging, not as part of our scheduled periodic jobs

openqabot/args.py
okurz: Fine. A deprecation cycle might be unnecessary for our single production environment. The deprecated aliases have been removed entirely. that still means we need to coordinate the release as "adapt the CI schedule quickly after merge"

okurz:
Updated with a backward incompatible change as suggested by @asmorodskyi which also means less duplication as pointed out by @Martchus . With this we move the existing sync commands to "advanced" and make the default "sync" cover both SMELT+gitea.

Martchus (approved):

openqabot/args.py
```diff
@@ -37,8 +38,17 @@
...
+advanced_app = typer.Typer(
+    name="advanced",
+    help="Advanced commands, e.g. for debugging.",
+    no_args_is_help=True,
+    add_completion=False,
+)
+app.add_typer(advanced_app)
 log = logging.getLogger("bot")
 
+DEFAULT_GITEA_PROJECT = "products/SLFO"
```
asmorodskyi: we already have `config_module.gitea_project` for that

openqabot/args.py
```diff
@@ -160,6 +196,35 @@ def _require_gitea_token(args: SimpleNamespace) -> None:
...
+@dataclass(frozen=True)
+class GiteaSyncOptions:
+    """Options of a Gitea sync run, shared by the "sync" and "advanced gitea-sync" commands."""
+
+    gitea_project: str = DEFAULT_GITEA_PROJECT
+    allow_build_failures: bool = False
+    consider_unrequested_prs: bool = False
+    pr_number: int | None = None
+    amqp: bool = False
+    amqp_url: str | None = None
```
asmorodskyi: should take default from `config_module.amqp_url` )

openqabot/args.py
okurz: done

openqabot/args.py
okurz: done

asmorodskyi (approved):

PR #501, opened 2026-04-20
refactor: centralize Gitea PR validation and simplify loader logic

asmorodskyi:
Integrated Gitea PR validation directly into the PullRequest dataclass
    via a new factory method, allowing for cleaner error handling across the
    loader and trigger modules.
    Key changes:
      - Refactored `PullRequest` to handle its own JSON parsing and validation,
         including robust error logging for missing or malformed API data.
      - Simplified `openqabot/loader/gitea.py` by extracting pagination and
          submission finalization logic into focused helper functions.
      - Updated `GiteaSync` and `GiteaTrigger` to leverage the refactored
          `PullRequest` class, reducing redundant try-except blocks.

### DIFF
--- a/openqabot/loader/gitea.py
+++ b/openqabot/loader/gitea.py
@@ -26,12 +26,12 @@ from osc.core import MultibuildFlavorResolver
 
 from openqabot import config
 from openqabot.errors import NoRepoFoundError
+from openqabot.types.pullrequest import PullRequest
 from openqabot.utils import retry10 as retried_requests
 
 if TYPE_CHECKING:
     from collections.abc import Iterator
 
-    from openqabot.types.pullrequest import PullRequest
     from openqabot.types.types import Repos
 
 ARCHS = {"x86_64", "aarch64", "ppc64le", "s390x"}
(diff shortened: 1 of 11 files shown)


okurz (changes requested):
seems like this is based on a rather old state. need to rebase

asmorodskyi:
> I removed some logic introduced recently not intentionally . fixing this now . don't review for now 

fixed

tests/test_loader_gitea_submissions.py
```diff
@@ -10,10 +10,18 @@
...
 
 
 def test_make_submission_from_gitea_pr_dry(mocker: MockerFixture) -> None:
-    pr = {"number": 124, "state": "open", "url": "url", "base": {"repo": {"full_name": "owner/repo", "name": "repo"}}}
+    pr_dict = {
+        "number": 124,
+        "state": "open",
+        "url": "url",
+        "base": {"repo": {"full_name": "owner/repo", "name": "repo"}},
+    }
```
okurz, line 22: the similar dict structure is repeated a lot. In before at least it did fit on single lines. I suggest to extract a base dict and feed a customized version into each test call.

openqabot/loader/gitea.py
```diff
@@ -195,26 +195,26 @@ def compute_repo_url_for_job_setting(
...
 
-    try:
-        # https://docs.gitea.com/api/1.25/#tag/repository/operation/repolistPullRequests
-        return list(iter_gitea_items(f"repos/{repo}/pulls?state=open", token))
-    except (requests.exceptions.RequestException, TypeError):
-        log.exception("Gitea API error: Could not fetch open PRs from %s", repo)
-        return []
+
+def _get_single_pr(token: dict[str, str], repo: str, number: int) -> list[PullRequest]:
+    if pr := PullRequest.create_from_json(cast("dict[str, Any]", get_json(f"repos/{repo}/pulls/{number}", token))):
```
okurz: network errors now propagate uncaught instead of being gracefully handled. Why this change?

openqabot/types/pullrequest.py
```diff
@@ -49,3 +56,28 @@ def has_all_labels(self, required_labels: set[str]) -> bool:
...
+
+    def is_active(self) -> bool:
+        """Check if the pull request is in an active state."""
+        return self.state == "open"
+
+    @classmethod
+    def create_from_json(cls: type[Self], pr_json: dict[str, Any]) -> Self | None:
+        """Return instance of PullRequest created from dict."""
+        try:
+            instance = cls(
```
okurz, line 65: AFAICS only KeyError is possible now so why catch requests.exceptions.RequestException, json.JSONDecodeError? Also, the syntax `instance = cls(…` isn't very clear.

openqabot/types/pullrequest.py
```diff
@@ -49,3 +56,28 @@ def has_all_labels(self, required_labels: set[str]) -> bool:
     def has_any_label(self, search_labels: set[str]) -> bool:
         """Check if pull request has at least one label from the input set."""
         return not self.labels.isdisjoint(search_labels)
+
+    def is_active(self) -> bool:
+        """Check if the pull request is in an active state."""
+        return self.state == "open"
+
+    @classmethod
+    def create_from_json(cls: type[Self], pr_json: dict[str, Any]) -> Self | None:
```
okurz: I think the common naming convention would be


```suggestion
    def from_json(cls: type[Self], pr_json: dict[str, Any]) -> Self | None:
```

okurz (changes requested):
Looks generally fine. Not sure if it makes the code cleaner but it can be acceptable. Just some minor issues about error handling, naming and refactoring tests

openqabot/types/pullrequest.py
asmorodskyi: you are right about exceptions I will remove them here and re-add where they were before . regarding syntax I am not sure what you mean . is it exact variable name confusing you ?

openqabot/types/pullrequest.py
asmorodskyi: done

openqabot/loader/gitea.py
asmorodskyi: I reverted this

openqabot/types/pullrequest.py
okurz: > is it exact variable name confusing you ?

no, not confusing. I guess this is actually Pythonic so, ok

okurz:
CI fails due to decrease of maintainability score

```
openqabot/loader/gitea.py - B
make[2]: *** [Makefile:57: check-maintainability] Error 1
```

asmorodskyi:
> CI fails due to decrease of maintainability score
> 
> ```
> openqabot/loader/gitea.py - B
> make[2]: *** [Makefile:57: check-maintainability] Error 1
> ```

thanks for duplicating info which I can get from CI .

asmorodskyi:
closing in favor of #504

PR #502, opened 2026-04-21
fix: approve increments when blockers only remain in dev  groups

okurz:
Motivation:
Product increment requests were incorrectly blocked when all scheduled openQA
jobs belonged to development groups. While these jobs were correctly filtered
out from evaluation, the approval logic misinterpreted the resulting empty job
list as a failure state.

Design Choices:
- Updated handle_approval to check processed_jobs in addition to ok_jobs. This
  allows distinguishing between truly missing jobs and jobs that were found
  but filtered out.
- Removed the disapproval reason in _handle_not_ready_jobs when jobs are
  filtered, replacing it with an informational log message.
- Added code comments to clarify the safeguard logic in handle_approval.

Benefits:
- Correctly approves increments that are only "blocked" by development jobs.
- Preserves the safety check against approving requests with no identified jobs.

Related issue: https://progress.opensuse.org/issues/200009

### DIFF
--- a/openqabot/incrementapprover.py
+++ b/openqabot/incrementapprover.py
@@ -200,7 +200,10 @@ class IncrementApprover:
         reqid = approval_status.request.reqid
         id_msg = f"OBS request {config.settings.obs_web_url}/request/show/{reqid}"
 
-        if not reasons_to_disapprove and not approval_status.ok_jobs:
+        # Safeguard: only block if NO jobs were ever identified for this request.
+        # If processed_jobs is set but ok_jobs is empty, it means all jobs were filtered
+        # (e.g. development groups) and approval should proceed if there are no other blockers.
+        if not reasons_to_disapprove and not approval_status.ok_jobs and not approval_status.processed_jobs:
             reasons_to_disapprove.append("No openQA jobs were found/checked for this request.")
 
         if self.comment and approval_status.builds:
@@ -448,9 +451,7 @@ class IncrementApprover:
             return 0
 
         if jobs_were_filtered:
-            approval_status.reasons_to_disapprove.append(
-                f"No jobs left for evaluation (all were filtered) for {info_str}"
-            )
+            log.info("All jobs filtered for %s (development groups ignored)", info_str)
             return 0
 
         approval_status.reasons_to_disapprove.append(f"No jobs scheduled for {info_str}")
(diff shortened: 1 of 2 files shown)


nicksinger:
The amount of blabla your LLM generated is astonishing. BTW; where can I find the following change?

> _handle_not_ready_jobs(): Removed the addition of "No jobs left for
evaluation (all were filtered)" to reasons_to_disapprove. When jobs
are filtered, we now only log an informational message and return
without blocking approval.

openqabot/incrementapprover.py
```diff
@@ -200,7 +200,7 @@ def handle_approval(self, approval_status: ApprovalStatus) -> int:
         reqid = approval_status.request.reqid
         id_msg = f"OBS request {config.settings.obs_web_url}/request/show/{reqid}"
 
-        if not reasons_to_disapprove and not approval_status.ok_jobs:
+        if not reasons_to_disapprove and not approval_status.ok_jobs and not approval_status.processed_jobs:
```
asmorodskyi, line 206: **question(non-blocking)** : can you elaborate on this ?  also I would leave this as comment in code .

asmorodskyi (approved):
LGTM

asmorodskyi:
> The amount of blabla your LLM generated is astonishing. BTW; where can I find the following change?
> 
> > _handle_not_ready_jobs(): Removed the addition of "No jobs left for
> > evaluation (all were filtered)" to reasons_to_disapprove. When jobs
> > are filtered, we now only log an informational message and return
> > without blocking approval.

I think we need some limit of log message overall size . No one will read this "book" later ...

openqabot/incrementapprover.py
okurz: when we let a request not block if there are "no jobs left after filtering" we should add the safeguard if no jobs at all were found. I will add an according comment

okurz: done

okurz:
> The amount of blabla your LLM generated is astonishing. BTW; where can I find the following change?
> 
> > _handle_not_ready_jobs(): Removed the addition of "No jobs left for
> > evaluation (all were filtered)" to reasons_to_disapprove. When jobs
> > are filtered, we now only log an informational message and return
> > without blocking approval.

fixed.

nicksinger:
> I think we need some limit of log message overall size . No one will read this "book" later ...

No worries, somebody will have the awesome idea to summarize these lengthy descriptions later with another LLM :+1:

okurz:
> The amount of blabla your LLM generated is astonishing. BTW; where can I find the following change?
> 
> > _handle_not_ready_jobs(): Removed the addition of "No jobs left for
> > evaluation (all were filtered)" to reasons_to_disapprove. When jobs
> > are filtered, we now only log an informational message and return
> > without blocking approval.

fixed. That was outdated content and a wrong reference.

openqabot/incrementapprover.py
kalikiana: It should be "nothing to approve", right?

kalikiana (approved):

PR #503, opened 2026-04-21
fix: drop radon from our pipeline

asmorodskyi:
Radon was the gold standard in 2017, but it hasn't kept pace with the Python language's evolution.
 Continuing to use it risks CI/CD failures as we adopt newer Python versions.  By moving its responsibilities to Ruff, we consolidate our toolchain, and ensure our complexity checks actually understand modern syntax.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,6 @@ dev = [
     "pytest-xdist",
     "pyupgrade",
     "coverage",
-    "radon",
     "responses>=0.21.0",
     "urllib3",
     "ruff",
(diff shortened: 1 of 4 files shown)


okurz (changes requested):
You mentioned " By moving its responsibilities to Ruff" but I see no changes to Ruff. So how do you think this works? Or do you mean to willingly accept the holes that dropping radon will bring?

pyproject.toml
okurz:
```suggestion
max-complexity = 9
```

then I guess we have done enough to "replace radon with ruff". The diff looks weird to me though. Just replace the "10" with "9" :)

okurz (changes requested):

pyproject.toml
asmorodskyi: done

okurz (approved):

Martchus (approved):

PR #504, opened 2026-04-21
refactor: centralize Gitea PR validation and simplify loader logic

okurz:
Alternative to https://github.com/openSUSE/qem-bot/pull/501 with slight refactoring to bring back maintainability score to A.

### DIFF
--- a/openqabot/giteatrigger.py
+++ b/openqabot/giteatrigger.py
@@ -147,11 +147,11 @@ class GiteaTrigger:
                         f"Request accepted for '{config.settings.obs_group}' "
                         f"based on data in {config.settings.dashboard_url()}"
                     )
-                    approve_pr(self.gitea_token, pullrequest.repo_name, pullrequest.number, pullrequest.commit_sha, msg)
+                    approve_pr(self.gitea_token, pullrequest.project, pullrequest.number, pullrequest.commit_sha, msg)
 
     def get_prs_by_label(self) -> None:
         """Get all open PRs and filter them by defined label."""
-        open_prs: list[Any] = get_open_prs(
+        open_prs: list[PullRequest] = get_open_prs(
             self.gitea_token,
             self.gitea_repo,
             number=self.pr_number,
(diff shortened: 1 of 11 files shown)


asmorodskyi (approved):
LGTM

okurz:
> LGTM

ok, then I will squash the commits

EDIT: done

JRivrain:
LGTM*

asmorodskyi (changes requested):
In GiteaSync._on_amqp_message(), the dictionary is correctly parsed into a PullRequest object (pr), but the function call to make_submission_from_gitea_pr() was missed. It still passes message["pull_request"] (a dict), whereas the refactored function now strictly expects a PullRequest instance. This will result in a TypeError or AttributeError at runtime when it tries to access pr.number.
2. Flawed and unused tests in tests/test_pullrequest.py
The two tests test_from_json_request_exception and test_from_json_json_exception are fundamentally flawed:
PullRequest.from_json() only operates on a dictionary and no longer makes any API calls. It correctly catches KeyError but will never throw a RequestException or a JSONDecodeError.
The tests patch the PullRequest class itself, meaning .from_json() becomes a mock object. assert <MagicMock> is None will silently fail or behave incorrectly depending on the Pytest version.
These tests should be completely removed, alongside their unused imports (requests, json, and MockerFixture) to maintain compliance with ruff linters.

asmorodskyi (approved):
all problems addressed in last push

okurz (commented):
Approved

openqabot/giteatrigger.py
```diff
@@ -162,26 +162,13 @@ def get_prs_by_label(self) -> None:
...
-                # we looking only for PRs which has ALL labels defined via '--pr-label' parameter AND
-                # at least one for labels defined in staging.config
-                qa_labels = set(self.staging_config_qa_labels.keys())
-                if pullrequest.has_all_labels(self.pr_required_labels) and pullrequest.has_any_label(qa_labels):
-                    self.prs.append(pullrequest)
-                else:
-                    log.debug("PR %s disregarded (labels: %s)", pr_id, pullrequest.labels)
-            except KeyError:
-                log.exception("Unable to process PR git:%s", pr_id)
+            # we looking only for PRs which has ALL labels defined via '--pr-label' parameter AND
```
nicksinger, line 165:
```suggestion
            # we're looking only for PRs which has ALL labels defined via '--pr-label' parameter AND
```

openqabot/types/pullrequest.py
```diff
@@ -49,3 +53,28 @@ def has_all_labels(self, required_labels: set[str]) -> bool:
...
+                url=pr_json.get("html_url", pr_json.get("url", "")),
+                commit_sha=pr_json.get("head", {}).get("sha", "unknown"),
+            )
+            log.debug("PR git:%i: %s", instance.number, instance)
+        except KeyError:
+            pr_id = pr_json.get("number", "?")
+            log.exception("PR git:%s ignored: Could not read PR metadata", pr_id)
+            return None
+        else:
+            return instance
```
nicksinger, line 80: how about making this whole logic part of the constructor if a json gets passed for object creation?

nicksinger (commented):
this doesn't really feel pythonic and more like the previous style in this project - which might be intentional. Also marked a minor grammar mistake (which might be on purpose as well ;) )

JRivrain (approved):

openqabot/giteatrigger.py
asmorodskyi: I will create follow up PR to fix this

openqabot/types/pullrequest.py
asmorodskyi: but it is some form of constructor in a way python defines them . The problem that python does not support method overloading which means that if you would like to have several ways of creating instance of same class you could get that by defining several methods like : 

```
@classmethod
def <method_name>(cls: type[Self], .....) -> Self
```

this is what I have added . This class does not have explicit `__init__` definition because it has `@dataclass` annotation which would generate init with simply enum all fields of that class . 

So TDLR: It is already done how you asking or I didn't get the question :)

openqabot/giteatrigger.py
asmorodskyi: #506

okurz:
> this doesn't really feel pythonic and more like the previous style in this project - which might be intentional. Also marked a minor grammar mistake (which might be on purpose as well ;) )

I agree. That's mostly on @asmorodskyi though. I merely only added minor refactoring to improve "maintainability". Maybe anyone wants to drive this even further

PR #505, opened 2026-04-21
perf: optimize more commands with parallel execution

okurz:
* perf(approver): parallelize submission filtering and approval
* perf(gitea-loader): parallelize build result fetching
* perf(commenter): parallelize incident commenting

Similar to #495 but independant.

This brings the runtime for one gitea-trigger cycle down from
40280±1654.52 ms to 33785±125.34 ms which is a 16%
improvement (speedup 1.2x).

For one sub-approve cycle this brings the time down from
135907±5291.28 ms to 20724±384.48 ms which is a big 84%
improvement (speedup 6.6x).

### DIFF
--- a/openqabot/loader/gitea.py
+++ b/openqabot/loader/gitea.py
@@ -619,7 +619,7 @@ def add_build_results(submission: dict[str, Any], obs_urls: list[str], *, dry: b
     """Aggregate build results from multiple OBS URLs into a submission."""
     results = BuildResults()
 
-    with futures.ThreadPoolExecutor() as executor:
+    with futures.ThreadPoolExecutor(max_workers=config.settings.max_workers) as executor:
         futs = [executor.submit(_fetch_obs_url_data, url, dry=dry) for url in obs_urls]
         _process_fetched_data(submission, results, futs)
 
@@ -849,7 +849,7 @@ def get_submissions_from_open_prs(
     # configure osc to be able to request build info from OBS
     osc.conf.get_config(override_apiurl=config.settings.obs_url)
 
-    with futures.ThreadPoolExecutor() as executor:
+    with futures.ThreadPoolExecutor(max_workers=config.settings.max_workers) as executor:
         future_sub = [
             executor.submit(
                 make_submission_from_gitea_pr,
--- a/openqabot/loader/gitea.py
+++ b/openqabot/loader/gitea.py
@@ -541,45 +541,56 @@ def is_build_result_relevant(res: etree._Element, relevant_archs: set[str] | Non
     return arch == "local" or relevant_archs is None or arch in relevant_archs
 
 
-def _get_project_results(obs_project: str, *, dry: bool, results: BuildResults) -> list[etree._Element]:
-    """Fetch build results for an OBS project."""
+def _get_project_results(obs_project: str, *, dry: bool) -> tuple[list[etree._Element], bool]:
+    """Fetch build results for an OBS project. Returns a tuple of (results, is_unavailable)."""
     build_info_url = osc.core.makeurl(config.settings.obs_url, ["build", obs_project, "_result"])
     if dry:
-        return read_xml("build-results-124-" + obs_project).getroot().findall("result")
+        return read_xml("build-results-124-" + obs_project).getroot().findall("result"), False
     try:
-        return osc.util.xml.xml_parse(http_GET(build_info_url)).getroot().findall("result")
+        return osc.util.xml.xml_parse(http_GET(build_info_url)).getroot().findall("result"), False
     except urllib.error.HTTPError:
-        results.unavailable.add(obs_project)
         log.info("Build results for project %s unreadable, skipping: %s", obs_project, build_info_url)
-        return []
+        return [], True
 
 
-def _process_obs_url(
+def _fetch_obs_url_data(
     url: str,
-    submission: dict[str, Any],
     *,
     dry: bool,
-    results: BuildResults,
-) -> None:
-    """Process an OBS URL and update submission build results."""
+) -> tuple[str | None, set[str] | None, list[etree._Element], bool]:
+    """Fetch all necessary data for an OBS URL concurrently."""
     if not (project_match := OBS_PROJECT_SHOW_REGEX.search(url)):
-        return
+        return None, None, [], False
     obs_project = project_match.group(1)
     log.debug("Checking OBS project %s", obs_project)
     relevant_archs = determine_relevant_archs_from_multibuild_info(obs_project, dry=dry)
 
-    for res in _get_project_results(obs_project, dry=dry, results=results):
-        if is_build_result_relevant(res, relevant_archs):
-            add_build_result(submission, res, results)
+    xml_results, is_unavailable = _get_project_results(obs_project, dry=dry)
+    return obs_project, relevant_archs, xml_results, is_unavailable
 
 
-def add_build_results(submission: dict[str, Any], obs_urls: list[str], *, dry: bool) -> None:
-    """Aggregate build results from multiple OBS URLs into a submission."""
-    results = BuildResults()
+def _process_fetched_data(
+    submission: dict[str, Any],
+    results: BuildResults,
+    futs: list[futures.Future[tuple[str | None, set[str] | None, list[etree._Element], bool]]],
+) -> None:
+    """Process data fetched concurrently in the main thread."""
+    for fut in futures.as_completed(futs):
+        obs_project, relevant_archs, xml_results, is_unavailable = fut.result()
 
-    for url in obs_urls:
-        _process_obs_url(url, submission, dry=dry, results=results)
+        if obs_project is None:
+            continue
 
+        if is_unavailable:
+            results.unavailable.add(obs_project)
+
+        for res in xml_results:
+            if is_build_result_relevant(res, relevant_archs):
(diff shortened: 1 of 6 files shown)


asmorodskyi (dismissed):
why do we need care so much for CPU ticks for script which runs once a while for several minutes ?

okurz:
> why do we need care so much for CPU ticks for script which runs once a while for several minutes ?

For multiple reasons:
1. Improved runtime for running the applications for the purpose of development matters a lot. Faster execution means humans wait less and can develop more efficiently. Given that we want to use qem-bot for additional purposes, e.g. openSUSE maintenance, this is significant.
2. We plan to trigger executions by AMQP events as well as time based so coming up with a proper architecture of which actions can actually be parallized matter.
3. We already had repeated problems that some actions occupied CI runners for long so making the default operations even quicker helps.
4. Using a functional style of how tasks can be executed matches the business domain better than pretending that independant entities need to be processed procedurally.
5. Saving execution time saves energy which is always good.

asmorodskyi:
> 1. Improved runtime for running the applications for the purpose of development matters a lot. Faster execution means humans wait less and can develop more efficiently. Given that we want to use qem-bot for additional purposes, e.g. openSUSE maintenance, this is significant.

-  have you tried to debug MT code ? This brings its own set of problems in case of MT and this will **slowdown** development process 
- there are other ways to speedup development process . Like mocking slow services . In a fact if you would follow TDD approach which you often promote you should not face with such problem during development process 

> 2. We plan to trigger executions by AMQP events as well as time based so coming up with a proper architecture of which actions can actually be parallized matter.

agree , but just replacing every loop in code with MT is not right way to go here. Let's do such changes in case by case basis in commits which actually make use of this for AMQP. 

> 3. We already had repeated problems that some actions occupied CI runners for long so making the default operations even quicker helps.

this problem is totally unrelated here . and can not be solved by MT . if gitlab jobs has TTL for several hours and we configure jobs to do retries for several hours does not matter if script runs for 10 minutes or 2 CI runners would be overbooked with ANY qem-bot error 

> 4. Using a functional style of how tasks can be executed matches the business domain better than pretending that independant entities need to be processed procedurally.

you can not say that you using functional style by replacing several loops with MT ! to get benefits from functional programming you need to have ALL code written in that way . when only tiny pieces of your code follow that and others don't it is not functional style it is just a mess ...

okurz:
Thanks for your review @asmorodskyi, and I  understand the general hesitation around multithreading. However, I'd like to clarify the specific approach taken here, as it avoids the typical pitfalls of MT.

> > 1. Improved runtime for running the applications for the purpose of development matters a lot. Faster execution means humans wait less and can develop more efficiently. Given that we want to use qem-bot for additional purposes, e.g. openSUSE maintenance, this is significant.
> 
>     * have you tried to debug MT code ? This brings its own set of problems in case of MT and this will **slowdown** development process

Yes, I have tried debugging MT code. But here we are talking about code that has no complexity of mutex, locks, shared resources, etc. I completely agree that multithreading can make debugging painful. However, the intention here was specifically to parallelize I/O-bound network requests (waiting on the OBS/Gitea APIs) where the threads just sit idle waiting for HTTP responses, rather than complex CPU-bound work. Aside from BuildResults, the operations are stateless. One can always set `max_workers=1` in the ThreadPoolExecutor if you really need to debug this specific execution.

>     * there are other ways to speedup development process . Like mocking slow services . In a fact if you would follow TDD approach which you often promote you should not face with such problem during development process

I do follow TDD but one needs to compare the mocked operations against the real-life expectations. What I do is https://en.wikipedia.org/wiki/Acceptance_test-driven_development
While mocking is absolutely the right approach for our automated tests using TDD, the goal of this PR is to speed up the actual execution of the script. When developers run this locally against real instances to test their changes, waiting 135s vs 20s makes a massive difference in iteration speed.

> > 3. We already had repeated problems that some actions occupied CI runners for long so making the default operations even quicker helps.
> 
> this problem is totally unrelated here . and can not be solved by MT . if gitlab jobs has TTL for several hours and we configure jobs to do retries for several hours does not matter if script runs for 10 minutes or 2 CI runners would be overbooked with ANY qem-bot error

While tuning CI retry logic and TTLs is important, executing the core tasks multiple times faster inherently frees up runners much quicker. For AMQP, these specific I/O loops are exactly the ones that will benefit from not blocking sequentially. This of course does not take away the need to prevent and fix any continuously blocked executions.

> > 4. Using a functional style of how tasks can be executed matches the business domain better than pretending that independant entities need to be processed procedurally.
> 
> you can not say that you using functional style by replacing several loops with MT ! to get benefits from functional programming you need to have ALL code written in that way . when only tiny pieces of your code follow that and others don't it is not functional style it is just a mess ...

Given that the implementation relies on standard, high-level Python concurrent futures without any low-level thread management or state-sharing, the maintenance burden is minimal while the performance gain is significant (an 84% speedup in one case). 
Because the complexity is so low and the immediate value is so high, I would strongly prefer to keep these optimizations. If debugging the network requests ever does become an issue, we could easily add a --sequential CLI flag in a future PR to force sequential execution during local debugging.

openqabot/loader/gitea.py
```diff
@@ -521,8 +521,8 @@ def add_build_results(submission: dict[str, Any], obs_urls: list[str], *, dry: b
     """Aggregate build results from multiple OBS URLs into a submission."""
     results = BuildResults()
 
-    for url in obs_urls:
-        _process_obs_url(url, submission, dry=dry, results=results)
+    with futures.ThreadPoolExecutor() as executor:
+        list(executor.map(lambda url: _process_obs_url(url, submission, dry=dry, results=results), obs_urls))
```
Martchus: This code constructs a `list` it doesn't use. Is there a better way to consume the generator?

Maybe this shouldn't use `map` if it doesn't actually use the return value of this function or it should use `map` but also use its return value. The latter alternative might also help with the locking problems mentioned in my other comment.

openqabot/approver.py
```diff
@@ -130,16 +131,20 @@ def __call__(self) -> int:
         )
 
         overall_result = True
-        submissions_to_approve = [sub for sub in subreqs if self.approvable(sub)]
+        with ThreadPoolExecutor() as executor:
+            is_approvable = list(executor.map(self.approvable, subreqs))
```
Martchus: `is_approvable` sounds like this variable contains a bool value but here a list is assigned.

openqabot/loader/gitea.py
Martchus: This modifies `results` from different threads. That probably needs locking. You probably don't see any problem right now because:

> Note: Operations like appending to a list (my_list.append(item)) or fetching an item from a dictionary are generally atomic in CPython and won't get interrupted mid-operation, but relying on this is generally considered bad practice for complex state management.

That's from Gemini so we should cross-check whether it is actually correct. Gemini also made another interesting point btw:

> **A Modern Caveat: Python 3.13+ and Free-Threading**
> It is worth noting that the Python ecosystem is currently undergoing a massive shift. Starting with Python 3.13, CPython introduced an experimental Free-Threading (No-GIL) mode. In a free-threaded Python build, the implicit GIL is completely removed to allow true multicore parallelism. In these environments, protecting shared variables with explicit locks is even more critical, as threads can now truly execute bytecode simultaneously.

---

So I think we should avoid bad practices here - even though it works right now - and especially because it might cease to work when running the code in No-GIL mode.

Martchus (changes requested):
The number of threads should be configurable.

A quick research with Gemini also showed that one needs to be careful with multi-threading - even in Python. I don't think you've checked whether shared variables are accessed safely, see my inline comment for details.

r-richardson:
Just a note, i see that https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor has a `max_workers` variable that defaults to `min(32, (os.process_cpu_count() or 1) + 4)`, while https://urllib3.readthedocs.io/en/latest/advanced-usage.html#customizing-pool-behavior states that for urllib3 "The PoolManager class ... will keep a maximum of 10 ConnectionPool instances." by default.

openqabot/loader/gitea.py
```diff
@@ -495,45 +495,56 @@ def is_build_result_relevant(res: etree._Element, relevant_archs: set[str] | Non
...
-def _process_obs_url(
+def _fetch_obs_url_data(
     url: str,
-    submission: dict[str, Any],
     *,
     dry: bool,
-    results: BuildResults,
-) -> None:
-    """Process an OBS URL and update submission build results."""
+) -> tuple[str | None, set[str] | None, list[etree._Element], bool]:
```
asmorodskyi, line 560: this does not looks like improvement . If we will merge this it would be really hard to understand what and where this function is returning

asmorodskyi (changes requested):

asmorodskyi:
I said that I would not stop you from introducing MT but it does not mean that I would agree with overcomplicating of code base to allow MT

openqabot/loader/gitea.py
okurz: done

openqabot/approver.py
okurz: done

okurz:
Updated to address two minor comments. I will create separate PRs for just *one* command at a time
* approve: #593
* `max-workers` CLI option: #594

PR #507, opened 2026-04-28
feat(gitea): support trigger configurations

asmorodskyi:
- Add `--trigger-config` option to `gitea-trigger` for external YAML-based configuration.
- Refactor `GiteaTrigger` to iterate over multiple configurations per pull request.
- Introduce `IsoMatch` and `TriggerConfig` types for better data abstraction in triggering.
- Standardize configuration loading using a new `get_configs_from_path` utility.
- Consolidate and improve test coverage for increment types, submissions, and the main bot.


okurz (changes requested):
This PR seems to mix some changes together, e.g. why did you remove tests/test_openqabot_simple.py and merge it into another test file? Maybe you can split out some minor changes into separate commits?

asmorodskyi:
> This PR seems to mix some changes together, e.g. why did you remove tests/test_openqabot_simple.py and merge it into another test file? Maybe you can split out some minor changes into separate commits?

I tried to keep functional code changes in separate commits, but I assumed that test cleanups and consolidations were less critical to isolate.

Additionally, these test changes were suggested by an AI assistant alongside the fixes for my code changes. Separating them now might be a bit tricky and could require extra work to ensure tests pass at every commit stage. If you feel strongly about it, I can revert the test refactoring that isn't directly related to this PR to keep things focused.

openqabot/loader/incrementconfig.py
```diff
@@ -100,75 +94,41 @@ def from_config_entry(entry: dict[str, Any]) -> IncrementConfig:
...
+        Only override when CLI args differ from expected defaults (not dataclass defaults).
+        This allows configs to specify their own values while still supporting CLI overrides.
+        """
+        cli_defaults = {"distri": "sle", "version": "any", "flavor": "any", "arch": "any"}
+        overrides = {}
+
+        for field_name, default_value in cli_defaults.items():
+            if hasattr(args, field_name):
+                arg_value = getattr(args, field_name)
+                if arg_value != default_value:
```
okurz: why did you revert the use of the walrus operator?

openqabot/args.py
```diff
@@ -393,13 +400,21 @@ def gitea_trigger(  # noqa: PLR0913
     fallback_contact: fallback_contact_option = None,
     generic_tool_issues_contact: generic_tool_issues_contact_option = None,
     max_detailed_comment_entries: max_detailed_comment_entries_option = None,
+    trigger_config: Annotated[
+        Path,
+        typer.Option(
+            "--trigger-config",
+            help="YAML document with necessary values needed for triggering",
```
okurz: what do you mean with "necessary values"? In before we did not need this option so this should be optional. Maybe just the phrasing of the help text needs to be improved

openqabot/giteatrigger.py
```diff
@@ -94,45 +97,54 @@ def check_pullrequest(self, pullrequest: PullRequest) -> None:
...
+        data = Data(
+            pullrequest.number,
+            "git",
+            0,
+            trigger_config.flavor,
+            matched_iso.arch,
+            trigger_config.distri,
+            matched_iso.version,
+            matched_iso.build,
+            matched_iso.product,
```
okurz: If we pass multiple values from "trigger_config" and "matched_iso" consider to just pass "trigger_config" and "matched_iso" as complete objects to prevent this statement spread over many lines

okurz (changes requested):

openqabot/loader/incrementconfig.py
asmorodskyi: no reason , reverted . thanks for catching this !

openqabot/args.py
asmorodskyi: well it is not optional actually . I wonder why do we need to play this "backward compatibility" game when we have single pipeline in whole world which would fail due to this and it is quite straightforward to fix it  . 

After thinking two times about this I realized that I am doing same mistake and introduce new option while I can simply use `--config` here . This would solve problem with help text but not question about change bot behavior to get required option

okurz: Please ensure that `qem-bot $cmd` always does something useful and does not have any mandatory option, same as every other command. Similar to what I just fixed in https://github.com/openSUSE/qem-bot/pull/509

openqabot/utils.py
```diff
@@ -54,6 +67,43 @@ def get_yml_list(path: Path) -> list[Path]:
     return [f for ext in ("yml", "yaml") for f in path.glob("*." + ext)] if path.is_dir() else [path]
 
 
+def get_configs_from_path(
```
okurz: as discussed better move it to openqabot/loader/config.py, possibly also the above `get_yml_list`

openqabot/loader/incrementconfig.py
```diff
@@ -100,75 +94,39 @@ def from_config_entry(entry: dict[str, Any]) -> IncrementConfig:
             diff_repo_template=entry.get("diff_repo_template", ""),
         )
 
-    @staticmethod
-    def from_config_file(file_path: Path, *, load_defaults: bool = True) -> Iterator[IncrementConfig]:
```
okurz, line 141: as discussed I prefer to not change the style to a procedural

```
configs = []
for i in …:
     …
     configs.extend(…)
```

but keep the list comprehension, e.g.

```
configs = [… for … in …]
```

openqabot/types/isomatch.py
```diff
@@ -0,0 +1,29 @@
...
+    build: str
+
+    def __init__(self, matched_iso: Match[str], pr_number: int) -> None:
+        """Initialize IsoMatch from a regex match and PR number.
+
+        Args:
+            matched_iso (Match[str]): The regex match from the ISO filename.
+            pr_number (int): The pull request number.
+
+        """
```
okurz: remove

openqabot/types/isomatch.py
```diff
@@ -0,0 +1,29 @@
...
+
+class IsoMatch:
+    """Information extracted from a matched ISO filename."""
+
+    product: str
+    version: str
+    arch: str
+    build: str
+
+    def __init__(self, matched_iso: Match[str], pr_number: int) -> None:
```
okurz:
```suggestion
    def __init__(self, iso_match: Match[str], pr_number: int) -> None:
```

openqabot/args.py
okurz: As discussed even better would be to just not add it at all and rely on the global "configs" that already exists

openqabot/giteatrigger.py
asmorodskyi: done

openqabot/args.py
asmorodskyi: done

openqabot/types/isomatch.py
asmorodskyi: but wait I can not remove it . ruff will complain :shrug:  or you want me to use `# noqa: D107` ?

openqabot/utils.py
asmorodskyi: done

openqabot/loader/incrementconfig.py
asmorodskyi: done

openqabot/types/isomatch.py
asmorodskyi: done

okurz (approved):

Wabri (approved):

PR #509, opened 2026-04-28
feat: auto-discover increment configs from global directory

okurz:
Motivation:
Users (like me) can forget to specify --increment-config for
increment-approve, leading to confusing 'No builds found' errors. Since
the global configuration directory is already known via the -c flag (or
.env), it should be used as a default source for increment
configurations.

Design Choices:
* Updated IncrementConfig.from_args to check args.configs if
  args.increment_config is omitted.
* Loaded configurations are still filtered by explicit CLI overrides.
* Added an explicit __init__ to Settings to help the 'ty' type checker.

Benefits:
* Improved DX: increment-approve now automatically finds relevant
  configurations based on the project's metadata.

### DIFF
--- a/openqabot/config.py
+++ b/openqabot/config.py
@@ -31,6 +31,13 @@ def get_default_obs_url() -> str:
 class Settings(BaseSettings):
     """Configuration settings managed by Pydantic."""
 
+    def __init__(self, *args: Any, **kwargs: Any) -> None:  # noqa: ANN401
+        """Initialize settings with optional overrides.
+
+        This explicit constructor helps type checkers like 'ty' recognize valid parameters.
+        """
+        super().__init__(*args, **kwargs)
+
     # Global options
     configs: Path = Field(default=Path("/etc/openqabot"), alias="QEM_BOT_CONFIGS")
     dry: bool = Field(default=False, alias="QEM_BOT_DRY")
--- a/tests/test_loader_incrementconfig.py
+++ b/tests/test_loader_incrementconfig.py
@@ -85,6 +85,25 @@ def test_config_parsing_from_args_with_path(config_index: int, expected_distri: 
     assert additional_settings.items() <= config.settings.items()
 
 
+def test_config_parsing_from_args_with_auto_discovery() -> None:
+    path = Path("tests/fixtures/config-increment-approver")
+    configs = IncrementConfig.from_args(
+        Namespace(increment_config=None, configs=path, distri="sle", version="16.0", flavor="Online-Increments")
+    )
+    assert len(configs) == 2
+    assert configs[0].distri == "foo"
+    assert configs[1].distri == "bar"
+
+
+def test_config_parsing_from_args_fallback_to_cli() -> None:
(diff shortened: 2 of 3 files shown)


asmorodskyi (changes requested):
**question(blocking)** : what the point to have two params pointing to the same directory ? can't we just use single one ? Do we actually support case when this two params has different values and how this would affect the flow ?

okurz:
> **question(blocking)** : what the point to have two params pointing to the same directory ? can't we just use single one ? Do we actually support case when this two params has different values and how this would affect the flow ?

I am not sure about that aspect though this is nothing that should be impacted by this PR. I am happy to follow up with that topic in https://github.com/openSUSE/qem-bot/issues/511 where we can have a focussed discussion with the original author.

asmorodskyi (approved):
the fact that now we have an issue where it would be tracked is totally fine for me so unblock this change

tests/test_loader_incrementconfig.py
```diff
@@ -85,6 +85,25 @@ def test_config_parsing_from_args_with_path(config_index: int, expected_distri:
...
+
+def test_config_parsing_from_args_fallback_to_cli() -> None:
+    path = Path("tests/fixtures/config")
+    configs = IncrementConfig.from_args(
+        Namespace(increment_config=None, configs=path, distri="sle", version="16.0", flavor="Online-Increments")
+    )
+    assert len(configs) == 1
+    assert configs[0].distri == "sle"
+
+
```
d3flex, line 106: those two could be merged. but it s fine

openqabot/config.py
```diff
@@ -91,6 +91,13 @@ def dashboard_url(self, *path: str | int) -> str:
     )
     max_detailed_comment_entries: int = Field(default=7, alias="QEM_MAX_DETAILED_COMMENT_ENTRIES")
 
+    def __init__(self, **kwargs: Any) -> None:  # noqa: ANN401
+        """Initialize settings with optional overrides.
+
+        This explicit constructor helps type checkers like 'ty' recognize valid parameters.
+        """
+        super().__init__(**kwargs)
```
nicksinger:
```suggestion
    def __init__(self, *args, **kwargs: Any) -> None:  # noqa: ANN401
        """Initialize settings with optional overrides.

        This explicit constructor helps type checkers like 'ty' recognize valid parameters.
        """
        super().__init__(*args, **kwargs)
```

I don't think it matters much but if we include this to make some tool happy, we should do it properly :)

d3flex (approved):
Approved but..

nicksinger (commented):
I also tend to put the `__init__`-definitions on top of the class

tests/test_loader_incrementconfig.py
okurz: here I think separate tests is better as the test function name is directly selectable in pytest and also rendered in output on failures.

openqabot/config.py
okurz: Sure. Done. Also moved `__init__` to the top

PR #513, opened 2026-04-29
perf: deduplicate openqa product increment requests

okurz:
Motivation:
When checking for new `.spdx.json` files during `increment-approve`, the bot
repeatedly performed the exact same HTTP requests to the OBS API for each
configuration variation (flavors, architectures). This caused severe slowdowns
and increased the chances of timeouts.

Design Choices:
Refactored `openqabot/incrementapprover.py` to group `IncrementConfig` items
by their shared OBS directory URLs and regex filters before fetching the
listing. Moved the architecture and flavor filtering out of the
`load_build_info` function so that it returns all builds found. This allows
processing multiple configs with just a single API call per unique OBS path.

Benefits:
The bot makes significantly fewer HTTP requests during increment approval
checking, vastly improving runtime execution speed and reducing the load on
the OBS APIs.

### DIFF
--- a/openqabot/incrementapprover.py
+++ b/openqabot/incrementapprover.py
@@ -25,7 +25,7 @@ from openqabot.pc_helper import apply_public_cloud_settings
 from .commenter import Commenter
 from .errors import AmbiguousApprovalStatusError, PostOpenQAError
 from .loader.buildinfo import load_build_info
-from .loader.incrementconfig import IncrementConfig
+from .loader.incrementconfig import GroupKey, IncrementConfig
 from .loader.sourcereport import compute_packages_of_request_from_source_report
 from .repodiff import Package, RepoDiff
 from .requests import find_request_on_obs
@@ -529,7 +529,9 @@ class IncrementApprover:
         self.requests_to_approve[request_id] = status
         return status
 
-    def process_request_for_config(self, request: osc.core.Request | None, config_inc: IncrementConfig) -> int:
+    def process_request_for_config(
+        self, request: osc.core.Request | None, config_inc: IncrementConfig, build_infos: set[BuildInfo]
+    ) -> int:
         """Process an OBS request for a specific configuration."""
         if request is None:
             return 0
@@ -537,14 +539,8 @@ class IncrementApprover:
         approval_status = self._get_approval_status(request)
         error_count = 0
         found_relevant_build = False
-        for build_info in load_build_info(
-            config_inc,
-            config_inc.build_regex,
-            config_inc.product_regex,
-            config_inc.version_regex,
-            self.get_regex_match,
-        ):
-            if len(config_inc.archs) > 0 and build_info.arch not in config_inc.archs:
+        for build_info in build_infos:
+            if not config_inc.accepts_build_info(build_info):
                 continue
             found_relevant_build = True
             error_count += self.process_build_info(config_inc, build_info, request, approval_status)
@@ -561,6 +557,8 @@ class IncrementApprover:
         single_request = (
             osc.core.Request.from_api(config.settings.obs_url, self.args.request_id) if self.args.request_id else None
         )
+
+        grouped_configs: dict[GroupKey, list[IncrementConfig]] = defaultdict(list)
         for config_inc in self.config:
             if single_request and single_request.actions[0].src_project != config_inc.build_project():
                 log.debug(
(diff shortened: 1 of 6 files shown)


openqabot/loader/incrementconfig.py
```diff
@@ -70,6 +70,20 @@ def build_project_url(self, base_url: str | None = None) -> str:
         base_path = self.build_project().replace(":", ":/")
         return f"{base_url or config.settings.obs_download_url}/{base_path}"
 
+    @property
+    def group_key(self) -> tuple[str, str, str, str, str, str, str, str]:
```
asmorodskyi:
```suggestion
    def group_key(self) -> str:
```

asmorodskyi: let's just generate a string combining all that . it would give same effect but will look less scary ( at least to me )

asmorodskyi (changes requested):

openqabot/loader/incrementconfig.py
```diff
@@ -70,6 +70,20 @@ def build_project_url(self, base_url: str | None = None) -> str:
...
+        """Return a unique key for grouping configs."""
+        return (
+            self.build_project(),
+            self.build_project_url(),
```
asmorodskyi: self.build_project_url() is actually created based on self.build_project() so I can not follow logic why you need both

openqabot/loader/incrementconfig.py
```diff
@@ -70,6 +70,20 @@ def build_project_url(self, base_url: str | None = None) -> str:
...
+            self.build_listing_sub_path,
+            self.build_regex,
```
asmorodskyi: build_regex is actually contains , product_regex and version_regex within . I have already discussed this with @Martchus  and if I not mistaken he generally agreed that in a fact we can drop other regex's and just get all necessary info from build_regex . So let's not create yet another place which using this not necessary complication

openqabot/incrementapprover.py
```diff
@@ -517,21 +517,22 @@ def _get_approval_status(self, request: osc.core.Request) -> ApprovalStatus:
...
         found_relevant_build = False
-        for build_info in load_build_info(
-            config_inc,
-            config_inc.build_regex,
-            config_inc.product_regex,
-            config_inc.version_regex,
-            self.get_regex_match,
-        ):
+        for build_info in build_infos:
+            if not all(
```
asmorodskyi:
```
  if not all(
                getattr(config_inc, k) in {"any", getattr(build_info, k)}
                for k in ("distri", "flavor", "version", "arch")
            ):
                continue
            if len(config_inc.archs) > 0 and build_info.arch not in config_inc.archs:
                continue
```

this looks hardly readable , especially in context that we checking buildinfo.arch two times now one time for `IncrementConfig.arch == build_info.arch` and another time for `buildinfo.arch in IncrementConfig.archs ` 🤔  what about method in IncrementConfig which would be single point of truth whether certain buildinfo is fine for certain IncrementConfig ?

openqabot/loader/incrementconfig.py
d3flex: I agree with @asmorodskyi but rather than str this seems like it could use its own type

okurz: I've refactored group_key to use a typed GroupKey frozen dataclass as suggested by @d3flex. This provides the same grouping efficiency as a concatenated string but keeps the code type-safe and more readable, as it avoids manually managing delimiters which we would need to do with a simple str.

openqabot/loader/incrementconfig.py
okurz: Good catch. I've removed build_project_url from the grouping key since it is redundant.

openqabot/loader/incrementconfig.py
okurz: ok, I refactored the loading logic to separate network fetching from filtering. This allows qem-bot to perform a single OBS listing request per shared project path and build_regex. The configuration-specific filtering for product_regex and version_regex has been moved into the new accepts_build_info method. This removes the unnecessary complication from the grouping logic but we still need those to ensure that overlapping configs still get the correct data which use those keys.

openqabot/incrementapprover.py
okurz: Done. I've implemented IncrementConfig.accepts_build_info(build_info) to encapsulate all filtering logic (architecture, flavor, and regex matching). This significantly cleaned up the loop in IncrementApprover

okurz:
I've addressed the review feedback by:
1. Implementing a typed GroupKey dataclass
2. Moving filtering logic into IncrementConfig.accepts_build_info to act as a single "point of truth"
3. Further optimizing the loops to ensure redundant OBS API calls (like find_request_on_obs) are executed only once per group
4. Adding regression tests for the regex filtering logic to show how we still need those multiple "regex" keys

Martchus (approved):

asmorodskyi:
looks much better now ! Still want to discuss extra loop which you have added in IncrementApprover but I prefer to discuss it in a call because it is easy to get it wrong :)

asmorodskyi (approved):
LGTM

PR #514, opened 2026-04-29
feat: support architecture filtering for additional builds

okurz:
Motivation:
Additional build configurations in product increment YAMLs could specify an
'archs' list, but it was being ignored by the bot. This meant all additional
builds were attempted for all architectures present in the source report,
even if they weren't supported. Also, placeholder packages in noarch source
reports were triggering redundant build attempts.

Design Choices:
Updated `_match_additional_build` in `openqabot/incrementapprover.py` to
respect the 'archs' key if present in the additional build definition.
Added a check to `extra_builds_for_package` to skip packages with empty
version strings (common for placeholders). Added unit tests for the new
architecture filtering logic.

Benefits:
Correctly restricts additional builds (like KLPs) to their supported
architectures and reduces incorrect test scheduling attempts for
placeholder packages.

### DIFF
--- a/openqabot/incrementapprover.py
+++ b/openqabot/incrementapprover.py
@@ -236,6 +236,9 @@ class IncrementApprover:
         build_info: BuildInfo,
     ) -> dict[str, str] | None:
         """Check if an additional build matches a package and return extra parameters."""
+        if (archs := additional_build.get("archs")) and package.arch not in archs:
+            return None
+
         package_name_match = self._match_package_name_and_version(package, additional_build)
         if package_name_match is None:
             return None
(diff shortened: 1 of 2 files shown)


Martchus (commented):
The change regarding version filtering should probably be a separate commit - although I also don't really care.

The commit message mentions source reports. Are we still using this approach? I don't think we do. This is also probably more generic so I'd just say "all detected archs" instead of mentioning where they come from.

Martchus (commented):
The change regarding version filtering should probably be a separate commit - although I also don't really care.

The commit message mentions source reports. Are we still using this approach? I don't think we do. This is also probably more generic so I'd just say "all detected archs" instead of mentioning where they come from.

asmorodskyi (approved):
LGTM

okurz:
> The change regarding version filtering should probably be a separate commit - although I also don't really care.

done

> The commit message mentions source reports. Are we still using this approach? I don't think we do. This is also probably more generic so I'd just say "all detected archs" instead of mentioning where they come from.

done

Martchus (approved):

PR #515, opened 2026-04-29
refactor: improve maintainability of KLP matching logic

okurz:
Motivation:
The previous implementation of KLP matching was procedural and relied on
"magic" hardcoded rules. This created technical debt and made the code
difficult to reason about or extend.

Design Choices:
Extracted hardcoded business rules into descriptive, named predicates
(`_is_initial_version`, `_is_placeholder`, `_is_debug_asset`). Refactored
`_match_additional_build` to use a functional pipeline for constructing the
openQA `BUILD` string, replacing sequential `if` blocks. Consolidated
duplicated tests into a single parametrized, data-driven test suite.

Benefits:
Achieves higher maintainability and readability. The intent of exclusion
rules is now explicit, and the test suite is significantly more concise and
easier to maintain.

Before:
* #516

### DIFF
--- a/openqabot/incrementapprover.py
+++ b/openqabot/incrementapprover.py
@@ -239,25 +239,24 @@ class IncrementApprover:
         if (archs := additional_build.get("archs")) and package.arch not in archs:
             return None
 
-        package_name_match = self._match_package_name_and_version(package, additional_build)
-        if package_name_match is None:
+        if not (package_name_match := self._match_package_name_and_version(package, additional_build)):
             return None
 
         groups = package_name_match.groupdict()
-        extra_build = [f"PI-{build_info.build}", additional_build["build_suffix"]]
-        extra_params: dict[str, str] = {}
+        kernel_version = (groups.get("kernel_version") or "").replace("_", ".")
 
-        if kind := groups.get("kind"):
-            extra_build.append(kind)
+        buildname_parts = [
+            f"PI-{build_info.build}",
+            additional_build["build_suffix"],
+            groups.get("kind"),
+            kernel_version,
+        ]
 
-        if kernel_version := groups.get("kernel_version"):
-            kernel_version = kernel_version.replace("_", ".")
-            extra_build.append(kernel_version)
-            extra_params["KERNEL_VERSION"] = kernel_version
+        params = {"BUILD": "-".join(str(part) for part in buildname_parts if part)}
+        if kernel_version:
+            params["KERNEL_VERSION"] = kernel_version
 
-        extra_params["BUILD"] = "-".join(extra_build)
-        extra_params.update(additional_build["settings"])
-        return extra_params
+        return params | additional_build["settings"]
 
     def _match_package_name_and_version(self, package: Package, additional_build: dict[str, Any]) -> re.Match | None:
         """Match package name and version against regexes."""
@@ -277,15 +276,17 @@ class IncrementApprover:
         build_info: BuildInfo,
     ) -> dict[str, str] | None:
         """Determine extra build parameters for a specific package."""
-        if not package.version or re.match(r"^1(?:\..*)?$", package.version):
-            return None
-        if "debug" in package.name or package.arch in {"src", "nosrc"}:
+        if package.is_placeholder or package.is_initial_version or package.is_debug_asset:
             return None
 
-        for additional_build in config_inc.additional_builds:
-            if (res := self._match_additional_build(package, additional_build, build_info)) is not None:
-                return res
-        return None
+        return next(
+            (
+                res
+                for additional_build in config_inc.additional_builds
+                if (res := self._match_additional_build(package, additional_build, build_info)) is not None
+            ),
+            None,
+        )
 
     def extra_builds_for_additional_builds(
         self,
--- a/openqabot/repodiff.py
+++ b/openqabot/repodiff.py
@@ -41,6 +41,21 @@ class Package(NamedTuple):
     rel: str
     arch: str
 
+    @property
+    def is_initial_version(self) -> bool:
+        """Check if package is an initial version."""
+        return bool(self.version and re.match(r"^1(?:\..*)?$", self.version))
+
+    @property
+    def is_placeholder(self) -> bool:
+        """Check if package version is a placeholder (empty)."""
+        return not self.version
+
+    @property
+    def is_debug_asset(self) -> bool:
+        """Check if package is a debug asset."""
+        return "debug" in self.name or self.arch in {"src", "nosrc"}
+
 
 class RepoDiff:
     """Repository diff computation."""
--- a/tests/test_incrementapprover_helpers.py
+++ b/tests/test_incrementapprover_helpers.py
@@ -4,7 +4,7 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 import pytest
 import responses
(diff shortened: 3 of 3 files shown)


openqabot/incrementapprover.py
```diff
@@ -239,25 +239,24 @@ def _match_additional_build(
...
+            additional_build["build_suffix"],
+            groups.get("kind"),
+            kernel_version,
+        ]
 
-        if kernel_version := groups.get("kernel_version"):
-            kernel_version = kernel_version.replace("_", ".")
-            extra_build.append(kernel_version)
-            extra_params["KERNEL_VERSION"] = kernel_version
+        params = {"BUILD": "-".join(filter(None, build_parts))}
```
nicksinger: 1. I'm surprised this didn't get caught by our style-rules
2. This will break if any of the values in `build_parts` are non-strings

So maybe just use a list comprehension:
```suggestion
        params = {"BUILD": "-".join([str(part) for part in build_parts if part])}
```

openqabot/incrementapprover.py
```diff
@@ -239,25 +239,24 @@ def _match_additional_build(
...
-        package_name_match = self._match_package_name_and_version(package, additional_build)
-        if package_name_match is None:
+        if not (match := self._match_package_name_and_version(package, additional_build)):
             return None
 
-        groups = package_name_match.groupdict()
-        extra_build = [f"PI-{build_info.build}", additional_build["build_suffix"]]
-        extra_params: dict[str, str] = {}
+        groups = match.groupdict()
+        kernel_version = (groups.get("kernel_version") or "").replace("_", ".")
```
nicksinger, line 246:
```suggestion
        kernel_version = groups.get("kernel_version", "").replace("_", ".")
```

openqabot/incrementapprover.py
```diff
@@ -239,25 +239,24 @@ def _match_additional_build(
...
 
-        if kind := groups.get("kind"):
-            extra_build.append(kind)
+        build_parts = [
```
nicksinger:
```suggestion
        buildname_parts = [
```
minor nitpick but `build_parts` sounds so imperative to me

openqabot/incrementapprover.py
```diff
@@ -270,22 +269,36 @@ def _match_package_name_and_version(self, package: Package, additional_build: di
...
         self,
         package: Package,
         config_inc: IncrementConfig,
         build_info: BuildInfo,
     ) -> dict[str, str] | None:
         """Determine extra build parameters for a specific package."""
-        if not package.version or re.match(r"^1(?:\..*)?$", package.version):
-            return None
-        if "debug" in package.name or package.arch in {"src", "nosrc"}:
+        if self._is_placeholder(package) or self._is_initial_version(package) or self._is_debug_asset(package):
```
nicksinger: For the fun of it:

```suggestion
        if any(getattr(self, f"_is_{check}")(package) for check in ["placeholder", "initial_version", "debug_asset"]):
```
(I don't expect you to actually take this suggestion, it's mad)

nicksinger (changes requested):
Most of the stuff is not really blocking and more of my style. However, the construction of the `BUILD` worries me as it would currently break if any non-string is used (which I assume e.g. the `kernel_version` could be?)

openqabot/incrementapprover.py
asmorodskyi, line 246: why do we need to support case when `kernel_version` didn't match ? I would say this is sign of the problem and we need to fail early

openqabot/incrementapprover.py
```diff
@@ -270,22 +269,36 @@ def _match_package_name_and_version(self, package: Package, additional_build: di
             return None
         return match
 
+    @staticmethod
+    def _is_initial_version(package: Package) -> bool:
```
asmorodskyi: this needs to go to Package class

openqabot/incrementapprover.py
```diff
@@ -270,22 +269,36 @@ def _match_package_name_and_version(self, package: Package, additional_build: di
...
+        return bool(package.version and re.match(r"^1(?:\..*)?$", package.version))
+
+    @staticmethod
+    def _is_placeholder(package: Package) -> bool:
```
asmorodskyi: this needs to go to Package class

openqabot/incrementapprover.py
```diff
@@ -270,22 +269,36 @@ def _match_package_name_and_version(self, package: Package, additional_build: di
...
+        return not package.version
+
+    @staticmethod
+    def _is_debug_asset(package: Package) -> bool:
```
asmorodskyi: this needs to go to Package class

openqabot/incrementapprover.py
asmorodskyi: if this methods not used anywhere else maybe it make sense to create SINGLE method of class Package for that check ?

asmorodskyi (changes requested):

openqabot/incrementapprover.py
okurz: thought about doing it that way too, but it turns out that `re.Match.groupdict()` behaves a bit weirdly with unmatched optional groups. 

If the user provides a regex with a named group like `(?P<kernel_version>\d+_)` but that group doesn't match anything in the package string, `groupdict()` still includes the key in the dictionary but assigns `None` rather than omitting the key. Then, because the key exists in the dict, `groups.get("kernel_version", "")` will return `None` (ignoring the default `""`). Calling `.replace("_", ".")` on `None` then throws an `AttributeError`.

Using `(groups.get("kernel_version") or "")` correctly catches the `None` value and creates an empty string before attempting the replacement, so it's safer for unmatched groups

openqabot/incrementapprover.py
okurz: done

openqabot/incrementapprover.py
okurz: done

openqabot/incrementapprover.py
okurz: done

openqabot/incrementapprover.py
okurz: The main reason for keeping them as separate properties on `Package` is to preserve the explicit domain language explaining exactly *why* a package gets filtered out. 

Instead of hiding the checks inside a single black-box method, having descriptive, explicit properties like `is_initial_version`, `is_placeholder`, and `is_debug_asset` makes the exclusion criteria self-documenting and easier for humans to reason about or change in the future in case there is a need.

openqabot/incrementapprover.py
okurz: The regexes defined in `additional_builds` for user configuration are permitted to define `kernel_version` as an optional matching group (e.g. `(?P<kernel_version>\d+_)?`).

Because not all packages under additional builds will or should match a kernel version, having it evaluate to `None` and safely falling back to an empty string (`""`) is the intended behavior. Failing early on a mismatch would break eligibility checks for any package that matches the build rules but doesn't specify/need a kernel version.

openqabot/incrementapprover.py
okurz: done

openqabot/incrementapprover.py
okurz: done

openqabot/incrementapprover.py
okurz: It's definitely mad and would give any sane static type checker an immediate panic attack. 😂

But jokes aside, based on other review feedback, I've actually moved these checks entirely out of `IncrementApprover` and onto the `Package` class as properties anyway (`package.is_placeholder`, etc.). So the `self._is_*` methods are gone now anyway

openqabot/incrementapprover.py
```diff
@@ -239,25 +239,24 @@ def _match_additional_build(
         if (archs := additional_build.get("archs")) and package.arch not in archs:
             return None
 
-        package_name_match = self._match_package_name_and_version(package, additional_build)
-        if package_name_match is None:
+        if not (match := self._match_package_name_and_version(package, additional_build)):
```
asmorodskyi: why you decided to rename `package_name_match` to abstract `match` ? I find older name was giving more context when you reading the code

asmorodskyi (approved):

openqabot/incrementapprover.py
asmorodskyi: ok

openqabot/incrementapprover.py
asmorodskyi: ok

openqabot/incrementapprover.py
asmorodskyi: none blocking

okurz: renamed back

openqabot/incrementapprover.py
Martchus: Makes sense

Martchus (approved):
The new code is harder to read but I'm not objecting. We're down this road far enough so it doesn't matter anymore anyway.

nicksinger (approved):

PR #516, opened 2026-04-29
feat: improve increment approval disapproval reason output

okurz:
Motivation:
The previous output of `increment-approve` was repetitive and lacked
sufficient detail about failed openQA jobs. It showed multiple identical
headers and plain URLs without context.

Design Choices:
- Grouped disapproval reasons by result across all builds for a request.
- Enriched job information with full name and job group name.
- Simplified the individual job line to show only the URL, name, and group.

Example output:
    The following openQA jobs ended up with result 'failed':
 - https://openqa.suse.de/tests/22061185 (sle-16.0-Azure-...) \
in group 'Public Cloud Azure'

Benefits:
Users now get a concise, grouped list of failing tests with meaningful
names and group context, making it easier to identify blockers.

Related issue: https://progress.opensuse.org/issues/201990

### DIFF
--- a/openqabot/incrementapprover.py
+++ b/openqabot/incrementapprover.py
@@ -10,8 +10,9 @@ from collections import defaultdict
 from concurrent.futures import ThreadPoolExecutor
 from enum import Enum, auto
 from functools import lru_cache
-from itertools import chain
+from itertools import chain, groupby
 from logging import getLogger
+from operator import itemgetter
 from pprint import pformat
 from typing import TYPE_CHECKING, Any
 
@@ -20,7 +21,7 @@ import osc.core
 
 from openqabot import config
 from openqabot.config import OBSOLETE_PARAMS
-from openqabot.openqa import OpenQAInterface
+from openqabot.openqa import ENRICH_KEYS, OpenQAInterface
 from openqabot.pc_helper import apply_public_cloud_settings
 
 from .commenter import Commenter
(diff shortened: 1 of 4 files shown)


openqabot/incrementapprover.py
```diff
@@ -179,15 +182,26 @@ def evaluate_list_of_openqa_job_results(
...
-        openqa_url = self.client.url.geturl()
         for results in list_of_results:
             self.evaluate_openqa_job_results(results, ok_jobs, not_ok_jobs, jobs, request)
-        reasons_to_disapprove = [
-            f"The following openQA jobs ended up with result '{result}':\n"
-            + "\n".join(f" - {openqa_url}/tests/{i}" for i in job_ids)
-            for result, job_ids in not_ok_jobs.items()
+        return (ok_jobs, [], jobs)
+
+    def _generate_job_reasons(self, jobs: list[dict[str, Any]]) -> list[str]:
```
asmorodskyi, line 194: suggestion(blocking) : looks like a method for openQA class

asmorodskyi (changes requested):

openqabot/openqa.py
```diff
@@ -184,12 +184,16 @@ def enrich_job_info(info: dict[str, Any], job_map: dict[int, dict[str, Any]]) ->
...
             return info
 
+        settings = job.get("settings", {})
         return info | {
             "group": job.get("group"),
             "group_id": job.get("group_id"),
-            "distri": job.get("distri"),
-            "version": job.get("version"),
-            "build": job.get("build"),
+            "distri": job.get("distri") or settings.get("DISTRI"),
```
asmorodskyi: question(blocking): can you elaborate why we need this change ? looks like workaround for actual problem in some other place

openqabot/incrementapprover.py
okurz: Not really. I would prefer to keep _generate_job_reasons within the approval/presentation layer (like IncrementApprover or a dedicated commenter helper) rather than moving it to the low-level OpenQAInterface client. OpenQAInterface is designed as a data-access layer wrapper to communicate with openQA APIs, whereas this method generates human-readable text messages specifically formatted for OBS request rejection reasons. Keeping them separated adheres to the Single Responsibility Principle and avoids coupling our API wrapper to specific OBS-approval workflows.

openqabot/openqa.py
okurz: In openQA keys like distri, version, build, and arch are extracted when we schedule, but can be null under various scenarios (such as manually triggered or cloned jobs). that's why we should add this fallback.

openqabot/incrementapprover.py
asmorodskyi: > Keeping them separated adheres to the Single Responsibility Principle

funny that I have same model in my head but opposite . I find this messay that logic related to "openQA jobs management" (getting jobs from openQA and normalize them in different way depending on certain use-case) is spreaded across different places  . I have a fear that this way we duplicating a lot of similar flows but with very tiny differences . First step to improve this would be gathering all logic related to normalizing data from openQA in single place and THIS would follow SIngle Responsibility Principle

openqabot/openqa.py
asmorodskyi: so you saying that if we get the job which does not have VERSION, DISTRI, ARCH and BUILD we would blindly use it to judge about certain submission ? How can we be sure that it's execution related to what we trying to analyse ?

Martchus (approved):

openqabot/openqa.py
okurz: Nevermind, changed the code to not fall back to those secondary settings

asmorodskyi: if I am not mistaken commit message and PR description was not updated with this change

openqabot/incrementapprover.py
asmorodskyi: when time will allow I actually plan to create a PR where I would like to collect such methods/functions which currently spread across different classes and move them in openQA class or/and maybe to some different class but still idea is to collect them all in one place . So we need to find common ground here or we would have similar discussion in different PR after some time :)

openqabot/openqa.py
okurz: done

okurz:
rebased

asmorodskyi (approved):

PR #517, opened 2026-04-29
build: pin dev dependencies and configure dependabot

asmorodskyi:
- Pin all development dependencies to exact versions in pyproject.toml to
  ensure reproducible CI environments.
- Add urllib3 to main dependencies to ensure availability.
- Add .github/dependabot.yml to automate and group development
   dependency updates weekly.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,29 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+    open-pull-requests-limit: 10
+    groups:
+      dev-dependencies:
+        patterns:
+          - "*"
+        update-types:
+          - "major"
+
+    labels:
+      - "dependencies"
+      - "python"
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+    open-pull-requests-limit: 5
+    labels:
+      - "dependencies"
+      - "github-actions"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ name = "qem-bot"
 dynamic = ["version"]
 description = "Tool for scheduling maintenance tests on openQA based on [SMELT](https://tools.io.suse.de/smelt) incidents and Gitea PRs."
 readme = "Readme.md"
-requires-python = ">=3.11"
+requires-python = "~=3.13.0"
 license = "MIT"
 authors = [
     { name = "SUSE LLC" },
@@ -26,6 +26,7 @@ dependencies = [
     "pydantic-settings",
     "python-dotenv",
     "typer",
+    "urllib3",
 ]
 
 [project.urls]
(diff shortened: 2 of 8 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,34 @@
...
+          - "ruff"
+          - "vulture"
+          - "ty"
+          - "pre-commit"
+          - "gitlint"
+          - "pyupgrade"
+    labels:
+      - "dependencies"
+      - "python"
+
```
okurz: replicating the list of dependencies in multiple places is a recipe for failure.

okurz (changes requested):

.github/dependabot.yml
asmorodskyi: agree , please check latest version

okurz:
Looks nice but how do we know that this works? Can we somehow check the validity of the dependabot config?

asmorodskyi:
> Looks nice but how do we know that this works? Can we somehow check the validity of the dependabot config?

we can yolo it in production :) in worst case we will lock tools for some reasonable time . let's merge it and create reminder in a month to see if dependabot is working . if during this month dependabot will not create any PRs I promise to update tools manually and revisit dependabot.yml

okurz:
> > Looks nice but how do we know that this works? Can we somehow check the validity of the dependabot config?
> 
> we can yolo it in production :) in worst case we will lock tools for some reasonable time . let's merge it and create reminder in a month to see if dependabot is working . if during this month dependabot will not create any PRs I promise to update tools manually and revisit dependabot.yml

then please create a ticket which should act as that reminder

asmorodskyi:
> then please create a ticket which should act as that reminder

https://progress.opensuse.org/issues/200564

okurz (approved):

Wabri (approved):

pyproject.toml
```diff
@@ -3,7 +3,7 @@ name = "qem-bot"
 dynamic = ["version"]
 description = "Tool for scheduling maintenance tests on openQA based on [SMELT](https://tools.io.suse.de/smelt) incidents and Gitea PRs."
 readme = "Readme.md"
-requires-python = ">=3.11"
+requires-python = "~=3.13.0"
```
okurz, line 6:
```
requires-python = "~=3.13.0"
```

that's incorrect. We don't "require" this exact version. We should state the supported range and so far I see no need to drop support for 3.11. However, you can specify 3.13 as higher limit. I would still not restrict on the patch level as those should not impact how we need to write code.

okurz (changes requested):
Now I am confused. Your PR says "pin dev dependencies" but your two commits say

* fix: lock python version to 3.13
* fix: update tests to align with latest library updates

It looks like by mistake you squashed commits together? I suggest you split independant changes again

asmorodskyi:
> Now I am confused. Your PR says "pin dev dependencies" but your two commits say
> 
> * fix: lock python version to 3.13
> * fix: update tests to align with latest library updates
> 
> It looks like by mistake you squashed commits together? I suggest

please finish sentence

pyproject.toml
asmorodskyi: > We should state the supported range

why ? we speaking about script which is running in SINGLE place in a world with certain version . why would we care about other versions ?

> I would still not restrict on the patch level

and I am NOT restricting on patch level . there is `~` which makes it flexible for minor updates . I just tried to delete `.venv` and do `uv sync; uv run make test` and it was executed with my local `3.13.13`

okurz:
> > Now I am confused. Your PR says "pin dev dependencies" but your two commits say
> > 
> > * fix: lock python version to 3.13
> > * fix: update tests to align with latest library updates
> > 
> > It looks like by mistake you squashed commits together? I suggest
> 
> please finish sentence

done. Wrote "I suggest you split independant changes again"

pyproject.toml
okurz: ok, good then

okurz (approved):

PR #520, opened 2026-05-08
fix: ensure captured kernel versions start with a digit

okurz:
Motivation:
Overly broad regexes in metadata can capture non-version strings (like product
names) as kernel versions, leading to invalid openQA job scheduling.

Design Choices:
- Added a safeguard in IncrementApprover._match_additional_build to verify
  that captured kernel_version starts with a digit.
- Returns None immediately if the check fails, effectively skipping the match.

Benefits:
- Prevents scheduling of invalid jobs when configuration patterns are loose.
- Increases robustness of the scheduling logic against external config errors.

### DIFF
--- a/openqabot/incrementapprover.py
+++ b/openqabot/incrementapprover.py
@@ -248,6 +248,8 @@ class IncrementApprover:
             extra_build.append(kind)
 
         if kernel_version := groups.get("kernel_version"):
+            if not kernel_version[0].isdigit():
+                return None
             kernel_version = kernel_version.replace("_", ".")
             extra_build.append(kernel_version)
             extra_params["KERNEL_VERSION"] = kernel_version


asmorodskyi (approved):
LGTM

Martchus (commented):
Ok - although I'm wondering whether we shouldn't just fix the regex instead of  introducing additional logic here.

okurz:
> Ok - although I'm wondering whether we shouldn't just fix the regex instead of introducing additional logic here.

Well, I did. Actually I should call this PR a "feature", not a fix.

Wabri (approved):

mdoucha:
> Ok - although I'm wondering whether we shouldn't just fix the regex instead of introducing additional logic here.

I agree with fixing just the regex. This code change is not needed. If someone wants to have weird package versions, let them.

okurz:
> > Ok - although I'm wondering whether we shouldn't just fix the regex instead of introducing additional logic here.
> 
> I agree with fixing just the regex. This code change is not needed. If someone wants to have weird package versions, let them.

I guess sooner or later someone will break the metadata and it would be on us again to investigate what's wrong. Ok, let's see if I am wrong :)

PR #521, opened 2026-05-11
feat: Add ability to trigger opensuse PullRequests

asmorodskyi:
related ticket : https://progress.opensuse.org/issues/189927

### DIFF
--- a/openqabot/args.py
+++ b/openqabot/args.py
@@ -389,13 +389,19 @@ def gitea_sync(  # noqa: PLR0913
 def gitea_trigger(  # noqa: PLR0913
     ctx: typer.Context,
     *,
-    gitea_project: gitea_project_arg = "products/SLFO",
     pr_label: Annotated[
         str,
         typer.Option("--pr-label", envvar="PR_LABEL", help="Gitea PRs label for which to trigger tests"),
     ] = "staging/In Progress",
     pr_number: pr_number_arg = None,
     comment: comment_option = True,
+    maintenance: Annotated[
+        bool,
+        typer.Option(
+            "--maintenance",
+            help="Controls type of pull requests processed",
+        ),
+    ] = False,
     enable_detailed_comments: enable_detailed_comments_option = None,
     fallback_contact: fallback_contact_option = None,
     generic_tool_issues_contact: generic_tool_issues_contact_option = None,
@@ -403,10 +409,10 @@ def gitea_trigger(  # noqa: PLR0913
 ) -> None:
     """Trigger testing for PR(s) with certain label."""
     args = ctx.obj
-    args.gitea_project = gitea_project
     args.pr_number = pr_number
     args.pr_label = pr_label
     args.comment = comment
+    args.maintenance = maintenance
 
     _apply_detailed_comment_options(
         args,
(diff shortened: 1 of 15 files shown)


openqabot/loader/gitea.py
```diff
@@ -626,6 +635,7 @@ def generate_repo_url(pullrequest: PullRequest, staging_config_qa_labels: dict[s
         NoRepoFoundError: raised when required_labels has more than one match in staging.config
 
     """
+    # TODO: revisit it because not we have repo_template as input parameter
     common_labels = set(staging_config_qa_labels.keys()) & pullrequest.labels
```
kalikiana: Why do we care about labels here? We already have exactly 1 PR here. There is nothing to filter, and openSUSE doesn't even use labels. This is unnecessarily SUSE-specific.

kalikiana: As discussed in the Unblock, best case we use `repo_template` for SUSE **and** openSUSE. Meaning only have a **single function** `generate_repo_url` for **both IBS and OBS**. So we can avoid special-casing.

openqabot/args.py
```diff
@@ -389,24 +389,30 @@ def gitea_sync(  # noqa: PLR0913
...
     ctx: typer.Context,
     *,
-    gitea_project: gitea_project_arg = "products/SLFO",
     pr_label: Annotated[
         str,
         typer.Option("--pr-label", envvar="PR_LABEL", help="Gitea PRs label for which to trigger tests"),
     ] = "staging/In Progress",
     pr_number: pr_number_arg = None,
     comment: comment_option = True,
+    opensuse: Annotated[
```
okurz: As we clarified today in our collab session there was some confusion regarding "giteasync vs. giteatrigger". As it turns out what you are doing is mostly "adding support for maintenance workflows in giteatrigger" so hence I wonder if we can find a better name for this switch and switch based on the workflow name if we need a "global switch" at all.

openqabot/args.py
```diff
@@ -389,24 +389,30 @@ def gitea_sync(  # noqa: PLR0913
...
+        bool,
+        typer.Option(
+            "--o3/--osd",
+            help="Use openSUSE instance (o3) or openQA.suse.de (osd)",
```
okurz: this description is a bit inconsistent. If the actual openQA instance matters at all then I suggest common wording like `openqa.opensuse.org (o3) or openqa.suse.de (osd)` however it looks like you are mostly using this flag for something different than *just* to switch which openQA instance is to be used

okurz (changes requested):

openqabot/args.py
asmorodskyi: fully agree with you . can you suggest something in particular ? `maintenance` ? I am open for anything except `fsdfe23454_wrap` :)

openqabot/args.py
asmorodskyi: yeah I will update it for sure , after we will finalize the name

openqabot/args.py
okurz: yeah `maintenance` is a bit better.

okurz (approved):

asmorodskyi:
added a label to avoid automatic merge . merge of PR should be synchronized with update of metadata project

kalikiana (approved):

jlausuch (approved):

okurz:
pushed re-signed commits

asmorodskyi:
manual merge were done , due to technical issues with @mergify

PR #522, opened 2026-05-12
fix: Use ISO_1_URL and ISO_1 in giteatrigger to load the correct asset

jlausuch:
According to the documentation [1], when giving a ISO variable
from a URL, you need to specify the URL and the ISO name that will
be stored locally and used by the openQA jobs. 
The variable ISO_1_URL need to go together with ISO_1.       

[1] https://open.qa/docs/#_specifying_assets_required_by_a_job

### DIFF
--- a/openqabot/giteatrigger.py
+++ b/openqabot/giteatrigger.py
@@ -112,7 +112,8 @@ class GiteaTrigger:
 
         if self.is_openqa_triggering_needed(matched_iso, trigger_config):
             openqa_settings = {
-                "ISO_URL": f"{repo_url}/{matched_iso_regex.group(0)}",
+                "ISO_1_URL": f"{repo_url}/{matched_iso_regex.group(0)}",
+                "ISO_1": f"{matched_iso_regex.group(0)}",
                 "_GITEA_PR": str(pullrequest.number),
                 "VERSION": matched_iso.version,
                 "FLAVOR": trigger_config.flavor,
--- a/tests/test_giteatrigger.py
+++ b/tests/test_giteatrigger.py
@@ -95,7 +95,8 @@ def test_check_pullrequest_triggers_job(
     assert settings["FLAVOR"] == "Online-Staging"
     assert settings["VERSION"] == "15.5:PR-123"
     assert settings["BUILD"] == "PR-123-1.1:SLES-15.5"
-    assert settings["ISO_URL"] == "http://fake.url//SLES-15.5-Online-x86_64-Build1.1.install.iso"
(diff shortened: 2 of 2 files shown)


asmorodskyi (approved):
LGTM

okurz (changes requested):
Please provide more context in the git commit message. The change looks simple enough but given that most tests "just work fine" for very long I think we should at least invest the effort to explain in detail why this change is needed now.

jlausuch:
> Please provide more context in the git commit message. The change looks simple enough but given that most tests "just work fine" for very long I think we should at least invest the effort to explain in detail why this change is needed now.

done

okurz (changes requested):
This change really shouldn't be necessary . https://open.qa/docs/#specifying-assets-required-by-a-job says

```
_URL Before starting these jobs, try to download these assets […]. For example ISO_1_URL=http://trusted.com/foo.iso would […] download the file foo.iso […] and set ISO_1=foo.iso.
```

so by default the according ISO should already be set.

And

```
If you set both ISO_1 and ISO_1_URL, the file pointed to by ISO_1_URL will be downloaded and renamed to the name set as ISO_1.
```

so that would be necessary only if you want the local file to be called *different*

jlausuch:
> This change really shouldn't be necessary . https://open.qa/docs/#specifying-assets-required-by-a-job says
> 
> ```
> _URL Before starting these jobs, try to download these assets […]. For example ISO_1_URL=http://trusted.com/foo.iso would […] download the file foo.iso […] and set ISO_1=foo.iso.
> ```
> 
> so by default the according ISO should already be set.
> 
> And
> 
> ```
> If you set both ISO_1 and ISO_1_URL, the file pointed to by ISO_1_URL will be downloaded and renamed to the name set as ISO_1.
> ```
> 
> so that would be necessary only if you want the local file to be called _different_

Then the documentation is wrong. See:
- Job with ISO_URL only: https://openqa.suse.de/tests/22278272#step/boot_agama/4 -> K.O.
- Job with ISO_1_URL: https://openqa.suse.de/tests/22278276#step/boot_agama/4 -> K.O.
- Job with ISO_1_URL and ISO_1 : https://openqa.suse.de/tests/22278275  -> O.K.

So, we need both ISO_1_URL and ISO_1 in order to make it work (don't ask me why :) )

okurz (approved):

okurz:
rebased and resigned …

jlausuch:
> rebased and resigned …

thanks!

PR #530, opened 2026-05-13
fix: align content of dependabot and gitlint

asmorodskyi:


kalikiana (approved):

.gitlint
```diff
@@ -1,3 +1,3 @@
 [general]
 contrib=contrib-title-conventional-commits
-ignore=body-min-length,body-is-missing
+ignore=body-min-length,body-is-missing,body-max-line-length
```
okurz: Please do not introduce rules that are inconsistent with our other projects. One can specify more specific rules, e.g. see what we do for `git subrepo` in https://github.com/os-autoinst/openQA/blob/master/external/os-autoinst-common/.gitlint#L10

okurz (changes requested):

.gitlint
asmorodskyi: this files are already inconsistent . also for example https://github.com/openSUSE/qem-dashboard does not have `.gitlint` **at all** so I am not sure which consistency you trying to protect . also keep in mind that many people in our team are keen to remove commit message length limitation **at all** .

CC: @perlpunk , @Martchus

okurz: why not streamline qem-dashboard then? -> https://github.com/openSUSE/qem-dashboard/pull/3299
But regardless, I am against disabling the check for "body-max-line-length" in an inconsistent manner

asmorodskyi: please check new version

okurz (approved):

PR #532, opened 2026-05-13
fix(klp): Support product name and HTTP URLs in reference repos

okurz:
Motivation:
We need to evaluate package diffs using reference repos specified by
product name and using direct HTTP URLs (e.g. for SL-Micro where
PUBLISH is outdated and ToTest differs heavily). The current logic
checks 'flavor' instead of 'product', and mangles explicit URLs by
prepending the OBS host and replacing colons.

Design Choices:
Updated `get_package_diff_from_repo` to first check `build_info.product`
in `reference_repos`, falling back to `build_info.flavor`. Also updated
`make_repodata_url` to bypass string mangling when an absolute HTTP/HTTPS
URL is provided. The tests in `test_incrementapprover_diff.py` and
`test_repodiff.py` were extended to cover these changes.

Benefits:
Allows explicit configuration of product reference repositories via URL,
avoiding mis-scheduling old kernel live patches due to outdated repo
comparisons.

Related issues:
* https://progress.opensuse.org/issues/198728
* https://progress.opensuse.org/issues/190542


Wabri (approved):

Wabri (changes requested):
thought(blocking): the CI is failing due to a `ruff format` error in `tests/test_repodiff.py`

asmorodskyi (dismissed):
if we need support for URLs let's just switch and support URLs ONLY . The current way where we generating URL can be switched to just read it . this would make code **simpler**

okurz:
> if we need support for URLs let's just switch and support URLs ONLY . The current way where we generating URL can be switched to just read it . this would make code **simpler**

ok, switched to that

openqabot/loader/incrementconfig.py
asmorodskyi: I think we can delete this one now

openqabot/incrementapprover.py
```diff
@@ -327,45 +327,58 @@ def get_package_diff_from_repo(
...
-                if config_inc.build_repo_template
-                else f"{build_project}/{channel}/{build_info.arch}"
-            )
-            diff_project = (
-                config_inc.diff_repo_template.format(**(params | {"base": diff_project}))
-                if config_inc.diff_repo_template
-                else f"{diff_project}/{build_info.version}/{config_inc.diff_project_suffix}/{build_info.arch}"
+        diff_url = config_inc.diff_project_url()
+        if build_info:
+            ref_repo = config_inc.reference_repos.get(build_info.product) or config_inc.reference_repos.get(
```
asmorodskyi, line 336: can you elaborate what happen here ?

openqabot/incrementapprover.py
```diff
@@ -327,45 +327,58 @@ def get_package_diff_from_repo(
...
+                    if config_inc.build_repo_template
+                    else f"{build_url}/{channel}/{build_info.arch}"
+                )
+                diff_url = (
+                    config_inc.diff_repo_template.format(**(params | {"base": diff_url}))
+                    if config_inc.diff_repo_template
+                    else f"{diff_url}/{build_info.version}/{config_inc.diff_project_suffix}/{build_info.arch}"
+                )
+
+        def _to_url(path: str) -> str:
```
asmorodskyi: why do we need this ? I thought we would really switch to ONLY urls

openqabot/loader/incrementconfig.py
okurz: done

openqabot/incrementapprover.py
okurz: This allows us to specify reference repository overrides by the specific product name (e.g., SL-Micro) for higher precision, while still falling back to the flavor (e.g., Online-Increments) for the general case. I've added a clarifying comment in the code to make this intent explicit.

openqabot/incrementapprover.py
okurz: I've removed the local `_to_url` helper and centralized the logic into a new `IncrementConfig.to_url` static method. This ensures we convert OBS project paths to URLs consistently and as early as possible. If you mean you want to switch qem-bot completely to only take full URLs already in the metadata configuration: We could remove the conversion logic entirely, but that would mean we'd have to update all our configuration YAMLs to use full HTTP URLs instead of OBS project paths. 
While that makes the Python code simpler, it makes the configuration much more verbose and harder to maintain for humans (e.g., replacing diff_project_suffix: PUBLISH/product with a full URL for every entry). 
IMHO using `to_url` is balanced: we keep the concise OBS-path syntax for the 95% of standard cases, while allowing explicit URL overrides for the 5% of special cases like SL-Micro product increments.

Wabri (changes requested):
I think is ok, but need a rebase first

openqabot/repodiff.py
```diff
@@ -102,39 +100,39 @@ def request_and_dump(
...
-            url,
-            f"repodata-listing-{project}.json",
+            repodata_url,
+            f"repodata-listing-{url.replace('/', '_').replace(':', '_')}.json",
             as_json=True,
             params=get_obs_filter_params(r".*-primary\.xml.*"),
         )
         if not repo_data_listing or not isinstance(repo_data_listing, dict):
-            log.error("Could not load repo data for project %s", project)
+            log.error("Could not load repo data for project %s", url)
```
r-richardson: Maybe
```suggestion
            log.error("Could not load repo data for URL %s", url)
```
Just a suggestion, same would apply for direct mentions of `project` further below

okurz:
> I think is ok, but need a rebase first

why do you need a review first to review the code? Anyway, rebased

openqabot/repodiff.py
okurz: ok, done.

r-richardson (approved):

openqabot/incrementapprover.py
```diff
@@ -327,45 +327,51 @@ def get_package_diff_from_repo(
...
+                    "suffix": config_inc.diff_project_suffix,
+                    "product": build_info.product,
+                }
+                build_url = (
+                    config_inc.to_url(config_inc.build_repo_template.format(**(params | {"base": build_url})))
+                    if config_inc.build_repo_template
+                    else f"{build_url}/{channel}/{build_info.arch}"
+                )
+                diff_url = (
+                    config_inc.to_url(config_inc.diff_repo_template.format(**(params | {"base": diff_url})))
```
asmorodskyi: currently actual value for this variable is `"{base}/{version}/{arch}/product"`  , how from code I can understand which variables are actually required ? what would happen if user will not specify one of them ?

okurz: Right. We can improve. Added an additional commit on top:
* b7bb184

Martchus (approved):

Wabri (approved):

PR #533, opened 2026-05-17
feat: Report submission rejection reason to dashboard

okurz:
Motivation
Previously, qem-bot would log rejection reasons locally, but they were
not visible on the dashboard. This made it difficult for users to understand
why a submission wasn't being approved.

Design Choices
Implemented `update_incident_reason` in the QEM loader to send PATCH requests
to the new dashboard endpoint.
Updated the Approver to call `update_incident_reason` with the formatted
rejection reason when a submission is not approvable.
Ensured the rejection reason is cleared (set to None) when a submission
becomes fully approvable.

Benefits
The dashboard now reflects the actual state of the bot's evaluation,
providing clear feedback to users about missing tests or job failures.

Needs:
* https://github.com/openSUSE/qem-dashboard/pull/3300

### DIFF
--- a/openqabot/loader/qem.py
+++ b/openqabot/loader/qem.py
@@ -318,3 +318,21 @@ def update_job(job_id: int, data: dict[str, Any]) -> None:
 
     except requests.exceptions.RequestException:
         log.exception("QEM Dashboard API request failed")
+
+
+def update_incident_reason(incident_number: int, reason: str | None) -> None:
+    """Update the rejection reason for a submission on the dashboard."""
+    try:
+        result = dashboard.patch(
+            f"api/incidents/{incident_number}/rejection_reason",
+            headers=config_module.settings.dashboard_token_dict,
+            json={"rejection_reason": reason},
+        )
+        if result.status_code != HTTPStatus.OK:
+            log.error(
+                "Dashboard API error: Could not update rejection reason for incident %s: %s",
+                incident_number,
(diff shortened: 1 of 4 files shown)


openqabot/loader/qem.py
```diff
@@ -318,3 +318,21 @@ def update_job(job_id: int, data: dict[str, Any]) -> None:
...
         log.exception("QEM Dashboard API request failed")
+
+
+def update_incident_reason(incident_number: int, reason: str | None) -> None:
+    """Update the rejection reason for a submission on the dashboard."""
+    try:
+        result = dashboard.patch(
+            f"api/incidents/{incident_number}/rejection_reason",
+            headers=config_module.settings.dashboard_token_dict,
+            json={"rejection_reason": reason},
```
asmorodskyi, line 329: does it actually make sense to sent {"rejection_reason": None} ? keeping in mind that we doing `patch` call we might overwrite valid reason previously defined

asmorodskyi (changes requested):

openqabot/loader/qem.py
kalikiana: The way I read `openqabot/approver.py` that only happens **after** checking all issues have been **ruled out**.

kalikiana (approved):

asmorodskyi (approved):
LGTM

PR #536, opened 2026-05-19
fix: remove python label for dependabot

asmorodskyi:
Since we are running 100% python package this label
does not make much sense

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,7 +16,6 @@ updates:
 
     labels:
       - "dependencies"
-      - "python"
 
   - package-ecosystem: "github-actions"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -16,7 +16,6 @@ updates:
 
     labels:
       - "dependencies"
-      - "python"
```
okurz, line 19: your commit message says

> Since we are running 100% python package this label does not make much sense

but that's not true. Just some lines below we have "dependabot". Why not just add this label? Or we remove it from "dependabot" as well and don't distinguish by label

okurz (changes requested):

PR #537, opened 2026-05-21
feat: fail explicitly when no configs were loaded

asmorodskyi:
if we point qem-bot to folder with some yamls but
there will be no yamls with trigger_config section
we will end up with fake succesful run. It is more
clear to just fail in that case.


okurz (approved):
tested locally with an older version of metadata fails as expected, with most recent passes.

Wabri (approved):

PR #546, opened 2026-06-01
fix(dependabot): extend ignore regex to cover different cases

asmorodskyi:


okurz (changes requested):
duplicate of https://github.com/openSUSE/qem-bot/pull/545

asmorodskyi:
> duplicate of #545

yes and I would suggest to close #545 unless you would overtake my change explicitly . #545 would not cover #524

okurz (approved):

okurz:
> > duplicate of #545
> 
> yes and I would suggest to close #545 unless you would overtake my change explicitly . #545 would not cover #524

We also need https://github.com/openSUSE/qem-bot/pull/547 as you broke mergify rules in 3086bd0f

okurz:
* #547

is merged. Please rebase

PR #547, opened 2026-06-01
fix(mergify): correct rules for target branch

okurz:
Fixes a regression introduced in 3086bd0f which wrongly copied a rule from
qem-dashboard which has a different default target branch.


asmorodskyi (approved):
LGTM

PR #548, opened 2026-06-02
ci(mergify): fix sporadically missing codecov checks blocking merge

okurz:
As full statement+branch coverage is ensured by "make test" already we do
not need to wait for codecov checks. This also helps in cases when codecov
checks are missing or only appearing very late.


asmorodskyi (approved):
LGTM

Wabri (approved):

PR #550, opened 2026-06-03
ci(mergify): update dependabot branches natively

asmorodskyi:
Dependabot ignores comments from mergify so we need
to update branches instead

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -30,10 +30,12 @@ pull_request_rules:
     actions: *merge
   - name: automatic merge for dependabot updates
     conditions:
-      - and: *base_checks
-      - and:
-          - author=dependabot[bot]
-          - "label=waited"
+      - base=master
+      - -label~=^acceptance-tests-needed|not-ready
+      - "#check-failure=0"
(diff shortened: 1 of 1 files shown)


asmorodskyi:
requirement for linear history does not make sense here because we doing squash merge anyway

.mergify.yml
```diff
@@ -30,10 +30,12 @@ pull_request_rules:
...
-      - and: *base_checks
-      - and:
-          - author=dependabot[bot]
-          - "label=waited"
+      - base=master
+      - -label~=^acceptance-tests-needed|not-ready
+      - "#check-failure=0"
+      - "#check-pending=0"
+      - author=dependabot[bot]
+      - "label=waited"
```
okurz, line 38: why do you want to introduce this duplication? I don't understand your reasoning regarding removing the "linear-history". We want a linear history for all commits, regardless if those are dependabot created or any other author

okurz (changes requested):

asmorodskyi:
after live discussion we agreed to close it

PR #551, opened 2026-06-05
Add a option for scheduling image tests in giteatrigger

jlausuch:
Some tests are not booting an ISO, but an HDD instead. By default, it will post the ISO_1, but if the metadata configuration contains image_regex, it will match that regex and use HDD_1 instead.

### DIFF
--- a/openqabot/giteatrigger.py
+++ b/openqabot/giteatrigger.py
@@ -115,8 +115,6 @@ class GiteaTrigger:
 
         if self.is_openqa_triggering_needed(matched_iso, trigger_config):
             openqa_settings = {
-                "ISO_1_URL": f"{repo_url}/{matched_iso_regex.group(0)}",
-                "ISO_1": f"{matched_iso_regex.group(0)}",
                 "_GITEA_PR": str(pullrequest.number),
                 "VERSION": matched_iso.version,
                 "FLAVOR": trigger_config.flavor,
@@ -124,6 +122,29 @@ class GiteaTrigger:
                 "DISTRI": trigger_config.distri,
                 "BUILD": matched_iso.build,
             }
+
+            # Add HDD parameters if image_regex is configured, otherwise use ISO
+            if trigger_config.image_regex:
+                # Images are in /images/ directory instead of /product/iso/
+                images_url = repo_url.replace("/product/iso", "/images")
+                matched_image_regex = Crawler(verify=not config.settings.insecure).get_regex_match_from_url(
+                    images_url, trigger_config.image_regex
+                )
+                if matched_image_regex:
+                    openqa_settings["HDD_1_URL"] = f"{images_url}/{matched_image_regex.group(0)}"
+                    openqa_settings["HDD_1"] = matched_image_regex.group(0)
+                else:
+                    log.warning(
+                        "No image found matching regex '%s' for PR %s in %s",
+                        trigger_config.image_regex,
+                        pullrequest.number,
+                        images_url,
+                    )
+            else:
+                # No image_regex configured, use ISO parameters
+                openqa_settings["ISO_1_URL"] = f"{repo_url}/{matched_iso_regex.group(0)}"
+                openqa_settings["ISO_1"] = matched_iso_regex.group(0)
+
             self.openqa.post_job(openqa_settings)
             log.info("Triggered openQA tests for PR %s on %s", pullrequest.number, matched_iso.arch)
         else:
--- a/openqabot/loader/triggerconfig.py
+++ b/openqabot/loader/triggerconfig.py
@@ -16,6 +16,7 @@ class TriggerConfig:
     distri: str
     flavor: str
     settings: dict[str, str] = field(default_factory=dict)
+    image_regex: str | None = None
 
     def __str__(self) -> str:
         """Return a string representation of the increment configuration."""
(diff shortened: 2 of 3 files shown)


jlausuch:
@asmorodskyi @okurz please review when you have time.

openqabot/loader/triggerconfig.py
```diff
@@ -29,4 +30,5 @@ def from_config_entry(cls, entry: dict[str, Any]) -> TriggerConfig:
             distri=entry["distri"],
             flavor=entry.get("flavor", "any"),
             settings=entry.get("settings", {}),
+            image_regex=entry.get("image_regex"),
```
asmorodskyi:
```suggestion
            image_regex=entry.get("image_regex", ""),
```

openqabot/giteatrigger.py
```diff
@@ -115,15 +115,36 @@ def check_pullrequest(self, pullrequest: PullRequest, trigger_config: TriggerCon
...
                 "_GITEA_PR": str(pullrequest.number),
                 "VERSION": matched_iso.version,
                 "FLAVOR": trigger_config.flavor,
                 "ARCH": matched_iso.arch,
                 "DISTRI": trigger_config.distri,
                 "BUILD": matched_iso.build,
             }
+
+            # Add HDD parameters if image_regex is configured, otherwise use ISO
+            if hasattr(trigger_config, "image_regex") and trigger_config.image_regex:
```
asmorodskyi: if image_regex will have default value empty string we can simplify check here 
```suggestion
            if trigger_config.image_regex:
```

asmorodskyi (changes requested):
we switching from iso to hdd as our base for tests but we still calculating VERSION and ARCH based on iso found . yeah maybe they suppose to be the same but especially in case of ARCH I can imagine cases when you x86 ISO and only arch64 image 🤔

jlausuch:
> we switching from iso to hdd as our base for tests but we still calculating VERSION and ARCH based on iso found . yeah maybe they suppose to be the same but especially in case of ARCH I can imagine cases when you x86 ISO and only arch64 image 🤔

Yes, I want to keep the same build/version as the ISO jobs, in fact the ISO build number determines the product build. We don't care about the image build, those go different numbering. Thanks!

asmorodskyi:
> > we switching from iso to hdd as our base for tests but we still calculating VERSION and ARCH based on iso found . yeah maybe they suppose to be the same but especially in case of ARCH I can imagine cases when you x86 ISO and only arch64 image 🤔
> 
> Yes, I want to keep the same build/version as the ISO jobs, in fact the ISO build number determines the product build. We don't care about the image build, those go different numbering. Thanks!

ok I see and what about ARCH ? any chance that architectures available for images won't match architectures available for isos ?

jlausuch:
> > > we switching from iso to hdd as our base for tests but we still calculating VERSION and ARCH based on iso found . yeah maybe they suppose to be the same but especially in case of ARCH I can imagine cases when you x86 ISO and only arch64 image 🤔
> > 
> > 
> > Yes, I want to keep the same build/version as the ISO jobs, in fact the ISO build number determines the product build. We don't care about the image build, those go different numbering. Thanks!
> 
> ok I see and what about ARCH ? any chance that architectures available for images won't match architectures available for isos ?

That's a good point. Currently the 16.1 stagings are targetting x86_64 only, actually there is no aarch64 repo built, however there is aarch64 images... I'd like to trigger only x86_64 jobs... maybe the regex in the metadata should contain x86_64 as part of the string match...

okurz (approved):

openqabot/loader/triggerconfig.py
```diff
@@ -16,6 +16,7 @@ class TriggerConfig:
     distri: str
     flavor: str
     settings: dict[str, str] = field(default_factory=dict)
+    image_regex: str | None = None
```
asmorodskyi:
```suggestion
    image_regex: str 
```

asmorodskyi (approved):
LGTM

openqabot/loader/triggerconfig.py
jlausuch: I can't remove this, as UT fail. `image_regex` is optional.

openqabot/giteatrigger.py
```diff
@@ -115,15 +115,36 @@ def check_pullrequest(self, pullrequest: PullRequest, trigger_config: TriggerCon
...
+            if trigger_config.image_regex:
+                # Images are in /images/ directory instead of /product/iso/
+                images_url = repo_url.replace("/product/iso", "/images")
+                matched_image_regex = Crawler(verify=not config.settings.insecure).get_regex_match_from_url(
+                    images_url, trigger_config.image_regex
+                )
+                if matched_image_regex:
+                    openqa_settings["HDD_1_URL"] = f"{images_url}/{matched_image_regex.group(0)}"
+                    openqa_settings["HDD_1"] = matched_image_regex.group(0)
+                else:
```
asmorodskyi, line 136: I wonder about this else . shouldn't we raise exception or at least prevent triggering somehow ? In currently merged version we basically end up with neither HDD_* vars nor ISO_* vars

jlausuch: Yes, I though of this, but in that case, the test should fail in purpose, because there might be something wrong in the config. It doesn't make sense to fallback to ISO in case of bootable images anyway...

PR #555, opened 2026-06-08
feat(requests): improve OBS URL feedback and allow metadata overrides

okurz:
Motivation:
When OBS_URL is unset, the bot connects to the wrong instance and logs
a confusing "no requests" message. Additionally, we need a way to prevent
evaluating internal SUSE projects against the public openSUSE API.

Design Choices:
Added obs_url to log messages in openqabot/requests.py for better developer
visibility. Extended IncrementConfig to accept an obs_url field and updated
the fetching logic and tests to respect this override.

Benefits:
Provides clearer log feedback when OBS_URL is misconfigured. Allows
explicitly mapping internal projects to the internal IBS server via metadata
to prevent accidental external queries.

### DIFF
--- a/openqabot/commenter.py
+++ b/openqabot/commenter.py
@@ -102,18 +102,25 @@ class Commenter:
         return msg, state
 
     def osc_comment_on_request(
-        self, request_id: str, msg: str, state: str, revisions: dict[str, Any] | None = None
+        self,
+        request_id: str,
+        msg: str,
+        state: str,
+        revisions: dict[str, Any] | None = None,
+        obs_url: str | None = None,
     ) -> None:
         """Comment on an OBS request."""
-        osc.conf.get_config(override_apiurl=config.settings.obs_url)
+        obs_url = obs_url or config.settings.obs_url
+        osc.conf.get_config(override_apiurl=obs_url)
+        commentapi = self.commentapi if obs_url == config.settings.obs_url else CommentAPI(obs_url)
         bot_name = "openqa"
         info: dict[str, Any] = {"state": state}
         if revisions:
             info.update(revisions)
 
         msg = truncate(add_marker(msg, bot_name, info).strip())
-        comments = self.commentapi.get_comments(request_id=request_id)
-        comment, _ = self.commentapi.comment_find(comments, bot_name, info)
+        comments = commentapi.get_comments(request_id=request_id)
+        comment, _ = commentapi.comment_find(comments, bot_name, info)
 
         # To prevent spam, assume same state/result
         # and number of lines in message is a duplicate message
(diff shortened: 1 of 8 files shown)


asmorodskyi:
I would like to kindly ask you to ask AI to generate shorter comments . when I see comment big like this I simply skip it and try to find more effective ways to understand what this PR is about

okurz:
> I would like to kindly ask you to ask AI to generate shorter comments . when I see comment big like this I simply skip it and try to find more effective ways to understand what this PR is about

I made the description more concise

asmorodskyi (changes requested):
I disagree that we need this duplication to also allow another we to pass URL . why not just specify a variable ?

okurz:
> I disagree that we need this duplication to also allow another we to pass URL . why not just specify a variable ?

sorry, I don't understand. What do you mean with "to also allow another we to pass URL"?

asmorodskyi:
> > I disagree that we need this duplication to also allow another we to pass URL . why not just specify a variable ?
> 
> sorry, I don't understand. What do you mean with "to also allow another we to pass URL"?

typo : /we/way/

Martchus (approved):

okurz:
> I disagree that we need this duplication to also allow another we to pass URL . why not just specify a variable ?

Turns out you're right that this needed a closer look — but the actual problem was deeper than any parameter duplication.

I looked at using a single global variable instead (i.e. temporarily overriding `config.settings.obs_url)`, but that doesn't work safely here for two reasons:
1. `get_obs_request_list`/`_find_request_on_obs_cached` are `@cached` by their explicit arguments only. If obs_url isn't part of the cache key, two config groups with the same OBS project name but different obs_url could silently get each other's cached (wrong-instance) results. Right now this won't matter but *might* in the future depending on how our OBS+IBS usage goes.
2. Approval is deferred: `handle_approval()` runs in a separate loop after all config groups have been processed. A temporarily-mutated global would already be gone (or hold a different group's value) by the time approval actually runs.

While checking this, I also found the override was incomplete even in the current implementation: `approve_on_obs` and `Commenter`'s OBS comment path still read `config.settings.obs_url` directly, ignoring any per-config override entirely. So a request on an internal IBS project would be *found* correctly but *approved/commented* against the wrong (public) instance.

Fixed by resolving `obs_url` once per config group and carrying it explicitly on `ApprovalStatus`, so `approve_on_obs`/`osc_comment_on_request` use the same URL that found the request, instead of re-reading the global. This keeps the explicit-parameter approach (needed for cache correctness) but closes the gap you were pointing at (indirectly).

openqabot/commenter.py
```diff
@@ -102,18 +102,25 @@ def get_stat(j: dict[str, Any]) -> str:
...
+        msg: str,
+        state: str,
+        revisions: dict[str, Any] | None = None,
+        obs_url: str | None = None,
     ) -> None:
         """Comment on an OBS request."""
-        osc.conf.get_config(override_apiurl=config.settings.obs_url)
+        obs_url = obs_url or config.settings.obs_url
+        osc.conf.get_config(override_apiurl=obs_url)
+        commentapi = self.commentapi if obs_url == config.settings.obs_url else CommentAPI(obs_url)
```
asmorodskyi, line 115: can you please elaborate this line for me ? I don't understand why during same call of qem-bot obs_url can differ from `config.settings.obs_url`

okurz: In a single run, `config.settings.obs_url` serves as the global default OBS instance. However, since the introduction of metadata overrides (`IncrementConfig.obs_url`), different configurations can target different OBS instances (e.g., public OBS vs. internal IBS). When processing a request, its specific `obs_url` is captured in its ApprovalStatus and forwarded here. This check ensures we construct a `CommentAPI` client targeting the correct, overridden OBS instance rather than the default one. If you prefer we could use strict validation to ensure that only one value is used at a time however that won't take away complexity and would still need everyone SUSE internal needing to configure OBS_URL hence I prefer the current approach.

asmorodskyi (approved):

PR #558, opened 2026-06-08
Revert "ci(mergify): let dependabot rebase its own PRs"

okurz:
This reverts commit 46d1898944c23b3bf2c723965368ecdb1ffffc2a which was
wrongly assuming that dependabot would be asked to rebase its own PRs
which was not the case. Now we see problems as in
https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/25500
where unauthorized mergify asks dependabot to rebase which dependabot
refuses.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -49,23 +49,9 @@ pull_request_rules:
       - and:
           - updated-at<4 days ago
           - author=dependabot[bot]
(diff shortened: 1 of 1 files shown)


Martchus (approved):

asmorodskyi (approved):
LGTM

PR #559, opened 2026-06-09
feat: Prevent jobs for submissions to be treated as consecutive builds

Martchus:
* Append the incident ID or PR number to the `VERSION` when scheduling jobs for submissions
* Note that job templates need to be adjusted making use of https://github.com/os-autoinst/openQA/pull/7489
* Note that tests need to be adjusted, e.g. https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/25684

Related ticket: https://progress.opensuse.org/issues/200991

### DIFF
--- a/openqabot/types/submissions.py
+++ b/openqabot/types/submissions.py
@@ -185,7 +185,9 @@ class Submissions(BaseConf):
             **self.settings,
             "ARCH": arch,
             "FLAVOR": flavor,
-            "VERSION": f"{self.settings['VERSION']}:{sub.type}-{sub.id}",
+            "VERSION": f"{self.settings['VERSION']}:{sub.type}-{sub.id}"
+            if ctx.data.get("versioned_by_submission", False)
+            else self.settings["VERSION"],
             "DISTRI": self.settings["DISTRI"],
             "INCIDENT_ID": sub.id,
             "REPOHASH": revs,
--- a/openqabot/types/submissions.py
+++ b/openqabot/types/submissions.py
@@ -185,7 +185,7 @@ class Submissions(BaseConf):
             **self.settings,
             "ARCH": arch,
             "FLAVOR": flavor,
-            "VERSION": self.settings["VERSION"],
+            "VERSION": f"{self.settings['VERSION']}:{'inc' if sub.type == 'smelt' else 'PR'}-{sub.id}",
             "DISTRI": self.settings["DISTRI"],
             "INCIDENT_ID": sub.id,
             "REPOHASH": revs,
--- a/openqabot/types/submissions.py
+++ b/openqabot/types/submissions.py
@@ -185,7 +185,7 @@ class Submissions(BaseConf):
             **self.settings,
             "ARCH": arch,
             "FLAVOR": flavor,
-            "VERSION": f"{self.settings['VERSION']}:{'inc' if sub.type == 'smelt' else 'PR'}-{sub.id}",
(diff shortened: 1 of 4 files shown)


openqabot/types/submissions.py
```diff
@@ -185,7 +185,7 @@ def get_base_settings(self, ctx: SubContext, revs: int, cfg: SubConfig) -> dict[
             **self.settings,
             "ARCH": arch,
             "FLAVOR": flavor,
-            "VERSION": self.settings["VERSION"],
+            "VERSION": f"{self.settings['VERSION']}:{'inc' if sub.type == 'smelt' else 'PR'}-{sub.id}",
```
okurz: can we instead just prepend with `sub.type` directly?

okurz (changes requested):

openqabot/types/submissions.py
Martchus: I find e.g. `16.0:PR-1234` nicer than `16.0:git-1234` and `15.7:inc-1234` nicer than `15.7:smelt-1234`. I'll change it to use `sub.type` directly for simplicity.

okurz: If you find PR nicer than git then we can consider to change the `__str__` of sub.type

Martchus: I guess that's good enough and changing the `__str__` type might affect other places as well (which we probably don't want).

okurz (approved):

asmorodskyi (approved):
LGTM

openqabot/types/submissions.py
```diff
@@ -185,7 +185,7 @@ def get_base_settings(self, ctx: SubContext, revs: int, cfg: SubConfig) -> dict[
             **self.settings,
             "ARCH": arch,
             "FLAVOR": flavor,
-            "VERSION": self.settings["VERSION"],
+            "VERSION": f"{self.settings['VERSION']}:{sub.type}-{sub.id}",
```
okurz: As discussed this should be configurable with a config section specific feature switch, e.g.


```suggestion
            "VERSION": self.settings.get('USE_GLOBAL_VERSION', False) ? self.settings["VERSION"] : f"{self.settings['VERSION']}:{sub.type}-{sub.id}",
```

and in preparation you can put `USE_GLOBAL_VERSION: 1` in all metadata config sections and then step by step update metadata sections or ask test owners to do that. Alternatively introduce the *new* feature with an optional feature switch, e.g. `VERSIONED_BY_SUBMISSION`, defaulting to `False`, but then we should plan to deprecate that switch soon and also migrate all sections

okurz (changes requested):

openqabot/types/submissions.py
Martchus: I now made it configurable per incident flavor, that's the same level where also relevant archs and e.g. `override_priority` are configured.

Wabri (approved):

asmorodskyi (changes requested):
question(blocking): why do we need support for both cases ?

Martchus:
> question(blocking): why do we need support for both cases ?

Merging this would otherwise be a breaking change and we needed to update all job templates at once. With this we can update job templates step by step. We have discussed this, see https://github.com/openSUSE/qem-bot/pull/559#pullrequestreview-4484467903.

asmorodskyi (approved):

okurz (approved):
why still draft?

PR #561, opened 2026-06-10
test: Make tests work despite `QEM_DASHBOARD_URL` in environment

Martchus:
This variable might be in someones environment as it is useful for development and also mentioned in `.env.example`.


Martchus:
>  If dependencies change, remember to update https://build.suse.de/package/show/QA:Maintenance:tw/qem-bot-container as well.

We don't need that in the container. It is only relevant for development and even then it is optional.

okurz (approved):

gauravpathak (approved):

PR #567, opened 2026-06-15
ci(dependabot): remove updates grouping

asmorodskyi:
Initially idea of grouping was to avoid spamming but it appears that amount of
spam remains the same just concentrated in single PR instead of several.
Also single PR for several updates makes overall dependency management more complicated

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,12 +13,6 @@ updates:
     commit-message:
       prefix: "chore"
       include: scope
-    groups:
-      dev-dependencies:
-        patterns:
-          - "*"
-        update-types:
-          - "major"
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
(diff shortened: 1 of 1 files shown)


okurz (changes requested):
s/concitrated/concentrated/.

One can still unsubscribe from an individual PR.

What do you mean with "spam in single PR" exactly? Can you provide some examples in your PR description?

I think there is a problem of grouping updates but then delaying merging of updates with an additional cooldown in mergify.

asmorodskyi:
> What do you mean with "spam in single PR" exactly? 

whenever new dependency needs to be updated dependabot close previous PR and opens new having all current dependencies + newly added this does not differ much from previous state where in such a case dependabot will create new PR . Because new PR basically created anyway :)  So whole point of grouping ( at least for me ) is neavilated , so in this case I would prefer separate PRs

Martchus:
But won't the grouping still help keep the number of opened PRs down?

asmorodskyi:
> But won't the grouping still help keep the number of opened PRs down?

I didn't have problem with amount of opened PRs because I am not merging them by hand anyway . I had problem with amount of messages which I am getting from project , somehow I thought that grouping would help me but it appears that it does not helping

PR #569, opened 2026-06-15
fix: update ruff to 0.15.17 and fix new RUF076 violations

okurz:
Bump ruff to 0.15.17 and add targeted pytest autouse exclusions.

Obsoletes:
* #565

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -69,13 +69,13 @@ def fake_responses_for_unblocking_submissions_via_openqa_comments(
         )
 
 
-@pytest.fixture(autouse=True)
+@pytest.fixture(autouse=True)  # noqa: RUF076 - pytest autouse required to ensure all tests run with fake openQA URL
 def fake_openqa_url() -> str:
     settings.openqa_instance = "http://instance.qa"
     return settings.openqa_instance
 
 
-@pytest.fixture(autouse=True)
+@pytest.fixture(autouse=True)  # noqa: RUF076 - pytest autouse required to ensure all tests run with fake openQA job stat URL
 def fake_openqa_url_job_stat(fake_openqa_url: str) -> str:
     settings.openqa_instance = "http://instance.qa"
     return f"{fake_openqa_url}/api/v1/isos/job_stats"
@@ -123,7 +123,7 @@ def make_approver(submission: int = 0, *, mocker: MockerFixture | None = None, c
     return instance()
 
 
-@pytest.fixture(autouse=True)
+@pytest.fixture(autouse=True)  # noqa: RUF076 - pytest autouse required to clear cache automatically before every test
 def _auto_clear_cache() -> None:
     clear_cache()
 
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -69,13 +69,13 @@ def fake_responses_for_unblocking_submissions_via_openqa_comments(
         )
 
 
-@pytest.fixture(autouse=True)
+@pytest.fixture(autouse=True)  # noqa: RUF076 - pytest autouse required to ensure all tests run with fake openQA URL
```
asmorodskyi, line 72: it looks like this ignorance should be moved to some ruff config or something like that so it would be just globally excluded

okurz: I tried but failed. I agree that centralizing rule ignores can be cleaner like in this instance. However it looks like ruff strictly validates rule codes in configuration files like pyproject.toml.

Because RUF076 is a preview rule introduced in version 0.15.17, adding it to pyproject.toml causes a hard TOML parsing crash (Unknown rule selector RUF076) for any developer or environment running a slightly older version of Ruff (like 0.15.15 in current openSUSE packages).

To avoid breaking local developer environments, we can either:
1. Keep the inline # noqa: RUF076 comments but add RUF102 (Invalid rule code in noqa) to per-file-ignores in pyproject.toml for tests/* so that older linters gracefully ignore the unrecognized code (recommended)
2. Alternatively, if we want to enforce the newer Ruff configuration globally, we should update our Makefile to invoke linting via uv run ruff instead of the raw ruff command to ensure everyone runs the locked 0.15.17 version (not recommended as that forces everyone to run an upstream "random download from the internet", instead of trusted system packages)

asmorodskyi: > To avoid breaking local developer environments, we can either:

from my personal experience running version of ruff which is different from one which is used in GHA is not supported anyway . if you will run version which is different than one which is use in prod you would end up with bunch of failures which won't reproduce in GHA so this scenario is not usable anyway

asmorodskyi: I am using a container to do a local runs with version exactly the same as in GHA , also recently I have learned that there is special setup allowing you to run GHA locally . so I think we should made our conclusions on the fact that any developer **required** to use same version of ruff as in prod

okurz: > > To avoid breaking local developer environments, we can either:
> 
> from my personal experience running version of ruff which is different from one which is used in GHA is not supported anyway . if you will run version which is different than one which is use in prod you would end up with bunch of failures which won't reproduce in GHA so this scenario is not usable anyway

I am literally adding that support for the change in question

Martchus (approved):

tests/conftest.py
asmorodskyi: sorry I didn't get your last message , can you please rephrase ?  from what I understand main motivation against adding ignore of this rule globally to allow people to use older versions , but I am saying that you can not use older versions anyway because there are a lot of other not backward compatible changes done by ruff anyway so older ruff version would fail in any case

okurz: Older Ruff versions actually check the codebase successfully today. For example, running ruff check using system-packaged Ruff 0.15.15 passes cleanly.

The Python code remains compatible. The only issue is that putting a new rule like RUF076 in pyproject.toml causes a TOML parsing failure on versions before 0.15.17.

By keeping the inline # noqa: RUF076 and ignoring RUF102 (invalid noqa) in pyproject.toml for tests/*, older versions will ignore the unrecognized noqa code, while newer versions will use it to suppress the autouse check. This keeps the codebase compatible with older system packages.

asmorodskyi (approved):

PR #571, opened 2026-06-17
fix: Prevent silent ignoring of failed tests in product increments

okurz:
Motivation:
qem-bot failed to block product increments when a failed test (e.g., job
22862360) was bundled with a development group test in the openQA stats
response. The bot incorrectly assigned the development group metadata to the
entire "failed" bucket.

Design Choices:
Updated `enrich_job_info` to filter out development jobs individually from the
`job_ids` list. Also modified `get_jobs_by_ids` to fetch metadata in chunks to
prevent HTTP 414 URI Too Long errors when querying openQA for all job details.

Benefits:
Ensures legitimate failing tests correctly block product increment approvals
even when they appear alongside ignored development tests in openQA's job stats.
Related issue: https://gitlab.suse.de/qa-maintenance/bot-ng/-/jobs/6919910

### DIFF
--- a/openqabot/incrementapprover.py
+++ b/openqabot/incrementapprover.py
@@ -128,13 +128,12 @@ class IncrementApprover:
         with ThreadPoolExecutor() as executor:
             stats = list(executor.map(fetch_stats, params))
 
-        # Fetch all relevant job details in a single API call to avoid N+1 query problem
         job_ids = [
-            int(ids[0])
+            int(i)
             for stat in stats
             for jobs in stat.values()
             for info in jobs.values()
-            if (ids := info.get("job_ids"))
+            for i in info.get("job_ids", [])
         ]
         job_map = {job["id"]: job for job in self.client.get_jobs_by_ids(job_ids)}
 
--- a/openqabot/openqa.py
+++ b/openqabot/openqa.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 import logging
 from functools import lru_cache
 from http import HTTPStatus
+from itertools import batched
 from pprint import pformat
 from typing import TYPE_CHECKING, Any
 from urllib.parse import urlparse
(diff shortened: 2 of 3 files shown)


Martchus:
Why do we need to query the individual jobs here? Why are the stats requested via `get_scheduled_product_stats` not good enough? I'm just wondering because when I wrote `get_scheduled_product_stats` I wanted to avoid having to do queries like that avoiding the problem you're fixing here in the first place.

okurz:
> Why do we need to query the individual jobs here? Why are the stats requested via `get_scheduled_product_stats` not good enough? I'm just wondering because when I wrote `get_scheduled_product_stats` I wanted to avoid having to do queries like that avoiding the problem you're fixing here in the first place.

Because of job groups that should be ignored as they are in a "Development" parent job group but don't have the term "Development" in their job group name itself. Maybe meanwhile openQA has sufficient support for that already but you know how it is with testing product increments that only pop up for a very short amount of time so I didn't want to just delete all that code and hope that openQA tests in development job groups would still be sufficiently handled :)

openqabot/incrementapprover.py
```diff
@@ -128,13 +128,12 @@ def fetch_stats(p: dict[str, Any]) -> OpenQAResult:
...
 
-        # Fetch all relevant job details in a single API call to avoid N+1 query problem
         job_ids = [
-            int(ids[0])
+            int(i)
             for stat in stats
             for jobs in stat.values()
             for info in jobs.values()
-            if (ids := info.get("job_ids"))
+            for i in info.get("job_ids", [])
```
asmorodskyi, line 136: can you elaborate what exactly you changing here and why ? from what I see overall logic haven't changed

asmorodskyi (approved):

Martchus (approved):

openqabot/incrementapprover.py
okurz: The logic actually changed here from fetching only the first job ID (`ids[0]`) to fetching all job IDs for the given bucket (`for i in info.get("job_ids", [])`). This is required because we now need metadata for all jobs in `enrich_job_info` so we can properly filter out development jobs from the list, rather than making a decision based solely on the first job in the bucket.

PR #575, opened 2026-06-18
feat: improve link formatting in OBS comments

okurz:
Motivation:
OBS's Markdown engine sanitizes all HTML, causing openQA badge image tags to
render as ugly, raw XML/HTML tags. We need clean, clickable text links instead.

Design Choices:
- Added keyword-only 'is_gitea' flag to comment-generation functions
- Render Markdown badges for Gitea and plain text links for OBS
- Updated commenter and gitea-trigger tests to assert correct output formats

Benefits:
- Highly polished, human-readable test summary comments in OBS request history
- Retains rich visual status badge presentation inside Gitea pull requests

Related issue: https://progress.opensuse.org/issues/197903

### DIFF
--- a/openqabot/commenter.py
+++ b/openqabot/commenter.py
@@ -95,7 +95,7 @@ class Commenter:
         log.debug("Determined comment state for %s: %s", sub, state)
 
         builds = {BuildIdentifier.from_job(j) for j in jobs if "build" in j}
-        msg = self.summarize_message(builds, jobs)
+        msg = self.summarize_message(sub, builds, jobs)
         if not msg:
             raise EmptyCommentError(sub)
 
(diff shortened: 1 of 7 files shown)


openqabot/commenter.py
```diff
@@ -66,15 +66,18 @@ def get_jobs(func: Callable[[int, str | None], list[dict[str, Any]]]) -> list[di
 
         all_jobs = get_jobs(get_submission_results) + get_jobs(get_aggregate_results)
 
-        if res := self.generate_comment(sub, all_jobs):
+        is_gitea = sub.type == "git"
```
asmorodskyi: this should be Submission class function

asmorodskyi (changes requested):
question(blocking) : do we actually have case when commenter processing mixed content ? If commenter always initialized only for Gitea or only for not Gitea it would be better to make is_gitea a class field

openqabot/commenter.py
okurz: done.

okurz:
> question(blocking) : do we actually have case when commenter processing mixed content ? If commenter always initialized only for Gitea or only for not Gitea it would be better to make is_gitea a class field

I think we also discussed this during our offline workshop. Yes, we do have cases where `Commenter` processes mixed content. The code calls `get_submissions()` which fetches all currently active submissions from the QEM dashboard `/api/incidents`. This list is a mixed collection containing both SMELT incidents and Gitea pull requests. `Commenter` then loops over this entire list in a single execution. To support this cleanly, the current design utilizes proper polymorphism and the strategy pattern.

asmorodskyi (approved):

nicksinger (approved):

PR #576, opened 2026-06-18
fix: regression from ec9ecad0, prevent dry run QEM writes

okurz:
Motivation:
During dry runs, the bot skipped posting openQA jobs but still posted
incident settings to the QEM Dashboard, creating a "phantom schedule".
This fixes a regression introduced by commit ec9ecad0 which removed
the early dry-run return from the main loop.

Design Choices:
Added a check for self.dry inside post_qem to log the action and return
early instead of making a PUT request to the dashboard.

Benefits:
Ensures dry runs are completely side-effect free and prevents submissions
from getting stuck in a half-scheduled state without running tests.

Related issue: https://progress.opensuse.org/issues/202824

### DIFF
--- a/openqabot/openqabot.py
+++ b/openqabot/openqabot.py
@@ -52,6 +52,10 @@ class OpenQABot:
             log.warning("Skipping dashboard update: No valid openQA configuration found for data: %s", data)
             return
 
+        if self.dry:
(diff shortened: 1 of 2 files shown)


asmorodskyi (approved):

okurz:
wait, why was this merged now? I don't think I asked mergify to do that.

PR #577, opened 2026-06-18
fix: information on missing jobs + fix: missing submission on CLI arg + enhance "No channels found" message

okurz:
* feat: enhance "No channels found" skip log with details
* fix: handle missing single submissions gracefully
* fix: distinguish missing jobs in approval

### DIFF
--- a/openqabot/approver.py
+++ b/openqabot/approver.py
@@ -8,6 +8,7 @@ import operator
 import re
 import string
 from datetime import UTC, datetime, timedelta
+from enum import Enum, auto
 from functools import lru_cache
 from http import HTTPStatus
 from logging import getLogger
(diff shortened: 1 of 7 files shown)


openqabot/approver.py
```diff
@@ -403,20 +414,20 @@ def get_jobs(self, job_aggr: JobAggr, api: str, sub: int, submission_type: str |
 
     def get_submission_result(
         self, jobs: list[JobAggr], api: str, sub: int, submission_type: str | None = None
-    ) -> bool:
+    ) -> bool | None:
```
asmorodskyi: I have seen this in some other places as well and generally find this very confusing . Imo it is abuse of bool concept . boolean should be True or False . If you need 3 states just switch to int or enum

asmorodskyi (changes requested):

openqabot/approver.py
okurz: alright, enum it is

asmorodskyi (approved):
Nice !

Martchus (approved):

PR #579, opened 2026-06-19
fix: correct a bug in submission priority setup

JRivrain:
- If override_priority as been set to 0 explicitely, delta_prio stays set to 0 instead of the intended
-50, then the job end-up with the default priority of 60 instead of 0.

- Currently if the priority in smelt is between 200 and 219, then 200 // returns 10,
to which we substract 10 then delta_prio returns 0 and the
method returns none, so the job theoritically gets the default Openqa prio of 50. the current
change makes the method return 50 (base_prio) which has the same end
result, but I find this clearer.

### DIFF
--- a/openqabot/types/submissions.py
+++ b/openqabot/types/submissions.py
@@ -198,7 +198,7 @@ class Submissions(BaseConf):
             **({"RRID": sub.rrid} if sub.rrid else {}),
         }
 
-    def get_priority(self, ctx: SubContext) -> int | None:  # noqa: PLR6301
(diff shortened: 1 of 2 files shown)


okurz (changes requested):
Hi, nice! Please add a test to cover this problem and amend your git commit message to follow conventional commits, e.g. `fix: …`

asmorodskyi:
Congrats with first contribution @JRivrain , looks good but yeah pipeline needs to be fixed

okurz (approved):

asmorodskyi (approved):

PR #584, opened 2026-06-24
fix(mergify): prevent endless dependabot PR recreation loops

okurz:
Motivation:
Dependabot's daily updates reset Mergify's 4-day wait timer. This causes
the PRs to close and reopen in an endless loop.

Design Choices:
Remove the 4-day wait rule and use Dependabot's built-in 7-day cooldown
instead, which naturally handles supply chain risks.

Benefits:
Stops the endless PR loops while maintaining secure automated updates.

Similar to https://github.com/openSUSE/qem-dashboard/pull/3362

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -34,25 +34,13 @@ pull_request_rules:
       - and: *base_checks
       - and:
           - author=dependabot[bot]
-          - "label=waited"
     actions:
       queue:
       merge:
         method: squash
-  - name: Wait for some time before validating merge
-    actions:
-      label:
-        add:
-          - waited
-        remove:
-          - waiting
+  - name: continuously rebase pull requests when it's labeled with `rebase`
     conditions:
-      - and:
-          - updated-at<4 days ago
-          - author=dependabot[bot]
-  - name: continuously rebase pull requests when it's labeled with `rebase` or `waited`
-    conditions:
-      - label~=rebase|waited
+      - label=rebase
     actions:
       rebase:
         bot_account: suse-qe-tools-mergify
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,8 @@ updates:
     schedule:
       interval: "weekly"
       day: "monday"
+    cooldown:
+      default-days: 7
     open-pull-requests-limit: 5
     labels:
       - "waiting"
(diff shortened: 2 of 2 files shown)


.mergify.yml
```diff
@@ -33,26 +33,15 @@ pull_request_rules:
     conditions:
       - and: *base_checks
       - and:
+          - base=master
```
asmorodskyi: this is not needed because it already contained in `*base_checks`

asmorodskyi (changes requested):

.mergify.yml
okurz: oops, rebase problem. fixed.

asmorodskyi (approved):

Martchus:
Where is the built-in cooldown of 7 days configured? Is that the default?

asmorodskyi:
> Where is the built-in cooldown of 7 days configured? Is that the default?

I am not sure why Oli stating 7 days , I see 4 https://github.com/openSUSE/qem-bot/blob/master/.mergify.yml#L51 . Idea is that PR is cooling for 4 days with one label (`waiting`) than mergify will replace label to `waited` than in next iteration mergify will merge it . 

At least this is how I got it :thinking:

okurz:
> Where is the built-in cooldown of 7 days configured? Is that the default?

See https://github.com/openSUSE/qem-bot/blob/0783f8f93ffac3524273b10f2cb7f3b5bcee0f20/.github/dependabot.yml#L8

```
    schedule:
      interval: "weekly"
      day: "monday"
```

asmorodskyi:
> > Where is the built-in cooldown of 7 days configured? Is that the default?
> 
> See
> 
> https://github.com/openSUSE/qem-bot/blob/0783f8f93ffac3524273b10f2cb7f3b5bcee0f20/.github/dependabot.yml#L8
> 
> ```
>     schedule:
>       interval: "weekly"
>       day: "monday"
> ```

disagree.  you can not count frequency of dependabot updates , how fast mergify merge PR is only what matters . Imagine that compromised package was release on Monday and dependabot immediately created PR for that .

asmorodskyi (changes requested):
I got confused by commit message and didn't pay attention to actual change .  actually change left us without cooldown period so it is not good idea

okurz:
> I got confused by commit message and didn't pay attention to actual change . actually change left us without cooldown period so it is not good idea

you are right. We are actually missing a "cooldown" setting here

EDIT: DONE

asmorodskyi (approved):

nicksinger (approved):

PR #588, opened 2026-07-03
fix: remove dead code

asmorodskyi:
we stop using source-report comparison following
best industry standards we need to drop it.


okurz (changes requested):
code change looks good so far. Please verify when an actual product increment request is open.

okurz (approved):
verified with `OBS_URL=https://api.suse.de ./qem-bot.py --dry increment-approve --schedule`

kalikiana (approved):

PR #589, opened 2026-07-03
feat: support sequential execution of multiple commands

okurz:
Motivation:
Running each CLI subcommand in separate CI pipelines incurs massive
environment and container image download overhead. Supporting multiple
commands sequentially in a single run drastically reduces setup runtime.

Design Choices:
Configure Typer with chain=True on the main app to support command chaining.
Refactor each subcommand to exit with sys.exit only on non-zero return codes,
allowing successful commands to return naturally and proceed down the chain.

Benefits:
Saves substantial GitLab CI pipeline runtime and container registry bandwidth.
Preserves existing fast-fail error propagation across sequential commands.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,10 +120,8 @@ test-integration: ## Automatically start environment, run integration tests, and
 
 .PHONY: test-integration-payload
 test-integration-payload: env-dashboard-wait  ## Test qem-bot against a local dashboard instance
-	$(QEM_BOT_BASE_CMD) gitea-sync
-	$(QEM_BOT_BASE_CMD) smelt-sync
-	$(QEM_BOT_BASE_CMD) --dry submissions-run
-	$(QEM_BOT_BASE_CMD) --dry sub-approve
+	$(QEM_BOT_BASE_CMD) gitea-sync smelt-sync
+	$(QEM_BOT_BASE_CMD) --dry submissions-run sub-approve
 
 .PHONY: setup-hooks
 setup-hooks: ## Install pre-commit git hooks
--- a/Readme.md
+++ b/Readme.md
@@ -20,7 +20,7 @@ updates information about submissions and related openQA tests.
 
     >>> qem-bot.py --help
 
-    Usage: qem-bot.py [OPTIONS] COMMAND [ARGS]...
+    Usage: qem-bot.py [OPTIONS] COMMAND1 [ARGS]... [COMMAND2 [ARGS]...]...
 
     QEM-Dashboard, SMELT, Gitea and openQA connector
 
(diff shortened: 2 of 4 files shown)


asmorodskyi (changes requested):
Generally I like idea and think it is right direction to move .

BUT :
this looks confusing for the user , command would become super big . also I wonder about handling of input parameters do you also need to repeat them of each command again and again ? 

don't we have `full-run` exactly for that purpose ? I think dedicated command which merge all classes internally and let user to start single command would be much better from many points

okurz:
> Generally I like idea and think it is right direction to move .
> 
> BUT : this looks confusing for the user , command would become super big . also I wonder about handling of input parameters do you also need to repeat them of each command again and again ?
> 
> don't we have `full-run` exactly for that purpose ? I think dedicated command which merge all classes internally and let user to start single command would be much better from many points

So maybe more "Aggregate commands"?

asmorodskyi:
> > Generally I like idea and think it is right direction to move .
> > BUT : this looks confusing for the user , command would become super big . also I wonder about handling of input parameters do you also need to repeat them of each command again and again ?
> > don't we have `full-run` exactly for that purpose ? I think dedicated command which merge all classes internally and let user to start single command would be much better from many points
> 
> So maybe more "Aggregate commands"?

can you elaborate what do you mean under "more aggregate commands" ?

Martchus (approved):

okurz:
> > > Generally I like idea and think it is right direction to move .
> > > BUT : this looks confusing for the user , command would become super big . also I wonder about handling of input parameters do you also need to repeat them of each command again and again ?
> > > don't we have `full-run` exactly for that purpose ? I think dedicated command which merge all classes internally and let user to start single command would be much better from many points
> > 
> > 
> > So maybe more "Aggregate commands"?
> 
> can you elaborate what do you mean under "more aggregate commands" ?

It means that if you don't want to support chaining commands then we need to define combined (or "aggregate" commands) which execute the actions of multiple commands, like the "full-sync" you mentioned but also for other potentially useful combinations, right?

However, using Typer's `chain=True` is actually much better for our use case:
1. Parameter scoping: We don't have to repeat parameters. Global options (like `--token` or `--configs`) are passed *once* before the first command. Only command-specific options are passed to each command. An aggregate command would need a bigger, confusing signature to accept all possible parameters for all the underlying steps.
3. Flexibility: We have many permutations of CI jobs. Hardcoding combination commands (e.g., `sync-and-approve`, `sync-approve-and-comment`) doesn't scale and will quickly pollute the CLI. `chain=True` lets the CI dynamically construct exactly the pipeline it needs.
4. Simplicity: It requires minimal code changes (as seen in this PR) compared to wiring up several new wrapper commands.

asmorodskyi (approved):

okurz:
Closing as newer versions of typer don't support this chaining anymore. I reported an upstream issue https://github.com/fastapi/typer/discussions/1894 about the unclear documentation . Depending on the outcome of that discussion (if any) I would reconsider this approach.

PR #592, opened 2026-07-06
Style improvements

okurz:
* fix(tests): remove unused ty ignore directives
* fix(style): use noun phrases for property docstrings
* fix(tests): make autouse fixtures ruff-compatible

### DIFF
--- a/openqabot/config.py
+++ b/openqabot/config.py
@@ -107,36 +107,36 @@ class Settings(BaseSettings):
 
     @property
     def smelt_graphql(self) -> str:
-        """Return the SMELT GraphQL API URL."""
+        """The SMELT GraphQL API URL."""
         return self.smelt_url + "/graphql"
 
     @property
     def obs_web_url(self) -> str:
-        """Return the OBS web UI URL."""
+        """The OBS web UI URL."""
         return self.obs_url.replace("api.", "build.")
 
     @property
     def download_maintenance(self) -> str:
-        """Return the maintenance download URL."""
+        """The maintenance download URL."""
         if self.download_maintenance_base_url:
             return self.download_maintenance_base_url
         return self.download_base_url + "/SUSE:/Maintenance:/"
 
     @property
     def git_review_bot_user(self) -> str | None:
-        """Return the git review bot username."""
+        """The git review bot username."""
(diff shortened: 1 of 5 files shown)


asmorodskyi (changes requested):
I disagree that we need to support several ruff version . it is getting too complicated and error prone . There are many ways how users may run exactly same version as used in github actions there is pretty straightforward way to just execute github actions locally

Wabri:
> I disagree that we need to support several ruff version . it is getting too complicated and error prone . There are many ways how users may run exactly same version as used in github actions there is pretty straightforward way to just execute github actions locally

Those are mainly style checks, not ruff versions, and all of them are automatically checked during ci and also locally both well defined with a makefile entry.

Wabri (approved):

asmorodskyi:
> > I disagree that we need to support several ruff version . it is getting too complicated and error prone . There are many ways how users may run exactly same version as used in github actions there is pretty straightforward way to just execute github actions locally
> 
> Those are mainly style checks, not ruff versions, and all of them are automatically checked during ci and also locally both well defined with a makefile entry.

How do you plan to remove RUF100 ? it basically instructing ruff to ignore other unneeded ruff ignore instructions so even when we will not need anymore we don't have sane way to notice that we : 
1. don't need RUF076 
2. don't need RUF100

this is deadlock which would stay in code FOREVER

okurz:
> > > I disagree that we need to support several ruff version . it is getting too complicated and error prone . There are many ways how users may run exactly same version as used in github actions there is pretty straightforward way to just execute github actions locally
> > 
> > 
> > Those are mainly style checks, not ruff versions, and all of them are automatically checked during ci and also locally both well defined with a makefile entry.
> 
> How do you plan to remove RUF100 ? it basically instructing ruff to ignore other unneeded ruff ignore instructions so even when we will not need anymore we don't have sane way to notice that we :
> 
>     1. don't need RUF076
> 
>     2. don't need RUF100
> 
> 
> this is deadlock which would stay in code FOREVER

ok, fine. We need to accept running version-pinned ruff+ty. Adapted the code accordingly

okurz:
rebased

asmorodskyi (approved):

PR #593, opened 2026-07-06
perf(approver): parallelize submission filtering and approval

okurz:
Motivation:
Approving multiple submissions is currently a sequential process. Each
submission requires multiple network requests to check openQA results
and aggregate statuses during the filtering phase, followed by further
network requests to OBS or Gitea during the approval phase.

Design Choices:
Introduce `concurrent.futures.ThreadPoolExecutor` to handle both the
`approvable` check and the `approve` call concurrently across all
submissions. Using `strict=True` in `zip` ensures alignment between
input submissions and their evaluation results.

Benefits:
Significantly reduces the time spent in the approval loop when multiple
incidents or pull requests are processed, especially when network
latency is high.

Together with the previous comments this brings the runtime for one
sub-approve cycle down from 135907±5291.28 ms to 20724±384.48 ms which
is a big 84% improvement (speedup 6.6x).

### DIFF
--- a/openqabot/approver.py
+++ b/openqabot/approver.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 import operator
 import re
 import string
+from concurrent.futures import ThreadPoolExecutor
 from datetime import UTC, datetime, timedelta
 from enum import Enum, auto
 from functools import lru_cache
@@ -149,7 +150,10 @@ class Approver:
         )
 
         overall_result = True
-        submissions_to_approve = [sub for sub in subreqs if self.approvable(sub)]
+        with ThreadPoolExecutor() as executor:
+            approvable_flags = list(executor.map(self.approvable, subreqs))
+
+        submissions_to_approve = [sub for sub, ok in zip(subreqs, approvable_flags, strict=True) if ok]
 
         log.info("Submissions to approve:")
         for sub in submissions_to_approve:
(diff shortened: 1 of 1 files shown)


openqabot/approver.py
```diff
@@ -149,16 +150,20 @@ def __call__(self) -> int:
         )
 
         overall_result = True
-        submissions_to_approve = [sub for sub in subreqs if self.approvable(sub)]
+        with ThreadPoolExecutor() as executor:
+            approvable_flags = list(executor.map(self.approvable, subreqs))
```
kalikiana: Should we use [map](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.map) with a **timeout** here (and below)? Otherwise this can run into pipeline timeouts which we should avoid.

okurz: ok but not really related to this PR so I added this as separate commit on top

asmorodskyi (changes requested):
your second commit showing that even on testing phase your already faced with threadpoolexecutor hanging forever . this is the price which I already told you we would pay for the speed ! what we would do when timeout which you set would become not enough and it would start hanging sporadically ?

okurz:
> your second commit showing that even on testing phase your already faced with threadpoolexecutor hanging forever . this is the price which I already told you we would pay for the speed ! what we would do when timeout which you set would become not enough and it would start hanging sporadically ?

Where is this coming from? I did not encounter any "hanging forever". I added timeout as suggested in https://github.com/openSUSE/qem-bot/pull/593#discussion_r3531569078 which I already explained there I don't see strictly related to this PR

asmorodskyi (approved):

openqabot/approver.py
kalikiana: You're introducing the code here, no? 🙃 Thanks!

kalikiana (approved):

PR #594, opened 2026-07-06
feat: add --max-workers CLI argument for parallel processing

okurz:
Motivation:
Different execution environments (local vs. CI/server) have different
concurrency limits and available resources. Providing a way to configure
the maximum number of worker threads allows for fine-tuning performance
and avoiding potential resource exhaustion or API rate limiting.

Design Choices:
Added a global max_workers setting in openqabot/config.py with a
corresponding --max-workers CLI option in openqabot/args.py. Propagated
this setting to all ThreadPoolExecutor instances across the codebase.
If not specified, it defaults to None, letting Python's concurrent.futures
package use its default (typically min(32, os.cpu_count() + 4)).

Benefits:
Users can now limit or increase the concurrency level based on their
needs. This also ensures consistency across all commands that utilize
parallel processing (syncing, triggering, commenting, etc.).

### DIFF
--- a/openqabot/aggrsync.py
+++ b/openqabot/aggrsync.py
@@ -7,6 +7,7 @@ from concurrent.futures import ThreadPoolExecutor, as_completed
 from itertools import chain
 from logging import getLogger
 
+from . import config
 from .loader.config import read_products
 from .loader.qem import get_aggregate_settings_data
 from .syncres import SyncRes
@@ -30,7 +31,7 @@ class AggregateResultsSync(SyncRes):
         update_setting = list(chain.from_iterable(get_aggregate_settings_data(product) for product in self.product))
 
         job_results = {}
-        with ThreadPoolExecutor() as executor:
+        with ThreadPoolExecutor(max_workers=config.settings.max_workers) as executor:
             future_j = {executor.submit(self.client.get_jobs, f): f for f in update_setting}
             for future in as_completed(future_j):
                 job_results[future_j[future]] = future.result()
--- a/openqabot/config.py
+++ b/openqabot/config.py
@@ -51,6 +51,7 @@ class Settings(BaseSettings):
     openqa_instance: str = Field(default="https://openqa.suse.de", alias="OPENQA_INSTANCE")
     singlearch: Path = Field(default=Path("/etc/openqabot/singlearch.yml"), alias="QEM_BOT_SINGLEARCH")
     retry: int = Field(default=2, alias="QEM_BOT_RETRY")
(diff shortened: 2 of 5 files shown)


Readme.md
```diff
@@ -69,6 +69,10 @@ updates information about submissions and related openQA tests.
     │ --retry            -r                   INTEGER  Number of retries           │
     │                                                  [env var: QEM_BOT_RETRY]    │
     │                                                  [default: 2]                │
+    │ --max-workers                           INTEGER  Maximum number of workers   │
+    │                                                  for parallel processing     │
+    │                                                  [env var:                   │
+    │                                                  QEM_BOT_MAX_WORKERS]        │
```
kalikiana:
```suggestion
    │                                                  QEM_BOT_MAX_WORKERS]        │
    │                                                  [default: calculated based on CPU]                │
```

Without that I would assume it's not using multiple workers by default.

asmorodskyi (changes requested):
why do we need to put so many things as input parameters . what if it would be just a field in config which can be overridden via env variable . I think it would be totally fine . Keep in mind that it is not some super popular solution used by huge userbase . it is tiny script executed in single place and have very small userbase . Another problem that currently you put it to place which would expose this parameter everywhere while it is used only in two classes . I found this confusing when I am debugging the code and see so many things which is actually not used anyhow because I can not be sure what is actually used and what is not

okurz:
> why do we need to put so many things as input parameters . what if it would be just a field in config which can be overridden via env variable . I think it would be totally fine . Keep in mind that it is not some super popular solution used by huge userbase . it is tiny script executed in single place and have very small userbase . Another problem that currently you put it to place which would expose this parameter everywhere while it is used only in two classes . I found this confusing when I am debugging the code and see so many things which is actually not used anyhow because I can not be sure what is actually used and what is not

I only did that in response to https://github.com/openSUSE/qem-bot/pull/505#pullrequestreview-4153781225 and didn't initially include it. On the other hand IMHO we should be fine to support many more parameters configurable by CLI and config and environment variable in parallel. The CLI parameter provides a nice help text. By the way, there is even a test `test_cli_envvars_covered_by_settings` that ensures consistency.

Readme.md
okurz: done

Martchus (approved):

okurz:
> why do we need to put so many things as input parameters […]

I now removed the CLI parameter and only kept the possibility to change by config file and env variable for this advanced functionality

asmorodskyi (approved):

PR #597, opened 2026-07-09
feat(config): centralize settings in config.yml

okurz:
Motivation:
Avoid duplication of configuration parameters like obs_url in metadata files.

Design Choices:
Implement update_from_dict on Settings and load config.yml in args.py using
ruamel.yaml. Clean up duplicated Typer defaults in args.py to ensure correct
fallback priority.

Benefits:
Simplifies the metadata repository and enforces clean configuration fallback.

### DIFF
--- a/openqabot/args.py
+++ b/openqabot/args.py
@@ -8,7 +8,7 @@ import logging
 import sys
 from pathlib import Path
 from types import SimpleNamespace
-from typing import Annotated
+from typing import Annotated, Any
 
 import responses
 import typer
@@ -101,6 +101,21 @@ max_detailed_comment_entries_option = Annotated[
 ]
 
 
+def _default(field: str) -> object:
+    """Return a Settings field default so CLI help mirrors config.py."""
+    return config_module.Settings.model_fields[field].get_default(call_default_factory=True)
+
+
+def _apply_cli_overrides(
+    settings: config_module.Settings,
+    overrides: dict[str, Any],
+) -> None:
+    """Override settings with CLI options that were explicitly provided."""
+    for name, value in overrides.items():
+        if value is not None and value is not False:
+            setattr(settings, name, value)
+
+
 def _apply_detailed_comment_options(
     args: SimpleNamespace,
     *,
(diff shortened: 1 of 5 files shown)


Martchus (approved):

asmorodskyi (changes requested):
this is not what I meant . I wanted to **move** settings to config not duplicate them in yet another place to make things even more complicated . commit message speaking about cleanup but I don't see less code but more code

openqabot/args.py
```diff
@@ -184,34 +218,42 @@ def main(  # noqa: PLR0913
...
     ] = None,
     openqa_instance: Annotated[
-        str,
+        str | None,
         typer.Option(
             "-i",
             "--openqa-instance",
             envvar="OPENQA_INSTANCE",
             help="The openQA instance to use\n Other instances than OSD do not update dashboard database",
+            show_default="https://openqa.suse.de",
```
Wabri: Should this be the opensuse instance and not the suse one?

Wabri (changes requested):

openqabot/args.py
Martchus: Unless something has changed, in this project we usually default to OSD and other internal services. Not great but for consistency we should probably not divert here.

Wabri: In `tests/test_args.py` there are multiple reference to o3 though

okurz:
> this is not what I meant . I wanted to **move** settings to config not duplicate them in yet another place to make things even more complicated . commit message speaking about cleanup but I don't see less code but more code

I now reworked to move config handling into `config.py` instead of duplicating it. Changes vs. the previous version:
- Dropped the custom `update_from_dict` and the ruamel loader in `args.py; config.yml` is now loaded via `Settings.load_config_yml()` using pydantic-settings' native `YamlConfigSettingsSource`, so `Settings` stays the single source of truth.
- Deduplicated the Typer defaults: `show_default` now derives from the `config.py` field defaults instead of hardcoded copies.
Net effect of the change is −50 lines. We still need a bit of code to handle that though

And tests fail. Let me take a deeper look later …

asmorodskyi (approved):
fix tests but code looks good to me now :)

okurz:
> fix tests but code looks good to me now :)

I assume a typer update changed how the CLI argument types are rendered hence failing tests so unrelated to my changes. I expect https://github.com/openSUSE/qem-bot/pull/603 to fix this

openqabot/args.py
okurz: @Wabri please review again. I now *removed* the explicit mention of OSD at this place while preserving the actual default from config.py as before

Wabri: Do we want to keep the openqa.suse.de also from the Readme?

okurz: yes

Wabri (approved):

kalikiana:
@Mergifyio queue

okurz:
> @Mergifyio queue

@kalikiana please don't do that. mergify queue skipped some CI checks in the past and also rebasing invalidates our signatures on commits

PR #598, opened 2026-07-13
refactor: move method to approrpiate class

asmorodskyi:
no doubts that function which creates dict based mostly
on data from one particular class should be in this class

### DIFF
--- a/openqabot/loader/gitea.py
+++ b/openqabot/loader/gitea.py
@@ -668,29 +668,6 @@ def _fetch_details(
     )
 
 
-def _init_submission_dict(pr: PullRequest) -> dict[str, Any]:
-    """Initialize a submission dictionary with default values from a PR."""
-    return {
-        "number": pr.number,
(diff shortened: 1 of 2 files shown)


okurz (changes requested):
s/approrpiate/appropriate/

I appreciate your effort to improve the code and I think I understand your motivation here: from a classic Object-Oriented perspective, the heuristic is usually to co-locate a method with the data it primarily operates on. However even within an OOP paradigm, the `PullRequest` class is a pure domain model representing a Git PR entity. The dictionary generated by `to_submission()`, however, introduces pipeline and scheduling-specific concepts (like `emu`, `inReviewQAM`, `embargoed`, and `channels`). A Git PR entity should not possess knowledge of our internal scheduling flags or loader pipeline formats. Binding this method to the data object tightly couples our core data models to the specific needs of the job loader. I would prefer keeping a pure, stateless data objects separate from the functions that transform them.

PR #600, opened 2026-07-13
fix: require tests for all flavors to pass for PR approval

okurz:
Motivation:
Evaluating PRs across multiple test flavors was done individually per
flavor.  This resulted in PRs being approved as soon as tests for any
single flavor passed, even if other flavors failed.

Design Choices:
OpenQA jobs from all flavors are aggregated into a single check,
ensuring approval is only granted if every flavor passes openQA tests.

Related issue: https://progress.opensuse.org/issues/203874

### DIFF
--- a/openqabot/giteatrigger.py
+++ b/openqabot/giteatrigger.py
@@ -196,9 +196,13 @@ class GiteaTrigger:
 
         any_triggered = any(outcome.triggered for outcome in outcomes)
         covered_data = [outcome.data for outcome in outcomes if outcome.data is not None]
+        approve = len(outcomes) == len(applicable_configs)
+
+        if not approve:
+            log.info("Approval blocked for PR %s: unevaluated configs", pullrequest.number)
 
         if self.comment and covered_data and not any_triggered:
-            self.comment_on_pr(pullrequest, covered_data)
+            self.comment_on_pr(pullrequest, covered_data, approve=approve)
 
     def _evaluate_config(self, pullrequest: PullRequest, trigger_config: TriggerConfig) -> EvaluationResult | None:
         """Trigger openQA for one config, or return its coverage Data."""
@@ -252,7 +256,7 @@ class GiteaTrigger:
         # No image_regex configured, use ISO parameters
         return {"ISO_1_URL": f"{repo_url}/{iso_name}", "ISO_1": iso_name}
 
-    def comment_on_pr(self, pullrequest: PullRequest, data_list: list[Data]) -> None:
+    def comment_on_pr(self, pullrequest: PullRequest, data_list: list[Data], *, approve: bool = True) -> None:
         """Comment on PR if openQA results are available."""
         try:
             # build first so an existing job "build" overrides it (setdefault semantics)
@@ -263,7 +267,7 @@ class GiteaTrigger:
 
         if res := self.commenter.generate_comment(pullrequest, all_jobs):
             self.commenter.gitea_comment(pullrequest, *res)
-            if res[1] == "passed":
+            if res[1] == "passed" and approve:
                 if self.dry:
                     log.info("Dry run: Would approve PR %s", pullrequest.number)
                 else:
--- a/openqabot/giteatrigger.py
+++ b/openqabot/giteatrigger.py
@@ -3,6 +3,7 @@
 """Trigger testing for PR(s) with certain label."""
 
 from argparse import Namespace
+from dataclasses import dataclass
 from logging import getLogger
 from typing import Any, cast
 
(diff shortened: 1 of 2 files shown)


openqabot/giteatrigger.py
```diff
@@ -167,30 +178,51 @@ def check_pullrequest(self, pullrequest: PullRequest, trigger_config: TriggerCon
...
-            trigger_config (TriggerConfig): The triggering configuration.
 
         """
-        if self._should_skip_pr(pullrequest, trigger_config):
+        applicable_configs = [
+            tc
+            for tc in self.config_list
+            if tc.project == pullrequest.project and not self._should_skip_pr(pullrequest, tc)
+        ]
+        if not applicable_configs:
```
asmorodskyi, line 212: maybe we want debug log message here ?

openqabot/giteatrigger.py
```diff
@@ -167,30 +178,51 @@ def check_pullrequest(self, pullrequest: PullRequest, trigger_config: TriggerCon
...
             return
 
-        log.info("Evaluating PR %s for openQA triggering with config: %s", pullrequest.number, trigger_config)
+        log.info("Evaluating PR %s for openQA triggering", pullrequest.number)
+
+        outcomes = [res for tc in applicable_configs if (res := self._evaluate_config(pullrequest, tc)) is not None]
```
asmorodskyi, line 218: question(blocking) : not sure how much this is possible in reality but if for some reason one of flavors don't have any tests we will get None here and approve PR without testing one of the flavors

asmorodskyi (changes requested):

openqabot/giteatrigger.py
```diff
@@ -313,7 +344,10 @@ def __call__(self) -> int:
         """
         for trigger_config in self.config_list:
             self.load_prs_for_project(trigger_config.project)
-            for pr in self.prs[trigger_config.project]:
-                self.check_pullrequest(pr, trigger_config)
+        distinct_prs = list(
```
asmorodskyi: I am not sure why we need this . I agree that current version in master is wrong but I think we can avoid whole distinct_prs dict creation by simply go with : 

```
for project in dict.fromkeys(tc.project for tc in self.config_list):
          self.load_prs_for_project(project)
          for pr in self.prs[project]:
              self.check_pullrequest(pr)
```

asmorodskyi: or even better : 

```[python]
for trigger_config in self.config_list:                                                                                                                                                                                                                      
      self.load_prs_for_project(trigger_config.project) 
                                                                                                                                                                                             

for pr in itertools.chain.from_iterable(self.prs.values()):                                                                                                                                                                                                  
      self.check_pullrequest(pr)  
```

asmorodskyi: just ignore version from yesterday but second one should gives same things what you are doing but in more simple way

openqabot/giteatrigger.py
okurz: done

openqabot/giteatrigger.py
okurz: agreed. this could be problematic. Covered. See the updated code

openqabot/giteatrigger.py
okurz: I used your second suggestion now.

okurz:
ready for review again

openqabot/giteatrigger.py
```diff
@@ -167,30 +179,56 @@ def check_pullrequest(self, pullrequest: PullRequest, trigger_config: TriggerCon
...
             return
 
-        log.info("Evaluating PR %s for openQA triggering with config: %s", pullrequest.number, trigger_config)
+        log.info("Evaluating PR %s for openQA triggering", pullrequest.number)
+
+        outcomes = [res for tc in applicable_configs if (res := self._evaluate_config(pullrequest, tc)) is not None]
+
+        any_triggered = any(outcome.triggered for outcome in outcomes)
+        covered_data = [outcome.data for outcome in outcomes if outcome.data is not None]
+        approve = len(outcomes) == len(applicable_configs)
```
asmorodskyi: nitpick : I would add some comment here which clarify what and why we doing here . I mean all three lines above

asmorodskyi (approved):

openqabot/giteatrigger.py
okurz: I have refactored the code and added some doc strings instead so that we shouldn't need the comment. Please re-check if that makes the logic clearer

asmorodskyi: ok , LGTM

Martchus (approved):

PR #608, opened 2026-07-17
fix: skip incident retrigger on transient dashboard API failure

okurz:
Motivation: A dashboard API error while checking already-scheduled
incident jobs was swallowed and treated as "no jobs", so incidents were
rescheduled on transient failures (seen as JSONDecodeError in job
7155840), wasting openQA capacity.

Design Choices: Make the dedup guard fail closed to match the aggregate
path. _get_scheduled_jobs raises a dedicated DashboardError on API
failure and always returns a list otherwise; is_scheduled_job catches it
and treats the incident as already scheduled, skipping it instead of
rescheduling on incomplete data. Signalling failure via an exception
avoids a tri-state None return that callers could misread.

Benefits: Transient dashboard outages no longer cause unnecessary
incident retriggers; submissions and aggregate paths now share the same
fail-closed default with a self-documenting error path.

Related issue: https://gitlab.suse.de/qa-maintenance/bot-ng/-/jobs/7155840

### DIFF
--- a/openqabot/types/submissions.py
+++ b/openqabot/types/submissions.py
@@ -11,6 +11,7 @@ from typing import TYPE_CHECKING, Any, NamedTuple
 import requests
 
 from openqabot.config import OBSOLETE_PARAMS, settings
+from openqabot.errors import DashboardError
 from openqabot.loader import gitea
 from openqabot.pc_helper import apply_pc_tools_image, apply_publiccloud_pint_image
 from openqabot.utils import retry3 as retried_requests
@@ -78,23 +79,31 @@ class Submissions(BaseConf):
 
     @staticmethod
     def _get_scheduled_jobs(sub_id: int, submission_type: str | None = None) -> list[dict[str, Any]]:
-        """Fetch scheduled jobs from the dashboard."""
+        """Fetch scheduled jobs from the dashboard, raising DashboardError on failure."""
         try:
             url = settings.dashboard_url("api", "incident_settings", sub_id)
             params = {"type": submission_type} if submission_type else {}
             res = retried_requests.get(url, headers=settings.dashboard_token_dict, params=params).json()
-            return res if isinstance(res, list) else []
-        except (requests.exceptions.RequestException, json.JSONDecodeError):
+        except (requests.exceptions.RequestException, json.JSONDecodeError) as e:
             log.exception("Dashboard API error: Could not retrieve scheduled jobs for submission %s", sub_id)
-            return []
+            raise DashboardError from e
+        return res if isinstance(res, list) else []
 
     @staticmethod
     def is_scheduled_job(ctx: SubContext, ver: str, submission_type: str | None = None) -> bool:
-        """Check if a job is already scheduled in the dashboard."""
+        """Check if a job is already scheduled in the dashboard.
+
+        If the dashboard API fails, treat the job as already scheduled so a
+        transient error does not trigger an unnecessary reschedule.
+        """
         if not (revs := ctx.sub.revisions_with_fallback(ctx.arch, ver)):
             return False
 
-        jobs = Submissions._get_scheduled_jobs(ctx.sub.id, submission_type)
+        try:
+            jobs = Submissions._get_scheduled_jobs(ctx.sub.id, submission_type)
+        except DashboardError:
+            return True
+
         return any(
             job["flavor"] == ctx.flavor
             and job["arch"] == ctx.arch
--- a/tests/test_submissions_handle.py
+++ b/tests/test_submissions_handle.py
@@ -4,10 +4,12 @@
 
 from __future__ import annotations
 
+import json
 import logging
 from typing import TYPE_CHECKING, Any
 
 import pytest
+import requests
 
 from openqabot.config import DEFAULT_SUBMISSION_TYPE
 from openqabot.errors import NoRepoFoundError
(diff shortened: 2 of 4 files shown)


mdoucha:
Could you extend the test case a bit to make sure that the bot will still correctly schedule any missing jobs in the next run after the dashboard failures? IOW, test that "dashboard error" will not result in incorrect "incident already scheduled" state.

okurz:
> Could you extend the test case a bit to make sure that the bot will still correctly schedule any missing jobs in the next run after the dashboard failures? IOW, test that "dashboard error" will not result in incorrect "incident already scheduled" state.

Because the bot is completely stateless between runs and does not cache the "already scheduled" status locally, a transient dashboard failure cannot permanently taint or block future runs. On the next run, the bot simply makes a fresh API call.
To explicitly safeguard this behavior, I have added a test case (`test_is_scheduled_job_recovers_after_error`) verifying that once the dashboard recovers and returns an empty list, `is_scheduled_job` correctly returns `False`, and the bot will proceed to schedule any missing jobs. Ok?

openqabot/types/submissions.py
```diff
@@ -78,23 +79,31 @@ def repo_osuse(chan: Repos) -> tuple[str, str, str] | tuple[str, str]:
...
             url = settings.dashboard_url("api", "incident_settings", sub_id)
             params = {"type": submission_type} if submission_type else {}
             res = retried_requests.get(url, headers=settings.dashboard_token_dict, params=params).json()
-            return res if isinstance(res, list) else []
-        except (requests.exceptions.RequestException, json.JSONDecodeError):
+        except (requests.exceptions.RequestException, json.JSONDecodeError) as e:
             log.exception("Dashboard API error: Could not retrieve scheduled jobs for submission %s", sub_id)
-            return []
+            raise DashboardError from e
+        return res if isinstance(res, list) else []
```
asmorodskyi, line 90: what else we can get here except list ? are we sure we want just ignore this cases ?

openqabot/types/submissions.py
```diff
@@ -78,23 +79,31 @@ def repo_osuse(chan: Repos) -> tuple[str, str, str] | tuple[str, str]:
...
+        On dashboard API failure assume the job is scheduled (fail closed) to
+        avoid unnecessary retriggers on transient errors.
+        """
         if not (revs := ctx.sub.revisions_with_fallback(ctx.arch, ver)):
             return False
 
-        jobs = Submissions._get_scheduled_jobs(ctx.sub.id, submission_type)
+        try:
+            jobs = Submissions._get_scheduled_jobs(ctx.sub.id, submission_type)
+        except DashboardError:
```
asmorodskyi, line 104: I know you would say that this is "pythonic" and maybe "best industry standard" but I honestly not understand this s better than not throwing exception and just do : 

```
if len(jobs) ==0:
   return True
```

would be happy to hear answer which does not include "best industry standards" and "pythonic"

asmorodskyi (changes requested):

openqabot/types/submissions.py
okurz: Haven't thought about it. This is just moved from former line 86

openqabot/types/submissions.py
okurz: I honestly wrote the code this way anticipating your review comment questioning the I change the return type to `list[dict[str, Any]] | None` :D

asmorodskyi (approved):

tests/test_submissions_handle.py
```diff
@@ -133,14 +135,37 @@ def test_handle_submission_with_ci_url(mocker: MockerFixture) -> None:
...
+def test_is_scheduled_job_non_list_response(mocker: MockerFixture) -> None:
+    sub = MockSubmission(rev_fallback_value=42)
     sub.id = 1
     mocker.patch("openqabot.types.submissions.retried_requests.get").return_value.json.return_value = {"error": "foo"}
     ctx = SubContext(sub, "arch", "flavor", {})
     assert not Submissions.is_scheduled_job(ctx, "ver")
 
 
+@pytest.mark.parametrize("failure", ["json", "request"])
+def test_is_scheduled_job_api_error_fails_closed(mocker: MockerFixture, failure: str) -> None:
```
Martchus: This function name is not easily comprehensible. Maybe it makes sense to add an explanatory comment if it isn't easy to come up with a simple function name for this test. Please be less cryptic than in the other comment you have already added.

tests/test_submissions_handle.py
```diff
@@ -133,14 +135,37 @@ def test_handle_submission_with_ci_url(mocker: MockerFixture) -> None:
...
+    if failure == "json":
+        get.return_value.json.side_effect = json.JSONDecodeError("msg", "doc", 0)
+    else:
+        get.side_effect = requests.exceptions.RequestException("boom")
+    ctx = SubContext(sub, "arch", "flavor", {})
+    assert Submissions.is_scheduled_job(ctx, "ver")
+
+
+def test_is_scheduled_job_recovers_after_error(mocker: MockerFixture) -> None:
+    # A dashboard error fails closed but leaves no state, so once the dashboard
```
Martchus: I wonder what it means for an *error* to fail. I also wonder what it means to fail "closed".

openqabot/types/submissions.py
```diff
@@ -78,23 +79,31 @@ def repo_osuse(chan: Repos) -> tuple[str, str, str] | tuple[str, str]:
...
-            return []
+            raise DashboardError from e
+        return res if isinstance(res, list) else []
 
     @staticmethod
     def is_scheduled_job(ctx: SubContext, ver: str, submission_type: str | None = None) -> bool:
-        """Check if a job is already scheduled in the dashboard."""
+        """Check if a job is already scheduled in the dashboard.
+
+        On dashboard API failure assume the job is scheduled (fail closed) to
```
Martchus: How can "is scheduled" mean or be explained by "fail closed"? What does "fail closed" even mean on its own? It sounds like "assume the job has failed in the state closed" but that doesn't make any sense as "closed" is not a job state.

Martchus (changes requested):

openqabot/types/submissions.py
okurz: Reworded the docstring to plain language: "If the dashboard API fails, treat the job as already scheduled so a transient error does not trigger an unnecessary reschedule." Done.

tests/test_submissions_handle.py
okurz: "fail-closed" means on failure it prevents anything coming through, e.g. like a safety valve or a fire safety door that closes when it looses power. I guess this is a bit too different engineering jargon. Reworded.

tests/test_submissions_handle.py
okurz: Renamed to `test_is_scheduled_job_treats_api_error_as_scheduled` and added a comment.

okurz:
@Martchus please re-review. your suggestions have been attended

Martchus (approved):

PR #612, opened 2026-07-20
build(Makefile): allow overriding the python runner

Wabri:
Add a `PYTHON_RUN` prefix empty by default, this let the target run inside a managed environment like `uv run`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,9 @@ all: help
 # Detect if pytest-xdist is installed for parallel testing
 PYTEST_XDIST := $(shell python3 -c "import xdist" 2>/dev/null && echo "-n auto" || echo "")
 
+# Override to use uv, e.g. PYTHON_RUN="uv run"
+PYTHON_RUN ?=
+
 .PHONY: test-unit
 test-unit: ## Run dynamic tests with coverage report
 	$(UNSHARE) python3 -m pytest $(PYTEST_XDIST) --cov --cov-report=xml --cov-report=term-missing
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -53,14 +56,14 @@ check-code-health: ## Find dead code (vulture)
...
-	ty check
+	$(PYTHON_RUN) ty check
 
 .PHONY: check-types
 check-types: check-types-ty
 
 .PHONY: check-lock
 check-lock: ## Verify uv.lock is in sync with pyproject.toml
-	uv lock --check
+	$(PYTHON_RUN) uv lock --check
```
asmorodskyi: uv run uv lock ?

Makefile
```diff
@@ -101,11 +104,11 @@ test-all-commands-unstable: ## Test all bot commands with fake data
 
 .PHONY: env-dashboard-start
 env-dashboard-start: ## Start a local qem-dashboard and database using podman for integration testing
-	python3 ./scripts/manage_dashboard.py start
+	$(PYTHON_RUN) python3 ./scripts/manage_dashboard.py start
```
asmorodskyi, line 107: uv run python3 ?

asmorodskyi (changes requested):
does not look like it is applicable for all places where you added it

Makefile
Wabri: Yes, you are right, here it's specific for uv. Let me fix it!

Makefile
Wabri: Why?

asmorodskyi: if you would define `PYTHON_RUN=uv run`  than at this line you will get `uv run python3  ./scripts/manage_dashboard.py start` or am I missing something ?

Wabri: Yes and it's correct

asmorodskyi (approved):

kalikiana (approved):

okurz (approved):

PR #614, opened 2026-07-23
feat: fetch Gitea PR priorities using SMELT API v2

okurz:
Motivation:
Gitea PR-based SLE maintenance updates did not show correct priorities on the
blocked dashboard.

Design Choices:
Query the SMELT API v2 experimental updates endpoint using the Gitea update ID
(host:project:pr_number) to retrieve priority and is_emergency values.

Benefits:
The blocked dashboard properly displays accurate priorities for Gitea PR
submissions, allowing proper default sorting and ordering.

Related issue: https://progress.opensuse.org/issues/197348

### DIFF
--- a/openqabot/loader/gitea.py
+++ b/openqabot/loader/gitea.py
@@ -25,6 +25,7 @@ from osc.connection import http_GET
 from osc.core import MultibuildFlavorResolver
 
 from openqabot import config
+from openqabot.loader.smelt import get_gitea_update_data
 from openqabot.types.pullrequest import PullRequest
 from openqabot.utils import retry10 as retried_requests
 
@@ -740,6 +741,8 @@ def _build_submission_record(
     if not submission["packages"]:
         log.info("PR git:%s skipped: No packages found", pr.number)
         return None
+
+    submission["priority"], submission["emu"] = get_gitea_update_data(pr.project, pr.number)
 
     return submission
 
--- a/openqabot/loader/smelt.py
+++ b/openqabot/loader/smelt.py
@@ -4,10 +4,12 @@
 
 from __future__ import annotations
 
+import urllib.parse
 from concurrent import futures
 from logging import getLogger
 from typing import Any, cast
 
+import requests
 from jsonschema import ValidationError, validate
 
 from openqabot import config
@@ -129,6 +131,29 @@ def get_json(query: str, host: str | None = None) -> dict[str, Any]:
     """Fetch JSON data from SMELT using a GraphQL query."""
     host = host or config.settings.smelt_graphql
     return retried_requests.get(host, params={"query": query}, verify=not config.settings.insecure).json()
+
+
+def _gitea_update_id(gitea_host: str, project: str, pr_number: int) -> str:
+    return f"{gitea_host}:{project.replace('/', ':')}:{pr_number}"
+
+
+def get_gitea_update_data(project: str, pr_number: int) -> tuple[int, bool]:
+    """Get priority and emu flag for a Gitea-based submission from SMELT API v2."""
+    gitea_host = urllib.parse.urlparse(config.settings.gitea_url).netloc
+    update_id = _gitea_update_id(gitea_host, project, pr_number)
+    encoded_id = urllib.parse.quote(update_id)
+    url = f"{config.settings.smelt_url}/api/experimental/v2/updates/{encoded_id}"
+
+    try:
+        response = retried_requests.get(url, verify=not config.settings.insecure)
+        response.raise_for_status()
+        res = response.json()
+    except (requests.exceptions.RequestException, ValueError, TypeError) as e:
+        log.warning("Could not get SMELT v2 update data for %s: %s", update_id, e)
+        return 0, False
+
+    data = res.get("data", {})
+    return data.get("priority", 0), data.get("is_emergency", False)
 
 
 def get_active_submission_ids() -> set[int]:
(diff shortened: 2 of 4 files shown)


Martchus (approved):

openqabot/loader/gitea.py
```diff
@@ -722,6 +724,10 @@ def _build_submission_record(
         log.info("PR git:%s skipped: No packages found", pr.number)
         return None
 
+    if not dry:
```
asmorodskyi: why we don't want to do the same not in dry mode ?

uv.lock
```diff
@@ -926,27 +926,27 @@ wheels = [
...
-    { url = "https://files.pythonhosted.org/packages/c3/36/f51d4666d2de6cf33c1f3a1fcc4bb6b70b197dd6ceaa491eef71d78fe8e8/ty-0.0.56-py3-none-win32.whl", hash = "sha256:b30687bb5cd9729d34c889a289edf32770388d9bb05243e534e723fb45e0381b", size = 11093759, upload-time = "2026-07-01T16:44:49.171Z" },
-    { url = "https://files.pythonhosted.org/packages/5e/b4/8fb5d4acfa4afb152245b20fa263069a7547bd1f8e4bfca4eda280c897d7/ty-0.0.56-py3-none-win_amd64.whl", hash = "sha256:ad4c8c47b6f4e3f9ed3fc0b1a5d650088d229e17dd8f63c1826d6bbe94cc3235", size = 12100327, upload-time = "2026-07-01T16:44:51.26Z" },
-    { url = "https://files.pythonhosted.org/packages/b8/fc/6a183e71edde90d0c35c2303f23f7a45b6891d1a2c45daf7b8f869831e19/ty-0.0.56-py3-none-win_arm64.whl", hash = "sha256:57538f273d444a5f1293fa7860e967178afe3917611fc5eff16b64e1204fe0d6", size = 11538780, upload-time = "2026-07-01T16:44:53.8Z" },
+version = "0.0.59"
```
asmorodskyi, line 914: I don't think it should work like that . each commit should not update uv.lock together with code changes . only dependabot should do it normally

openqabot/loader/gitea.py
```diff
@@ -722,6 +724,10 @@ def _build_submission_record(
...
+        gitea_host = urllib.parse.urlparse(config.settings.gitea_url).netloc
```
asmorodskyi: we don't need to pass to function something what function can calculate

openqabot/loader/gitea.py
```diff
@@ -722,6 +724,10 @@ def _build_submission_record(
...
+        submission["priority"], submission["emu"] = get_gitea_update_data(gitea_host, pr.project, pr.number)
```
asmorodskyi: why we not doing this in _init_submission ?

asmorodskyi (dismissed):

uv.lock
okurz: It *is* a separate commit. But I will try to handle any further version discrepancies in separate PRs.

asmorodskyi: ok , **resolved** for this particular PR . but if this will happen with every **PR** now I think this mean we doing something wrong

okurz: agreed.

asmorodskyi: https://github.com/openSUSE/qem-bot/pull/616

asmorodskyi: now you can merge my PR , delete uv.lock related commit and rebase your branch on master

openqabot/loader/gitea.py
okurz: right, dry parameter can be simply passed on. done

openqabot/loader/gitea.py
okurz: done

openqabot/loader/gitea.py
okurz: `_init_submission_dict` is a pure pr-to-dict transform; the SMELT call needs network + fake-data awareness and belongs *after* package validation to avoid wasted calls on skipped PRs. I prefer to keep it here but I collapsed the caller to a single clean line.

okurz:
Addressed all review suggestions

openqabot/loader/smelt.py
```diff
@@ -131,6 +133,31 @@ def get_json(query: str, host: str | None = None) -> dict[str, Any]:
     return retried_requests.get(host, params={"query": query}, verify=not config.settings.insecure).json()
 
 
+def _gitea_update_id(gitea_host: str, project: str, pr_number: int) -> str:
+    return f"{gitea_host}:{project.replace('/', ':')}:{pr_number}"
+
+
+def get_gitea_update_data(project: str, pr_number: int, *, dry: bool) -> tuple[int, bool]:
+    """Get priority and emu flag for a Gitea-based submission from SMELT API v2."""
+    if dry:  # no fixture data available, fall back to defaults
```
asmorodskyi: and why we want to skip **GET** call ? In my understand whole idea behind dry is to avoid creation of mess in prod by inserting broken/fake data when we just debug something or experimenting with code in new branch . but any GET calls are not applicable for that and as more real actions we doing in dry is better for us and more chances to catch problem without doing expensive and risky real calls

okurz: well, the comment should have explained that: `# no fixture data available, fall back to defaults` but looking back (after some days) we can do much better. simplified meaning that we don't need that and also don't need to pass on the dry argument. thank you, fixed

r-richardson (approved):

PR #615, opened 2026-07-23
feat: add native support for retry bot execution

asmorodskyi:
ticket: https://progress.opensuse.org/issues/204009

currently we using external tooling to achieve the same.
Native retry also track if similar exceptions are happening
and quit early when we just re-trying on same error. Inspired by https://github.com/okurz/leaky_bucket_error_count

### DIFF
--- a/openqabot/main.py
+++ b/openqabot/main.py
@@ -2,20 +2,62 @@
 # SPDX-License-Identifier: MIT
 """Entry point for the application."""
 
+import hashlib
 import sys
+import time
+import traceback
+from collections import defaultdict
+from logging import Logger
 
 from dotenv import load_dotenv
 
+import openqabot.config as config_module
+
 from .args import app
 from .utils import create_logger
+
+errorcnt = defaultdict(int)
+error_limit = 10
+
+
+def _handle_exception(e: Exception, log: Logger) -> None:
+    tb = traceback.extract_tb(e.__traceback__)
+    frames = [(frame.filename, frame.name) for frame in tb]
+    signature = f"{type(e).__name__}:" + "->".join(f"{filename}:{func_name}" for filename, func_name in frames)
+    k = hashlib.sha256(signature.encode("utf-8")).hexdigest()
+    errorcnt[k] += 1
+    count = errorcnt[k]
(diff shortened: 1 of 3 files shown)


Martchus:
Is there a related ticket?

asmorodskyi:
this certainly not ready to be merged . I published as POC to open discussion if we want to go this way

openqabot/main.py
```diff
@@ -2,20 +2,57 @@
...
+            handle_exception(ex, log)
+        delay = config_module.settings.app_backoff_factor * attempt
+        log.warning(
+            "attempt %d/%d failed, retrying in %ds",
+            attempt,
+            config_module.settings.app_max_retries,
+            delay,
+        )
+        time.sleep(delay)
+        attempt += 1
```
okurz: I suggest to extract the body. This way also there are less line breaks:

```python
def _run_attempt(attempt: int, log: Logger) -> bool:
    try:
        app()
    except KeyboardInterrupt:
        log.info("Interrupted by user")
        sys.exit(1)
    except Exception as ex:  # noqa: BLE001
        _handle_exception(ex, log)
    else:
        return True

    retries = config_module.settings.app_max_retries
    delay = config_module.settings.app_backoff_factor * attempt
    log.warning("attempt %d/%d failed, retrying in %ds", attempt, retries, delay)
    time.sleep(delay)
    return False
```

and use like

```python
    if _run_attempt(attempt, log):
        break
```

okurz (changes requested):
At the very least it would be nice if you reference the "external tool" https://github.com/okurz/leaky_bucket_error_count :)

asmorodskyi:
> At the very least it would be nice if you reference the "external tool" https://github.com/okurz/leaky_bucket_error_count :)

yeah for sure ! main idea stolen from there ! :)

okurz:
> > At the very least it would be nice if you reference the "external tool" https://github.com/okurz/leaky_bucket_error_count :)
> 
> yeah for sure ! main idea stolen from there ! :)

it's not stolen if the license says "please use it … but attribute"

okurz (approved):

kalikiana (approved):

PR #616, opened 2026-07-24
fix(ci): sync uv.lock with project.toml and prevent async in future

asmorodskyi:
We have added uv.lock to git and even added a check to make sure that it is in sync. But we haven't instruct uv to actually **use** lock when creating venv.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,7 @@
 ---
 version: 2
 updates:
-  - package-ecosystem: "pip"
+  - package-ecosystem: "uv"
     <<: &common
       directory: "/"
       schedule:
--- a/uv.lock
+++ b/uv.lock
@@ -159,41 +159,26 @@ wheels = [
 
 [[package]]
 name = "coverage"
-version = "7.14.1"
+version = "7.15.2"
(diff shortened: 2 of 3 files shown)


uv.lock
okurz, line 1: why the changes in this file in the same commit? Can this be separate?

.github/dependabot.yml
```diff
@@ -1,7 +1,7 @@
 ---
 version: 2
 updates:
-  - package-ecosystem: "pip"
+  - package-ecosystem: "uv"
```
okurz, line 4: Can you please provide a reference for this?

okurz (changes requested):

.github/dependabot.yml
asmorodskyi: what do you mean ? we using uv but fooling dependabot that we using pip . in https://docs.github.com/en/code-security/reference/supply-chain-security/supported-ecosystems-and-repositories you can see that uv is separate entity .

uv.lock
asmorodskyi: done

kalikiana (approved):

okurz (approved):

PR #618, opened 2026-07-24
feat: eliminate architecture ambiguity in IncrementConfig

asmorodskyi:
Motivation:
Previously, `IncrementConfig` tracked target architectures using two separate fields: a singular `arch` string (from the CLI or config) and an `archs` set. This duplication led to contradictory behavior when a CLI override (e.g., `--arch aarch64`) and config definition (e.g., `archs=[x86_64]`) conflicted, as different parts of the codebase read from different fields.

Design Choices:
- Removed the singular `arch` field from `IncrementConfig` in favor of a single `archs` set representing all target architectures.
- Consolidated parsing in `from_config_entry` to automatically fold the singular `arch` config key into `archs`.
- Configured CLI overrides to directly replace `archs` with `{arch}` when a specific architecture is passed via `--arch`, ensuring CLI parameters take precedence.
- Updated unit tests to verify the unified architecture behavior and maintain 100% test coverage.

Benefits:
Unifies architecture filtering under a single source of truth (`archs`), resolving conflicts between CLI arguments and configuration files.

### DIFF
--- a/openqabot/loader/incrementconfig.py
+++ b/openqabot/loader/incrementconfig.py
@@ -43,7 +43,6 @@ class IncrementConfig:
     distri: str
     version: str
     flavor: str
-    arch: str = "any"
     flavor_suffix: str = DEFAULT_FLAVOR_SUFFIX
     project_base: str = ""
     build_project_suffix: str = ""
@@ -125,9 +124,7 @@ class IncrementConfig:
 
     def accepts_build_info(self, build_info: BuildInfo) -> bool:
         """Return True if the build information matches this configuration."""
-        if not all(
-            getattr(self, k) in {"any", getattr(build_info, k)} for k in ("distri", "flavor", "version", "arch")
-        ):
+        if not all(getattr(self, k) in {"any", getattr(build_info, k)} for k in ("distri", "flavor", "version")):
             return False
         if self.archs and build_info.arch not in self.archs:
             return False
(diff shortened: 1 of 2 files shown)


kalikiana (approved):

okurz (approved):

Martchus (changes requested):
This looks like a functional change and not a refactoring. What's your intention here? The commit message body is empty …

okurz (changes requested):
also s/ambiguty/ambiguity/

asmorodskyi:
> This looks like a functional change and not a refactoring. What's your intention here? The commit message body is empty …

currently we have arch defined in two ways within same class . there is "arch" text field which holds arch passed via parameter `--arch` another is `archs` array which reads arches defined in certain config . if you would specify `--arch aarch64` and would have `archs=[x86_64]` some places of our code would read one field and another would read another and do contradicting things . this patch trying to remove this ambiguity

asmorodskyi:
> This looks like a functional change and not a refactoring. What's your intention here? The commit message body is empty …

can you define functional change ? what now would or would not happen which was / was not happening before ?

asmorodskyi:
> also s/ambiguty/ambiguity/

fixed

okurz (approved):

Martchus:
> > This looks like a functional change and not a refactoring. What's your intention here? The commit message body is empty …
> 
> currently we have arch defined in two ways within same class . there is "arch" text field which holds arch passed via parameter `--arch` another is `archs` array which reads arches defined in certain config . if you would specify `--arch aarch64` and would have `archs=[x86_64]` some places of our code would read one field and another would read another and do contradicting things . this patch trying to remove this ambiguity

Ok, so then it is as I suspected: not a refactoring but a functional change that prevents misconfiguration

So change the commit message accordingly (use e.g. `feat: ` and add your explanation to the commit message body).

asmorodskyi:
> > > This looks like a functional change and not a refactoring. What's your intention here? The commit message body is empty …
> > 
> > 
> > currently we have arch defined in two ways within same class . there is "arch" text field which holds arch passed via parameter `--arch` another is `archs` array which reads arches defined in certain config . if you would specify `--arch aarch64` and would have `archs=[x86_64]` some places of our code would read one field and another would read another and do contradicting things . this patch trying to remove this ambiguity
> 
> Ok, so then it is as I suspected: not a refactoring but a functional change that prevents misconfiguration
> 
> So change the commit message accordingly (use e.g. `feat: ` and add your explanation to the commit message body).

DONE

Martchus (approved):

PR #619, opened 2026-07-24
fix: add check to make sure that we not processing several builds

asmorodskyi:
We were relying on set nature uniqueness before, but
seems like it is not enough because sometimes we may
have race condition when old build yet exists in OBS
and new one already uploaded. We should not proceed further
when such state detected. This patch solve this problem

### DIFF
--- a/openqabot/errors.py
+++ b/openqabot/errors.py
@@ -51,6 +51,10 @@ class AmbiguousApprovalStatusError(Error):
     """Raised when several request IDs pointing to the same openQA job."""
 
 
+class AmbiguousBuildsError(Error):
+    """Raised when several builds are available simultenously in IBS."""
+
+
 class PostOpenQAError(Error):
     """Raised when posting a job to openQA fails."""
 
--- a/openqabot/loader/buildinfo.py
+++ b/openqabot/loader/buildinfo.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 from logging import getLogger
 from typing import TYPE_CHECKING, Any
 
+from openqabot.errors import AmbiguousBuildsError
 from openqabot.types.increment import BuildInfo
 from openqabot.utils import get_obs_filter_params
 from openqabot.utils import retry10 as retried_requests
@@ -33,6 +34,7 @@ def load_build_info(
     params = get_obs_filter_params(build_regex)
     log.debug("Checking for '%s' files on %s with params %s", build_regex, url, params)
     rows = retried_requests.get(url, params=params).json().get("data", [])
+    loaded_build_info_set = {}
 
     def get_build_info_from_row(row: dict[str, Any]) -> BuildInfo | None:
         name = row.get("name", "")
@@ -54,4 +56,12 @@ def load_build_info(
 
         return BuildInfo(distri, product, version, flavor, arch, build)
 
-    return {build_info for row in rows if (build_info := get_build_info_from_row(row))}
+    loaded_build_info_set = {build_info for row in rows if (build_info := get_build_info_from_row(row))}
+
+    # we expecting that all BuildInfo objects corresponds to same build
+    # if it is not like that we can not proceed further
+    if len({build_info.build for build_info in loaded_build_info_set}) > 1:
+        error_text = f"{loaded_build_info_set} contains several builds in {url} collected with {config}"
+        raise AmbiguousBuildsError(error_text)
+
+    return loaded_build_info_set
--- a/tests/test_incrementapprover_build_info.py
+++ b/tests/test_incrementapprover_build_info.py
@@ -8,6 +8,7 @@ from pytest_mock import MockerFixture
 from responses import GET
 
 from openqabot.config import BUILD_REGEX
+from openqabot.errors import AmbiguousBuildsError
 from openqabot.loader.buildinfo import load_build_info
 from openqabot.loader.incrementconfig import IncrementConfig
 from openqabot.repodiff import Package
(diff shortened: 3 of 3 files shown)


openqabot/errors.py
```diff
@@ -51,5 +51,9 @@ class AmbiguousApprovalStatusError(Error):
     """Raised when several request IDs pointing to the same openQA job."""
 
 
+class AmbiguousBuildsError(Error):
+    """Raised when several builds are available simultenously in IBS."""
```
okurz, line 55:
```suggestion
    """Raised when several builds are available simultaneously in IBS."""
```

openqabot/loader/buildinfo.py
```diff
@@ -54,4 +56,12 @@ def get_build_info_from_row(row: dict[str, Any]) -> BuildInfo | None:
...
         return BuildInfo(distri, product, version, flavor, arch, build)
 
-    return {build_info for row in rows if (build_info := get_build_info_from_row(row))}
+    loaded_build_info_set = {build_info for row in rows if (build_info := get_build_info_from_row(row))}
+
+    # we expecting that all BuildInfo obects corresponds to same build
+    # if it is not like that we can not proceed further
+    if len({build_info.build for build_info in loaded_build_info_set}) > 1:
+        error_text = f"{loaded_build_info_set} contains several builds in {url} collected with {config}"
+        raise AmbiguousBuildsError(error_text)
```
okurz, line 65: If this is raised then it should be handled explicitly on a higher level and ignore the build or something and still handle other configs or product increments and such

okurz (changes requested):

Wabri (approved):

openqabot/loader/buildinfo.py
asmorodskyi: I don't think it make sense . if we raising this exception this means that currently IBS has two builds and **we don't know** which one should be tested , so all what we can is do another try from scratch . which means that I would expect that this problem will be handled by https://github.com/openSUSE/qem-bot/blob/master/openqabot/main.py#L60

okurz: I don't think the "last resort catch all" in `main()` is what we should rely upon for expected operation. I suggest to *explicitly* handle the exception together with a log message explaining exactly what you stated with
>  this means that currently IBS has two builds and we don't know which one should be tested , so all what we can is do another try from scratch

asmorodskyi: at which level you would expect this handling to happen ? keep in mind that we detected that IBS is in ambiguous state so processing ANYTHING does not make any sense

openqabot/loader/buildinfo.py
```diff
@@ -54,4 +56,12 @@ def get_build_info_from_row(row: dict[str, Any]) -> BuildInfo | None:
 
         return BuildInfo(distri, product, version, flavor, arch, build)
 
-    return {build_info for row in rows if (build_info := get_build_info_from_row(row))}
+    loaded_build_info_set = {build_info for row in rows if (build_info := get_build_info_from_row(row))}
+
+    # we expecting that all BuildInfo objects corresponds to same build
```
Martchus:
```suggestion
    # we expect that all BuildInfo objects correspond to same build
```

openqabot/loader/buildinfo.py
okurz: I don't know anymore, nevermind.

okurz:
please fix "simultenously", rest is fine

openqabot/loader/buildinfo.py
Martchus: I think this part of what @okurz already said is quite important (as the configs are processed independently from each other, unless you've changed that):

> and still handle other configs

So this should be caught at latest so we can proceed with the next config. A good log message wouldn't hurt as well.

> keep in mind that we detected that IBS is in ambiguous state so processing ANYTHING does not make any sense

Maybe the next config does not run into this condition.

Martchus (changes requested):

openqabot/errors.py
asmorodskyi: done

openqabot/loader/buildinfo.py
asmorodskyi: done

asmorodskyi:
> please fix "simultenously", rest is fine

done , also on @Martchus request I changed logic a bit . I personally would like to just fail early but to save time on lengthy discussions around that I just do what was requested

openqabot/loader/buildinfo.py
asmorodskyi: check latest version please

openqabot/loader/buildinfo.py
```diff
@@ -54,4 +54,17 @@ def get_build_info_from_row(row: dict[str, Any]) -> BuildInfo | None:
...
         return BuildInfo(distri, product, version, flavor, arch, build)
 
-    return {build_info for row in rows if (build_info := get_build_info_from_row(row))}
+    loaded_build_info_set = {build_info for row in rows if (build_info := get_build_info_from_row(row))}
+
+    # we expect that all BuildInfo objects correspond to same build
+    # if it is not like that we can not proceed further
+    if len({build_info.build for build_info in loaded_build_info_set}) > 1:
+        log.warning(
+            "%s contains several builds in %s collected with %s. Skipping whole set as ambiguos",
```
okurz:
```suggestion
            "%s contains several builds in %s collected with %s. Skipping whole set as ambiguous",
```

tests/test_incrementapprover_build_info.py
```diff
@@ -146,3 +146,25 @@ def test_extra_builds_package_version_regex_no_match(caplog: pytest.LogCaptureFi
...
+    mocker.patch("openqabot.loader.buildinfo.retried_requests.get").return_value.json.return_value = {
+        "data": [
+            {"name": "SLES-16.0-Online-x86_64-Build1.1.spdx.json"},
+            {"name": "SLES-16.0-Online-x86_64-Build2.2.spdx.json"},
+        ]
+    }
+    res = load_build_info(config, config.build_regex, approver.get_regex_match)
+    assert res == set()
+    assert "contains several builds in" in caplog.text
+    assert "Skipping whole set as ambiguos" in caplog.text
```
okurz:
```suggestion
    assert "Skipping whole set as ambiguous" in caplog.text
```

okurz (changes requested):

openqabot/loader/buildinfo.py
asmorodskyi: done

tests/test_incrementapprover_build_info.py
asmorodskyi: done

okurz (approved):

Martchus (approved):

PR #623, opened 2026-07-26
fix: abort gitea-trigger crawl on 404 instead of retrying

okurz:
Motivation: A 404 for an IBS repo folder means the content is absent and
will not appear within the retry window, yet the crawler retried five times
with backoff (~30s per call), stacking minutes of pointless delay onto every
gitea-trigger PR evaluation.

Design Choices: Drop 404 from the Retry status_forcelist so raise_for_status
raises immediately and the existing handler returns an empty listing. Keep
transient codes (403, 413, 429, 503) that can succeed on retry. Expose the
strategy as an attribute to assert the forcelist in tests.

Benefits: gitea-trigger evaluates PRs without absent folders slowing job
execution, freeing CI runtime.

Related issue: https://progress.opensuse.org/issues/204114

### DIFF
--- a/openqabot/loader/crawler.py
+++ b/openqabot/loader/crawler.py
@@ -31,12 +31,16 @@ class Crawler:
 
         """
         self.verify: bool = verify
-        retry_strategy = Retry(
+        # A 404 for an IBS repo folder means the content is absent and will
+        # not appear within the retry window, so abort immediately instead of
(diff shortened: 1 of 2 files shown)


r-richardson (approved):

kalikiana (approved):

kalikiana:
@okurz I guess you still need to rebase here

PR #631, opened 2026-07-27
perf: optimize repodiff memory usage via streaming

okurz:
Motivation
The `increment-approve` command consumed more than 7.6 GiB of memory during dry
runs on live metadata repositories, causing OOM kills on GitLab runners with
limited memory.

Design Choices
Refactored `repodiff.py` to streaming-parse repository primary XML data via
`etree.iterparse` instead of building the full XML DOM tree in memory. Created
a shared static helper `_extract_package` to avoid duplication between the
stream parser and the element fallback parser, and added unit/integration tests
to maintain 100% statement and branch coverage.

Benefits
Decreased peak Maximum resident set size from ~7.6 GiB down to ~1.4 GiB
(measured via `/usr/bin/time -v`) while ensuring memory is safely cleared
during stream processing and keeping code duplication at zero.

Related issue: https://progress.opensuse.org/issues/204753

### DIFF
--- a/openqabot/repodiff.py
+++ b/openqabot/repodiff.py
@@ -5,6 +5,7 @@
 from __future__ import annotations
 
 import gzip
+import io
 import json
 import re
 import sys
@@ -149,8 +150,8 @@ class RepoDiff:
             raise NoResultsError(error_msg)
         return min(packages, key=lambda p: p.name).name
 
-    def load_repodata(self, url: str) -> etree.Element | None:
-        """Load and parse repository primary metadata for a repository URL."""
+    def load_repodata(self, url: str) -> bytes | etree._Element | None:
+        """Load repository primary metadata for a repository URL."""
         repodata_url = self.make_repodata_url(url)
         repo_data_listing = self.request_and_dump(
             repodata_url,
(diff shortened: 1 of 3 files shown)


openqabot/repodiff.py
```diff
@@ -170,30 +171,72 @@ def load_repodata(self, url: str) -> etree.Element | None:
...
+    def load_packages(self, url: str) -> defaultdict[str, set[Package]]:
+        """Load the list of packages from a repository URL."""
+        repo_data = self.load_repodata(url)
+        if repo_data is None:
+            return defaultdict(set)
+        if isinstance(repo_data, bytes):
+            log.debug("Loading package list for repository %s via stream-parser", url)
+            return self._parse_packages_from_stream(repo_data)
+        if not hasattr(repo_data, "iterfind"):
+            log.error("Could not load repo data for URL %s", url)
```
kalikiana: The error handling feels inconsistent here. The first case is either err: no valid repo or warn: no metadata, second case err: nothing returned by load_repodata.

I guess we check the result when using `load_packages` so nothing is ignored here. Just wanted to mention this since it confused me.

kalikiana (approved):

openqabot/repodiff.py
okurz: Good catch, you're right the error handling was inconsistent. Made the silent None return log a warning like the others, and dropped the dead iterfind branch since load_repodata can only return bytes/element/None anyway.

openqabot/repodiff.py
```diff
@@ -149,8 +150,8 @@ def get_staged_update_name(self, repo_url: str) -> str:
             raise NoResultsError(error_msg)
         return min(packages, key=lambda p: p.name).name
 
-    def load_repodata(self, url: str) -> etree.Element | None:
-        """Load and parse repository primary metadata for a repository URL."""
+    def load_repodata(self, url: str) -> bytes | etree._Element | None:
```
Martchus, line 153: I'm not familiar with this but it seems odd to use a private type here (with leading `_`).

Martchus (approved):

PR #632, opened 2026-07-28
feat: support sequential execution of commands

okurz:
Motivation:
Running each CLI subcommand in separate CI pipelines incurs significant
environment and container image download overhead. Supporting multiple
commands sequentially in a single run drastically reduces setup runtime.

Design Choices:
Configure Typer with chain=True on the main app. Since Typer 0.26+ ignores
chain=True, subclass TyperGroup to custom _ChainedTyperGroup to implement
Click-compatible command chaining and restore backward-compatible help output.

Benefits:
Saves substantial GitLab CI pipeline runtime and container registry bandwidth.
Guarantees robust sequential command chaining compatibility across Typer/Click
versions with 100% statement and branch test coverage.

### DIFF
--- a/openqabot/args.py
+++ b/openqabot/args.py
@@ -12,6 +12,7 @@ from typing import Annotated, Any
 
 import responses
 import typer
+from typer.core import TyperGroup
 
 import openqabot.config as config_module
 
@@ -31,11 +32,71 @@ from .smeltsync import SMELTSync
 from .subsyncres import SubResultsSync
 from .utils import create_logger
 
+
+class _ChainedTyperGroup(TyperGroup):
+    """Support command chaining under newer Click/Typer versions."""
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:  # noqa: ANN401 - dictated by Click's interface
+        kwargs.pop("chain", True)
+        super().__init__(*args, **kwargs)
+        self.chain = True
+
+    def parse_args(self, ctx: typer._click.core.Context, args: list[str]) -> list[str]:
+        if not args and self.no_args_is_help and not ctx.resilient_parsing:
+            raise typer._click.exceptions.NoArgsIsHelpError(ctx)  # noqa: SLF001 - private Click/Typer APIs
+
+        rest = typer._click.core.Command.parse_args(self, ctx, args)  # noqa: SLF001 - private Click/Typer APIs
+
+        ctx._protected_args = rest  # noqa: SLF001 - private Click/Typer APIs
+        ctx.args = []
+
+        return ctx.args
+
+    def collect_usage_pieces(self, ctx: typer._click.core.Context) -> list[str]:
+        rest = typer._click.core.Command.collect_usage_pieces(self, ctx)  # noqa: SLF001 - private Click/Typer APIs
+        rest.append("COMMAND1 [ARGS]... [COMMAND2 [ARGS]...]...")
+        return rest
+
+    def invoke(self, ctx: typer._click.core.Context) -> Any:  # noqa: ANN401 - dictated by Click's interface
+        if not ctx._protected_args:  # noqa: SLF001 - private Click/Typer APIs
+            ctx.fail("Missing command.")
+
+        args = [*ctx._protected_args, *ctx.args]  # noqa: SLF001 - private Click/Typer APIs
+        ctx.args = []
+        ctx._protected_args = []  # noqa: SLF001 - private Click/Typer APIs
+
+        with ctx:
+            ctx.invoked_subcommand = "*" if args else None
+            typer._click.core.Command.invoke(self, ctx)  # noqa: SLF001 - private Click/Typer APIs
+
+            contexts = []
+            while args:
+                cmd_name, cmd, args = self.resolve_command(ctx, args)
+                assert cmd is not None  # noqa: S101 - type narrowing
+                sub_ctx = cmd.make_context(
+                    cmd_name,
+                    args,
+                    parent=ctx,
+                    allow_extra_args=True,
+                    allow_interspersed_args=False,
+                )
+                contexts.append(sub_ctx)
+                args, sub_ctx.args = sub_ctx.args, []
+
+            rv = []
+            for sub_ctx in contexts:
+                with sub_ctx:
+                    rv.append(sub_ctx.command.invoke(sub_ctx))
+            return rv
+
+
 app = typer.Typer(
     name="qem-bot",
     help="QEM-Dashboard, SMELT, Gitea and openQA connector",
     no_args_is_help=True,
     add_completion=False,
+    cls=_ChainedTyperGroup,
+    chain=True,
 )
 log = logging.getLogger("bot")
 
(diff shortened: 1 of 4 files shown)


asmorodskyi (changes requested):
generally I don't mind but if I get it right we are implementing functionality which is missing natively . I would like to have extended discussion what exactly we doing here and why and if such big effort with new code which we would need to support now is really worth it

okurz:
> generally I don't mind but if I get it right we are implementing functionality which is missing natively . I would like to have extended discussion what exactly we doing here and why and if such big effort with new code which we would need to support now is really worth it

yeah, it's a bit unfortunate. I found the command chaining and implemented just to be surprised that meanwhile upstream removed that feature. I looked into alternative ways to implement command chaining (including the upstream recommendations of migrating completely to Click or using shell chaining), but the current implementation remains the most robust and maintainable path for the following reasons:

1. Typer (as of version 0.26.0+) vendored Click but explicitly removed `chain=True` support by inheriting `TyperGroup` directly from `Command` instead of `MultiCommand`. Because of this, standard Typer apps ignore the chaining parameter.
2. Why a full Click rewrite is not viable: One upstream suggestion is to "use Click directly." Doing so would require us to rewrite all options across our 13 subcommands in `openqabot/args.py` into native Click syntax, losing Typer's type-hinting benefits and seamless integration with our settings.
3. Why I see shell chaining (`&&`) as not enough: Running sequential shell commands (e.g. `qem-bot cmd1 && qem-bot cmd2`) forces the Python interpreter, modules, and configurations to re-initialize for every single subcommand. Executing multiple commands within a single Python runtime significantly reduces setup overhead and configuration load times in CI.
4. Mitigating the risk of private APIs: While `_ChainedTyperGroup` relies on private `click` and `typer` parsing internals (like `_protected_args`), we have fully covered this logic with 100% statement+branch coverage in our unit tests. If an upstream update to `typer` or `click` ever breaks these internals, our CI pipeline will catch it immediately during dependency bumps, preventing any production impact.

asmorodskyi:
> 
> yeah, it's a bit unfortunate. I found the command chaining and implemented just to be surprised that meanwhile upstream removed that feature. I looked into alternative ways to implement command chaining (including the upstream recommendations of migrating completely to Click or using shell chaining), but the current implementation remains the most robust and maintainable path for the following reasons:
> 

ok I get it , I fully understand and support your thinking and if we really want to use chaining this would be our best option I agree . But before approving it let me ask last question - initially chaining was sold as "cheap alternative" to simply write single command and avoid chaining at all . To me it does not sound so cheap anymore . Maybe much easier would be just to write some flavor of "full_run" command instead ?

okurz:
> > before approving it let me ask last question - initially chaining was sold as "cheap alternative" to simply write single command and avoid chaining at all . To me it does not sound so cheap anymore

Yes, initially it was a "cheap" implementation. Now it's "cheap" to run, not anymore the implementation

> Maybe much easier would be just to write some flavor of "full_run" command instead ?

Yes, that's an alternative but it wouldn't be "some flavor" but possibly multiple or rather all combinations that can make sense. There is one more alternative, similar to what some options to commands already do, e.g. `increment-approve --schedule`. We could make every suitable command also an option

asmorodskyi:
> Yes, that's an alternative but it wouldn't be "some flavor" but possibly multiple or rather all combinations that can make sense

are you sure that we really need/want to support ALL possible combinations ? My understanding that we actually need ONE extra command which includes full cycle and this would cover 99% of our needs . Or am I wrong can you enum at least 5 combinations which you really like to have ability to start ?

okurz:
> > Yes, that's an alternative but it wouldn't be "some flavor" but possibly multiple or rather all combinations that can make sense
> 
> are you sure that we really need/want to support ALL possible combinations ?

I think this is a leading question and you don't expect an literal answer.

> My understanding that we actually need ONE extra command which includes full cycle and this would cover 99% of our needs

A single "full cycle" command would actually be quite hard to design and maintain, because the subcommands serve completely distinct stages of the lifecycle (syncing metadata, scheduling runs, fetching openQA results, and writing OBS approvals/comments) and different workflows (maintenance submissions, product staging submissions, increments, openSUSE+SUSE) and each stage requires highly specific, non-overlapping CLI options and tokens (e.g. --gitea-token vs. --token).

If we tried to combine them into a single command, we would end up with a mega-command accepting dozens of parameters, and we would have to hardcode a fixed sequence of execution. Command chaining allows us to keep our CLI thin, modular, and flexible.

> Or am I wrong can you enum at least 5 combinations which you really like to have ability to start ?

Why 5 but why not :) Here are 5 useful, real-world combinations that we should support sequentially in a single run to optimize our CI pipelines:

1. gitea-sync smelt-sync
   Synchronizes all external metadata sources (Gitea and SMELT) into the QEM Dashboard sequentially in a single pre-run pipeline -> this is what I would do in production but allow to keep the separate commands for debugging or workaround if one of the two should not be executed
2. submissions-run sub-approve
   First schedules new openQA test runs for maintenance submissions, and then immediately processes OBS approvals for any submissions that have already completed their tests - avoiding the latency of waiting for a separate pipeline run.
3. sub-sync-results aggr-sync-results
   Synchronizes the latest results from openQA to the Dashboard for both standard submissions and aggregate runs in a single execution.
4. sub-sync-results aggr-sync-results sub-approve increment-approve
   Performs a full results-sync for both standard and aggregate jobs, and then sequentially executes approvals/comments for both standard submissions and product increments -> helpful for ensuring full consistency after some pipeline, infrastructure or test issue was fixed
5. gitea-sync submissions-run sub-sync-results sub-approve
   A complete, self-contained Gitea PR pipeline run: imports PR metadata, schedules the corresponding test runs, synchronizes results, and processes approvals/comments in a single container invocation -> this can help in cases like "this gitea PR is super-important, we need to bring it out the door ASAP, an openQA test just passed"

Because of these quite diverse workflows, I think native-like command chaining is the most robust and elegant way to be ready to conduct common useful combinations both during regular production schedules as well as explicit manual recovery work.

Martchus (approved):

asmorodskyi:
I would like to have a live discussion with you before approving this

okurz:
> I would like to have a live discussion with you before approving this

ok, done. We discussed the general approach and you wanted to check the more low-level of the code. please do so and request changes or approve.

openqabot/args.py
```diff
@@ -31,11 +32,71 @@
 from .subsyncres import SubResultsSync
 from .utils import create_logger
 
+
+class _ChainedTyperGroup(TyperGroup):
```
asmorodskyi: **suggestion(non-blocking)** what if we will move this to dedicated file ? as bonus we can noqa whole file and decrease amount of spamming comments

asmorodskyi:
after reading reply on non-blocking suggestion I will approve in any case . overall code looks good to me

openqabot/args.py
okurz: ok, moved to separate file. However, IMHO we should keep targeted and well documented style exclusions where needed and not for whole files.

okurz:
> after reading reply on non-blocking suggestion

the suggestion was partially adressed

asmorodskyi (approved):

PR #633, opened 2026-07-28
fix: abort early on missing or invalid Gitea token

okurz:
Motivation:
Using Gitea commands without a valid token causes requests to hang or retry
repeatedly.

Design Choices:
Added a CLI token validation guard for Gitea-specific commands, created a Gitea
retry session that bypasses 403 status code retries, and added immediate abort
on 401/403 responses.

Benefits:
Ensures early feedback for missing or invalid credentials instead of hanging.

### DIFF
--- a/openqabot/loader/gitea.py
+++ b/openqabot/loader/gitea.py
@@ -6,11 +6,13 @@ from __future__ import annotations
 
 import json
 import re
+import sys
 import urllib.error
 from collections import Counter
 from concurrent import futures
 from dataclasses import dataclass, field
 from functools import lru_cache
+from http import HTTPStatus
 from io import BytesIO
 from logging import getLogger
 from pathlib import Path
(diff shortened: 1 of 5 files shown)


openqabot/loader/gitea.py
```diff
@@ -65,6 +66,13 @@ def make_token_header(token: str) -> dict[str, str]:
     return {} if token is None else {"Authorization": "token " + token}
 
 
+def check_response(response: requests.Response) -> None:
+    """Abort execution if response indicates Gitea authorization failure."""
+    if response.status_code in {401, 403}:
+        log.error("Gitea API error: Missing or invalid Gitea token (status code %d)", response.status_code)
+        sys.exit(1)
```
Wabri: thought: Could it be better to split the status code logs?

Wabri (changes requested):

Martchus (approved):

openqabot/loader/gitea.py
okurz: can do. done

Wabri (approved):

PR #639, opened 2026-08-02
Improve code readability

asmorodskyi:
this PR contains two minor commits : 

- style: clarify name for function
- style: give variables meaningful names

### DIFF
--- a/openqabot/openqa.py
+++ b/openqabot/openqa.py
@@ -55,7 +55,7 @@ class OpenQAInterface:
         """
         return self.url.netloc == config.settings.main_openqa_domain
 
-    def post_job(self, settings: dict[str, Any]) -> None:
+    def post_iso(self, settings: dict[str, Any]) -> None:
         """Post a job to openQA with the given settings."""
         log.info(
             "openqa-cli api --host %s -X post isos %s",
--- a/openqabot/giteatrigger.py
+++ b/openqabot/giteatrigger.py
@@ -237,7 +237,7 @@ class GiteaTrigger:
 
         if self.is_openqa_triggering_needed(matched_iso, trigger_config):
             openqa_settings = self._build_openqa_settings(pullrequest, trigger_config, matched_iso, repo_url, iso_name)
-            self.openqa.post_job(openqa_settings)
+            self.openqa.post_iso(openqa_settings)
             log.info("Triggered openQA tests for PR %s on %s", pullrequest.number, matched_iso.arch)
             return EvaluationResult(triggered=True)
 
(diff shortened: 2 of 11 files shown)


openqabot/openqa.py
```diff
@@ -55,7 +55,7 @@ def __bool__(self) -> bool:
         """
         return self.url.netloc == config.settings.main_openqa_domain
 
-    def post_job(self, settings: dict[str, Any]) -> None:
+    def isos_post(self, settings: dict[str, Any]) -> None:
```
okurz: `post_job` is a verb. This name `isos_post` has no clear meaning in the context of qem-bot and does not read like a proper English term. What's wrong with the old name?

okurz (changes requested):

openqabot/openqa.py
asmorodskyi: > What's wrong with the old name?

it is doing "isos post" api call in openQA . also in openQA there is "job post" so post job looks like related to this api call . 

> This name isos_post has no clear meaning in the context of qem-bot

well in this case `post_job` also has no meaning in qem-bot context . But both has meaning in openQA context and represents two api endpoints one for posting single job another for "posting iso" .

>  post_job is a verb

which means posting single job but the problem that it is not what function is doing

Wabri: Even if there are no good practices on https://peps.python.org/pep-0008/#naming-conventions:
> The naming conventions of Python’s library are a bit of a mess, so we’ll never get this completely consistent

I think the name of a method/function must answer this question "What this block of code do?" and most of the time after the subject there is a verb, so like @okurz suggest I prefer to have a verb at the start of the name.

Martchus: If this is really posting an iso this *should* be renamed.

```suggestion
    def post_iso(self, settings: dict[str, Any]) -> None:
```

Otherwise it looks like the function returns some object that represents an iso post making it not very clear whether the function is actually doing the request on the fly or just returning an object.

kalikiana: I would agree this function clearly calls the `isos` API route... officially we call this [spawning jobs or scheduling products](https://open.qa/docs/#spawning-multiple-jobs-based-on-templates---isos-post).

okurz: how about `schedule_openqa_jobs`?

asmorodskyi: well for me `post_iso` looks better and more "logical" but if I can not convince you on this "schedule_openqa_jobs" is certainly better from what we have now :)

okurz: `post_iso` is also fine for me

okurz (approved):

kalikiana (approved):

PR #645, opened 2026-08-03
refactor(lint): replace name-based ruff selectors with rule codes

asmorodskyi:
Migrate the Ruff configuration and inline suppressions from name-based selectors to standard rule codes.

Selecting rules by name now requires Ruff's experimental preview mode (`preview = true`), which is disabled to ensure linter stability in production.

Changes:
 - Disabled the experimental `preview`    mode in `pyproject.toml`.
 - Converted all name-based linter   selectors and ignores in `pyproject.toml`   to standard rule codes (e.g., `T10`,
   `D203`, `PLC1802`, `D213`, `COM812`).
 - Added `ARG` to `per-file-ignores`   for `tests/*` to handle mock/fixture   signature requirements.
 - Programmatically migrated all inline   `ruff: ignore` comments in Python   files to standard rule codes.
 - Removed preview-only and inactive   rules (such as `PLR0904`, `DOC201`,   `DOC501`, and `pydoclint` selectors)  and cleaned up redundant inline ignores.
 - Reformatted modified files with  `ruff format`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,7 +79,7 @@ find = {}
 [tool.ruff]
 line-length = 120
 exclude = ['metadata']
-preview = true
+preview = false
 
 [tool.ruff.lint]
 # E, F: Standard pycodestyle and Pyflakes (default flake8)
@@ -88,7 +88,7 @@ preview = true
 # PL: Pylint
 extend-select =[
     "E", "F", "D", "Q", "PL",
-    "debugger", # https://docs.astral.sh/ruff/rules/#flake8-debugger-t10
+    "T10", # https://docs.astral.sh/ruff/rules/#flake8-debugger-t10
     "T2", # https://docs.astral.sh/ruff/rules/#flake8-print-t20
     "TD", # https://docs.astral.sh/ruff/rules/#flake8-todos-td
     "W",
@@ -97,7 +97,6 @@ extend-select =[
     "B",
     "SIM",
     "C4",
-    "T10",
     "FIX",
     "ISC",
     "ICN",
@@ -149,20 +148,17 @@ extend-select =[
 ]
 
 ignore = [
-    "incorrect-blank-line-before-class",
-    "len-test",  # use-implicit-booleaness-not-len (name is non-obvious)
-    "multi-line-summary-second-line",
-    "too-many-public-methods",
-    "docstring-missing-returns",
-    "docstring-missing-exception",
-    "missing-trailing-comma",
+    "D203",      # incorrect-blank-line-before-class
+    "PLC1802",   # len-test / use-implicit-booleaness-not-len (name is non-obvious)
+    "D213",      # multi-line-summary-second-line
+    "COM812",    # missing-trailing-comma
 ]
 
 [tool.ruff.lint.mccabe]
 max-complexity = 9
 
 [tool.ruff.lint.per-file-ignores]
-"tests/*" = ["magic-value-comparison", "any-type", "too-many-arguments", "assert", "hardcoded-password-string", "hardcoded-password-func-arg", "too-many-positional-arguments", "undocumented-public-function", "invalid-rule-code"]
+"tests/*" = ["PLR2004", "ANN401", "PLR0913", "S101", "S105", "S106", "PLR0917", "D103", "RUF102", "ARG"]
 
 [tool.ruff.format]
 # Enable reformatting of code snippets in docstrings.
--- a/openqabot/approver.py
+++ b/openqabot/approver.py
@@ -233,7 +233,7 @@ class Approver:
                 e,
             )
 
-    @lru_cache(maxsize=512)  # ruff: ignore[cached-instance-method]
+    @lru_cache(maxsize=512)  # ruff: ignore[B019]
     def is_job_marked_acceptable_for_submission(self, job_id: int, sub: int) -> bool:
         """Check if a job is marked as acceptable for a submission."""
         regex = re.compile(ACCEPTABLE_FOR_TEMPLATE.format(sub=sub), re.DOTALL)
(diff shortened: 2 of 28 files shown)


okurz (changes requested):
I would appreciate if you can reason in your pull requests why the previous decision needs to be revised not just make it sound like anyone before that applied such changes was wrong. Also please investigate which rules this would effect and which we would loose.

asmorodskyi:
> I would appreciate if you can reason in your pull requests why the previous decision needs to be revised not just make it sound like anyone before that applied such changes was wrong. Also please investigate which rules this would effect and which we would loose.

```
Why Is This Still Gated Behind Preview Mode?
Astral (the creators of Ruff) keeps human-readable names behind preview mode for a few reasons:

Stability Guarantees: Once a feature leaves Preview Mode, Astral guarantees it won't break across minor versions. Because rule names or aliases could still be tweaked or renamed before stabilization, they remain in preview to prevent breaking your CI/CD setup if a name changes.

Comment Suppression Alignment: Rule names are also used in inline suppression comments (e.g., # noqa: unused-import). Normalizing this parser across both configuration files and code comments is part of the preview lifecycle.
```

my reason is quite plain simple - **no experimental features in production grade environment** . 

may I ask back what is the reasoning to enable preview mode ? from what I understand it is just "I like to have this fancy feature ! I don't care that developers think that it is unstable " . Or am I missing something ?

pyproject.toml
okurz, line 169: please check if this is still effective when you disable preview. I think I also enabled preview for https://docs.astral.sh/ruff/settings/#lintflake8-copyright but it looks like this is out of preview and in stable now.

asmorodskyi:
```
1. Rules and Categories Affected in Your Config
Below are examples of rules within the prefixes listed in your extend-select that are skipped when preview = false:

RUF (Ruff-specific rules): Newly introduced or experimental RUF checks (e.g., RUF073) will not run until they are stabilized in a future release.

FURB (Refurb): Experimental refactoring rules under FURB are omitted.

DOC (pydoclint): Experimental docstring checks in preview status won't trigger.

E / W (pycodestyle): Preview rules such as E115 (no-indented-block-comment) or E228 (missing-whitespace-around-modulo-operator) will not run.

PL (Pylint): Experimental Pylint rules (such as PLR1712 for swap-with-temporary-variable) are skipped.

2. Extended Preview Behaviors Disabled
In addition to skipping entire rules, setting preview = false disables extended preview behaviors on standard rules you have enabled:

PYI033 (legacy-type-comment): By default, this rule only checks .pyi stub files. With preview enabled, it also checks standard .py files.

SIM401 (if-else-block-instead-of-dict-get): Under preview mode, this rule additionally suggests replacing if-else expressions with dict.get() calls.
```

**IMPORTANT**: additionally to introducing new fully experimental rules , `preview=true`,  *also extend standard check* with more experimental behavior .  For me ( at least ) this is another strong argument why we should NOT keep it in true

pyproject.toml
asmorodskyi: CPY001 (missing-copyright-notice) was stabilized in Ruff version 0.16.0.

okurz:
> ```
> 1. Rules and Categories Affected in Your Config
> Below are examples of rules within the prefixes listed in your extend-select that are skipped when preview = false:
> 
> RUF (Ruff-specific rules): Newly introduced or experimental RUF checks (e.g., RUF073) will not run until they are stabilized in a future release.
> 
> FURB (Refurb): Experimental refactoring rules under FURB are omitted.
> 
> DOC (pydoclint): Experimental docstring checks in preview status won't trigger.
> 
> E / W (pycodestyle): Preview rules such as E115 (no-indented-block-comment) or E228 (missing-whitespace-around-modulo-operator) will not run.
> 
> PL (Pylint): Experimental Pylint rules (such as PLR1712 for swap-with-temporary-variable) are skipped.
> 
> 2. Extended Preview Behaviors Disabled
> In addition to skipping entire rules, setting preview = false disables extended preview behaviors on standard rules you have enabled:
> 
> PYI033 (legacy-type-comment): By default, this rule only checks .pyi stub files. With preview enabled, it also checks standard .py files.
> 
> SIM401 (if-else-block-instead-of-dict-get): Under preview mode, this rule additionally suggests replacing if-else expressions with dict.get() calls.
> ```

I suggest to use https://docs.astral.sh/ruff/preview/#selecting-single-preview-rules to explicitly cover all those which IIUC would effectively freeze our current rule selection to ensure we are not loosing any rule coverage but also prevent any new "experimental" features to need our adaptations before stabilization

asmorodskyi:
> I suggest to use https://docs.astral.sh/ruff/preview/#selecting-single-preview-rules to explicitly cover all those which IIUC would effectively freeze our current rule selection to ensure we are not loosing any rule coverage but also prevent any new "experimental" features to need our adaptations before stabilization

1. what would you do with extension of standard checks with extra experimental functionality ?
2. how you would keep track of what features have left preview state ? this list would soon become garbage

okurz:
> > I suggest to use https://docs.astral.sh/ruff/preview/#selecting-single-preview-rules to explicitly cover all those which IIUC would effectively freeze our current rule selection to ensure we are not loosing any rule coverage but also prevent any new "experimental" features to need our adaptations before stabilization
> 
>     1. what would you do with extension of standard checks with extra experimental functionality ?

I expect those to be disabled by the use of selecting single preview rules

>     2. how you would keep track of what features have left preview state ? this list would soon become garbage

well, if we realize that (maybe ruff will even alert about that) then we simply remove those list entries

asmorodskyi:
I still think that we need to fully disable preview and it is wrong to use it . I don't like half measures with selecting certain features , but seems like I can not convince other people so I give up and close this PR

PR #655, opened 2026-08-04
feat: omit PR-suffix from VERSION and use SUBMISSION_ID for Gitea builds

okurz:
Motivation
Staging builds for Gitea pull requests had the pull request number appended
to the openQA VERSION string, which polluted the product version history.

Design Choices
Remove the :PR-suffix from Gitea-triggered ISO matches' version string.
Always set the SUBMISSION_ID openQA parameter to the Gitea pull request ID.

Benefits
This simplifies product version management in openQA and leverages the
SUBMISSION_ID mechanism to isolate the job history of different pull requests.

Related issue: https://progress.opensuse.org/issues/200952

### DIFF
--- a/openqabot/types/isomatch.py
+++ b/openqabot/types/isomatch.py
@@ -28,6 +28,5 @@ class IsoMatch:
         version = iso_match.group("version")
         arch = iso_match.group("arch")
         build_num = iso_match.group("build")
-        version_str = f"{version}:PR-{pr_number}"
         build_str = f"PR-{pr_number}-{build_num}:{product}-{version}"
-        return cls(product, version_str, build_str, arch)
+        return cls(product, version, build_str, arch)
--- a/openqabot/giteatrigger.py
+++ b/openqabot/giteatrigger.py
@@ -177,6 +177,7 @@ class GiteaTrigger:
             "ARCH": matched_iso.arch,
             "FLAVOR": trigger_config.flavor,
             "_GITEA_PR": str(pullrequest.number),
+            "SUBMISSION_ID": f"git:{pullrequest.number}",
             "GITEA_SHA": pullrequest.commit_sha,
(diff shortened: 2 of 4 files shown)


asmorodskyi (approved):

ricardobranco777 (approved):

openqabot/types/isomatch.py
```diff
@@ -28,6 +28,6 @@ def from_regex_match(cls: type[Self], iso_match: Match[str], pr_number: int) ->
         version = iso_match.group("version")
         arch = iso_match.group("arch")
         build_num = iso_match.group("build")
-        version_str = f"{version}:PR-{pr_number}"
+        version_str = version
```
Wabri: suggestion(blocking): why?

Wabri (changes requested):

openqabot/types/isomatch.py
okurz: That's literally what this PR states. To keep different pull requests' job histories isolated, we now use the `SUBMISSION_ID` parameter (set to `git:<pr_number>`) as the standard openQA mechanism for job isolation, allowing `VERSION` to remain clean, i.e. without a suffix as test owners argued that they still see the version of the product the same as the base version and any asset build including code changes from any pull request on top of the base product should still be seen as the same version, e.g. an ISO file for PR 42 for SLES 16.1 is still SLES 16.1

Wabri: Yes ok, but why not use `version` instead reassign it to another value?

asmorodskyi: nice catch @Wabri  ! @okurz  we can just pass version directly to cls

okurz: yes, of course. Done

Wabri (approved):

PR #658, opened 2026-08-07
fix: Reduce log volume on repodiff for huge package lists

okurz:
Motivation
The repo-diff log output in bot-ng gets huge and unreadable when there
are hundreds or thousands of packages differences in repos, as seen in recent
failures.

Design Choices
We now reduce the logging level from info to debug for this message
to avoid cluttering standard CI logs. Additionally, we enforce a maximum string
length of 200 characters for the package list. If exceeded, it gets truncated
and an ellipsis is appended. We also appended instructions for users on
how to manually run a complete diff comparison.

Benefits
Better readability and lower log clutter in CI pipelines without losing
the ability to examine exact repository differences locally.

### DIFF
--- a/openqabot/repodiff.py
+++ b/openqabot/repodiff.py
@@ -33,6 +33,7 @@ name_tag = ns + "name"
 version_tag = ns + "version"
 arch_tag = ns + "arch"
 primary_re = re.compile(r".*-primary.xml(?:.(gz|zst))?$")
+MAX_PKG_LIST_LENGTH = 200
 
 
 class Package(NamedTuple):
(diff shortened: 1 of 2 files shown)


openqabot/repodiff.py
```diff
@@ -251,7 +252,11 @@ def compute_diff_for_packages(
             diff_by_arch[arch] = diff
             if diff:
                 pkg_list = ", ".join(f"{p.name}-{p.version}-{p.rel}" for p in sorted(diff, key=lambda x: x.name))
-                log.info("Repo diff for arch %s: %d new packages: %s", arch, len(diff), pkg_list)
+                if len(pkg_list) > MAX_PKG_LIST_LENGTH:
```
asmorodskyi, line 255: what if we will keep old way in debug level but display your new shorten message on info ?

asmorodskyi (changes requested):

openqabot/repodiff.py
okurz: Actually also for debug I consider it too much. If you wanna see, open the latest "approve increments" job in the bot-ng pipelines :)

Wabri (approved):

openqabot/repodiff.py
asmorodskyi: ok yeah probably it is too much

asmorodskyi (approved):

PR #663, opened 2026-08-10
perf(giteatrigger): parallelize PR evaluation

okurz:
Motivation:
Evaluating multiple Gitea pull requests sequentially is slow as each PR
requires independent network requests to fetch metadata, check openQA
status, and potentially post comments or jobs.

Design Choices:
Use `concurrent.futures.ThreadPoolExecutor` to process pull requests
concurrently. This is an I/O-bound task where threads are effective
despite the GIL, and it avoids the significant refactoring required for
asyncio.

Benefits:
Reduces total execution time when multiple pull requests need to be
evaluated, especially in environments with high network latency to Gitea
or openQA.

Together with the previous comments this brings the runtime for one
gitea-trigger cycle down from 40280±1654.52 ms to 33785±125.34 ms which
is a 16% improvement (speedup 1.2x).

### DIFF
--- a/openqabot/giteatrigger.py
+++ b/openqabot/giteatrigger.py
@@ -3,6 +3,7 @@
 """Trigger testing for PR(s) with certain label."""
 
 from argparse import Namespace
+from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass
 from itertools import chain
 from logging import getLogger
@@ -362,16 +363,24 @@ class GiteaTrigger:
         self.prs[project] = get_open_prs(self.gitea_token, project, number=self.pr_number)
         log.info("Loaded %d active PRs from %s", len(self.prs[project]), project)
 
+    def _evaluate_pullrequest_safely(self, pullrequest: PullRequest) -> bool:
+        try:
+            self.check_pullrequest(pullrequest)
+        except Exception:
+            log.exception("Unexpected failure while evaluating PR %s", pullrequest.number)
+            return True
+        return False
+
     def __call__(self) -> int:
         """Run test triggering logic.
 
         Returns:
-            int: 0 if all went well
+            int: 0 if every pull request was evaluated, 1 if any evaluation raised
 
         """
         for trigger_config in self.config_list:
             self.load_prs_for_project(trigger_config.project)
-        for pr in chain.from_iterable(self.prs.values()):
-            self.check_pullrequest(pr)
 
-        return 0
+        with ThreadPoolExecutor(max_workers=config.settings.max_workers) as executor:
+            failures = sum(executor.map(self._evaluate_pullrequest_safely, chain.from_iterable(self.prs.values())))
+        return 1 if failures else 0
--- a/tests/test_giteatrigger.py
+++ b/tests/test_giteatrigger.py
@@ -340,6 +340,32 @@ def test_trigger_call_execution(
     mock_check.assert_called_once_with(mock_pr)
 
 
+@pytest.mark.parametrize(
+    ("side_effect", "expected_rc", "expected_logged"),
+    [
+        (None, 0, set()),
+        (RuntimeError("boom"), 1, {1, 2, 3}),
+    ],
+    ids=["all PRs evaluated, rc 0", "every raising PR logged with its number, rc 1"],
+)
+def test_trigger_call_fans_out_and_isolates_failures(
+    trigger: GiteaTrigger,
+    mocker: MockerFixture,
+    side_effect: Exception | None,
+    expected_rc: int,
+    expected_logged: set[int],
+) -> None:
+    """Every PR is evaluated concurrently; a raising PR is logged with its number instead of aborting the run."""
+    mocker.patch.object(trigger, "load_prs_for_project")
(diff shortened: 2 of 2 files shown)


openqabot/giteatrigger.py
```diff
@@ -371,7 +372,13 @@ def __call__(self) -> int:
...
             self.load_prs_for_project(trigger_config.project)
-        for pr in chain.from_iterable(self.prs.values()):
-            self.check_pullrequest(pr)
+
+        with ThreadPoolExecutor(max_workers=config.settings.max_workers) as executor:
+            futures = [executor.submit(self.check_pullrequest, pr) for pr in chain.from_iterable(self.prs.values())]
+            for future in as_completed(futures):
+                try:
+                    future.result()
+                except Exception:
```
asmorodskyi: we hiding ANY issue under 0 exit code

openqabot/giteatrigger.py
```diff
@@ -371,7 +372,13 @@ def __call__(self) -> int:
         """
         for trigger_config in self.config_list:
             self.load_prs_for_project(trigger_config.project)
-        for pr in chain.from_iterable(self.prs.values()):
-            self.check_pullrequest(pr)
+
+        with ThreadPoolExecutor(max_workers=config.settings.max_workers) as executor:
```
asmorodskyi, line 384: **Ctrl-C no longer interrupts the run**

  All futures are submitted up front, and with ThreadPoolExecutor(...) calls shutdown(wait=True) without cancel_futures=True. A KeyboardInterrupt in as_completed propagates, then __exit__
  drains the entire queue before exiting. With a large PR set that's minutes of unstoppable work where the old loop stopped at the next PR.

openqabot/giteatrigger.py
```diff
@@ -371,7 +372,13 @@ def __call__(self) -> int:
         """
         for trigger_config in self.config_list:
             self.load_prs_for_project(trigger_config.project)
-        for pr in chain.from_iterable(self.prs.values()):
-            self.check_pullrequest(pr)
+
```
asmorodskyi, line 383: use of MT is fully cancel any positive effects of lru_cache:

  is_devel_group, get_job_group_info, get_single_job (openqa.py) and get_product_version_from_repo_listing (gitea.py:379) are all cold at start. lru_cache's dict is lock-protected so there's
  no corruption, but N threads will each issue the same uncached request before the first one populates. Wasted requests only — correctness fine.

asmorodskyi (changes requested):

openqabot/giteatrigger.py
okurz: agreed. I updated the logic to track failures safely. The run will now return 1 if any of the PR evaluations raise an unexpected exception, while still ensuring one failing PR doesn't block the evaluation of the rest.

openqabot/giteatrigger.py
okurz: I now use a clean executor.map over a safe wrapper to solve that

openqabot/giteatrigger.py
okurz: I checked all examples you mentioned
- get_single_job – only caller is approver.py:276
- is_devel_group – only via is_in_devel_group, called from syncres.py:71, enrich_job_info (-> incrementapprover) and incrementapprover.py:119. commenter.py never touches it.
- get_product_version_from_repo_listing – chain ends at get_submissions_from_open_prs -> giteasync.py:65, a different subcommand. That path already ran under a ThreadPoolExecutor (loader/gitea.py:830) before this PR.

get_job_group_info is the one that applies, via commenter.get_job_groups_with_failures <- generate_comment <- giteatrigger.py:290. It's keyed by group_id, and get_job_groups_with_failures already dedupes within a call, so the cache only helps across PRs. Worst case is one burst of ≤ max_workers duplicate GET job_groups/{id} per distinct group, first wave only, and only on the commenting path. Everything after hits cache.

A keyed mutex isn't worth the complexity for that; any pre-warming isn't possible since group ids aren't known until jobs are fetched. And the net benefit already includes the "duplication" so still worthwhile

openqabot/giteatrigger.py
```diff
@@ -362,16 +363,24 @@ def load_prs_for_project(self, project: str) -> None:
         self.prs[project] = get_open_prs(self.gitea_token, project, number=self.pr_number)
         log.info("Loaded %d active PRs from %s", len(self.prs[project]), project)
 
+    def _evaluate_pullrequest_safely(self, pullrequest: PullRequest) -> bool:
+        try:
+            self.check_pullrequest(pullrequest)
+        except Exception:
+            log.exception("Unexpected failure while evaluating PR %s", pullrequest.number)
```
asmorodskyi, line 370: maybe it make sense to add some info about what exactly happened ? e.g. exception stack trace and/or error message ?

okurz: That information will be included already. From https://docs.python.org/3/library/logging.html#logging.exception
> Exception info is added to the logging message. This function should only be called from an exception handler.

asmorodskyi (approved):

Wabri (approved):

PR #664, opened 2026-08-10
fix: include OBS revisions and PC images in aggregate repohash

okurz:
Motivation:
PR 607 decoupled the skip logic from the daily counter to save worker
capacity for unchanged aggregates. However, Aggregate was calculating
repohash using only incident IDs (sub.id). Even if an incident
package was rebuilt (causing a new OBS revision) or a dynamic setting
like a new Public Cloud PINT image changed, the repohash remained
identical, incorrectly skipping the job permanently.

Design Choices:
- Call compute_revisions_for_product_repo on all submissions before
  processing so that OBS revisions are fetched.
- Compute repohash using a combination of incident IDs and their OBS
  revisions (sub.id:rev).
- Evaluate apply_public_cloud_settings earlier to include dynamic
  Public Cloud image IDs in the repohash calculation.

Manual verification:
```
QEM_DASHBOARD_URL=https://dashboard.qam.suse.de ./qem-bot.py --dry updates-run
```

Instead of printing "A build with the same RepoHash already exists" and
skipping these products, the bot correctly calculated new repohashes
(e.g., e49b6f81ba6619ee8c3f26bc27cc6424 for SLES15SP6 ppc64le) and
printed "Dry run: Would update QEM Dashboard for api/update_settings". 

Benefits:
Eliminates unnecessary retriggers when nothing actually changes while
correctly scheduling new tests when incident packages are updated or
new external public cloud images are released.

Related issue: https://progress.opensuse.org/issues/205461

### DIFF
--- a/openqabot/types/aggregate.py
+++ b/openqabot/types/aggregate.py
@@ -177,6 +177,7 @@ class Aggregate(BaseConf):
         arch: str,
         data: PostData,
         ci_url: str | None,
+        settings_data: dict[str, Any],
     ) -> dict[str, Any] | None:
         """Create the full post data for the dashboard."""
         openqa_data: dict[str, Any] = {"REPOHASH": data.repohash, "BUILD": data.build}
@@ -187,10 +188,6 @@ class Aggregate(BaseConf):
         }
         if ci_url:
             full_post["openqa"]["__CI_JOB_URL"] = ci_url
-
-        settings_data = apply_public_cloud_settings(self.settings.copy())
-        if settings_data is None:
-            return None
 
         full_post["openqa"].update(settings_data)
         full_post["openqa"]["FLAVOR"] = self.flavor
@@ -214,6 +211,26 @@ class Aggregate(BaseConf):
         self._finalize_post(full_post, arch)
         return full_post
 
+    @staticmethod
+    def _compute_repohash(
+        test_submissions: defaultdict[str, list[Submission]],
+        settings_data: dict[str, Any],
+        issues_arch: str,
+        test_issues: dict[str, ProdVer],
+    ) -> str:
+        hashes = {
+            f"{sub.id}:{sub.revisions_with_fallback(issues_arch, test_issues[issue].version)}"
+            for issue, subs in test_submissions.items()
+            if issue != ALL_ISSUES_KEY
+            for sub in subs
+        } | {
+            str(settings_data[key])
+            for key in ("PUBLIC_CLOUD_IMAGE_ID", "PUBLIC_CLOUD_TOOLS_IMAGE_BASE")
+            if settings_data.get(key)
+        }
+
+        return merge_repohash(sorted(hashes))
+
     def process_arch(
         self,
         arch: str,
@@ -228,9 +245,11 @@ class Aggregate(BaseConf):
 
         test_submissions, test_repos = self.get_test_submissions_and_repos(valid_submissions, issues_arch)
 
-        repohash = merge_repohash(
-            sorted({str(sub.id) for sub in chain.from_iterable(test_submissions.values())}),
-        )
+        settings_data = apply_public_cloud_settings(self.settings.copy())
+        if settings_data is None:
+            return None
+
+        repohash = self._compute_repohash(test_submissions, settings_data, issues_arch, self.test_issues)
 
         try:
             old_jobs = dashboard.get_json(
@@ -273,6 +292,7 @@ class Aggregate(BaseConf):
             arch,
             PostData(test_submissions, test_repos, repohash, build),
             ci_url,
+            settings_data,
         )
 
     def __call__(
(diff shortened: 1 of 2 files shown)


openqabot/types/aggregate.py
```diff
@@ -283,7 +303,8 @@ def __call__(
         ignore_onetime: bool = False,
     ) -> list[dict[str, Any]]:
         """Process all architectures and return a list of posts for the dashboard."""
-        valid_submissions = self.filter_submissions(submissions)
+        active_submissions = [s for s in submissions if s.compute_revisions_for_product_repo(None, None)]
```
asmorodskyi: this seems like unrelated to defined commit message description . can you elaborate why do we need this extra filtering ?

asmorodskyi: also quote from AI (which make sense to me overall but TBH I haven't checked statements line by line): 

```
Failure mode: submissions-run schedules the incident with withAggregate: True, updates-run excludes it from the aggregate, and approver.py:214 then rejects it forever with "Required 
  aggregate tests missing".

  That said, the explicit pass does have real value I'd keep — it normalizes revisions to (None, None) before hashing. Without it, in full-run the lazy path would reuse whatever a preceding
  Submissions worker computed (possibly with limit_archs), making the repohash depend on worker order. So: keep the pass, drop the filtering.

  valid_submissions = self.filter_submissions(submissions)
  for s in valid_submissions:
      s.compute_revisions_for_product_repo(None, None)

  Submissions that fail then contribute id:None to the hash (stable, and revisions_with_fallback hits the params cache without re-fetching) instead of vanishing.
```

tests/test_aggregate.py
```diff
@@ -434,7 +434,7 @@ def test_aggregate_duplicate_submissions(aggregate_factory: Any, submission_mock
     test_repos["REPOS"] = ["repo"]
 
     post_data = PostData(test_submissions, test_repos, "hash", "build")
-    res = agg.create_full_post("x86_64", post_data, None)
+    res = agg.create_full_post("x86_64", post_data, None, {})
```
asmorodskyi, line 438: Both test changes are mechanical signature updates. Nothing asserts that a changed OBS revision or a changed PUBLIC_CLOUD_IMAGE_ID yields a different repohash, or that an unchanged state
  yields the same one — every aggregate test uses MockSubmission with rev_fallback_value=None, so all hashes are id:None. And no test exercises compute_revisions_value=False in the aggregate
  path (the new drop). Line coverage will pass; the regression this commit fixes would not be caught if reverted. A direct _compute_repohash test with two revision values and a PC image ID
  would cover it cheaply.

asmorodskyi (changes requested):

openqabot/types/aggregate.py
okurz: I am not sure what I should do with this AI quote. If you want to use AI maybe consider to make specific code suggestions as GitHub code suggestions or create a change in a separate PR

tests/test_aggregate.py
okurz: Hello Claude :)

asmorodskyi: writing with my own words if you like it more - changes which you introduce requires more tests which would cover such cases like : 
1. changed PUBLIC_CLOUD_IMAGE_ID generate different repohash 
2. there should be test which has rev_fallback_value value different from None 
3. there should be test which has compute_revisions_value=False

openqabot/types/aggregate.py
asmorodskyi: ok just ignore it for now . start from answering question above https://github.com/openSUSE/qem-bot/pull/664#discussion_r3756182258 which I wrote with my own fingers no AI involved :)

okurz: ok, possibly we don't need that extra filtering. I had this included for an initial attempt but partially reverting that seems to work as well.

tests/test_aggregate.py
okurz: ok, extended test cases

asmorodskyi (approved):

Wabri (approved):

openqabot/types/aggregate.py
```diff
@@ -214,6 +211,26 @@ def create_full_post(
...
         return full_post
 
+    @staticmethod
+    def _compute_repohash(
+        test_submissions: defaultdict[str, list[Submission]],
+        settings_data: dict[str, Any],
+        issues_arch: str,
+        test_issues: dict[str, ProdVer],
+    ) -> str:
+        hashes = {
```
mpagot, line 221: I personally find this portion of code not so readable

```
hashes = {
            f"{sub.id}:{sub.revisions_with_fallback(issues_arch, test_issues[issue].version)}"
            for issue, subs in test_submissions.items()
            if issue != ALL_ISSUES_KEY
            for sub in subs
        } | {
            str(settings_data[key])
            for key in ("PUBLIC_CLOUD_IMAGE_ID", "PUBLIC_CLOUD_TOOLS_IMAGE_BASE")
            if settings_data.get(key)
        }
```

asmorodskyi: just saying what looks not good not helping . feel free to open PR with better version or at least offer it here in comment

mpagot (commented):
LGTM

openqabot/types/aggregate.py
mpagot: https://github.com/openSUSE/qem-bot/pull/665

PR #665, opened 2026-08-12
fix: rewrite _compute_repohash

mpagot:
Exactly same behavior, only try to make it more readable.

Improve small part of https://github.com/openSUSE/qem-bot/pull/664

Related issue: https://progress.opensuse.org/issues/205461

### DIFF
--- a/openqabot/types/aggregate.py
+++ b/openqabot/types/aggregate.py
@@ -218,17 +218,20 @@ class Aggregate(BaseConf):
         issues_arch: str,
         test_issues: dict[str, ProdVer],
     ) -> str:
-        hashes = {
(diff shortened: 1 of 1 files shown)


asmorodskyi (approved):

grisu48 (approved):

okurz:
guys, it would have been nice to give me the chance to review as the author of the original code.

> Exactly same behavior, only try to make it more readable.

that's subjective. What I see the change is doing is replacing functional comprehensions with procedural adding to a set which is both less Pythonic and less performant. Maybe we can make the code more readable while still using a proper set comprehension?

Please see https://github.com/openSUSE/qem-bot/pull/667 as my proposal that should keep your idea in mind but expressing the intention of the code in a cleaner way while also saving on lines of code

PR #667, opened 2026-08-12
style: simplify _compute_repohash

okurz:
Motivation: The procedural loops spell out how a set is built instead of
what it contains, which is more code for the same result.

Design Choices: Restore the two set comprehensions and extract the
`_submission_hash` helper so the comprehension body stays a single short
expression. Hoist the public cloud setting keys into a module constant.
This partially reverts commit 6458c42b.

Benefits: Half the lines, no intermediate mutable state, and the long
nested f-string is now a named function readable on its own.

### DIFF
--- a/openqabot/types/aggregate.py
+++ b/openqabot/types/aggregate.py
@@ -31,6 +31,10 @@ ALL_ISSUES_KEY = "TEST_ISSUES[]"
 VALID_AGGREGATE_KEYS = frozenset({"FLAVOR", "archs", "onetime", "packages", "excluded_packages", "test_issues"})
 
 
+def _submission_hash(submission: Submission, issues_arch: str, version: str) -> str:
+    return f"{submission.id}:{submission.revisions_with_fallback(issues_arch, version)}"
+
+
 class PostData(NamedTuple):
     """Data to be posted to dashboard."""
 
@@ -218,21 +222,15 @@ class Aggregate(BaseConf):
         issues_arch: str,
         test_issues: dict[str, ProdVer],
     ) -> str:
-        submission_hashes = set()
-        for issue, subs in test_submissions.items():
-            if issue == ALL_ISSUES_KEY:
-                continue
-            for sub in subs:
-                revision = sub.revisions_with_fallback(issues_arch, test_issues[issue].version)
-                submission_hashes.add(f"{sub.id}:{revision}")
-
-        setting_hashes = set()
-        for key in ("PUBLIC_CLOUD_IMAGE_ID", "PUBLIC_CLOUD_TOOLS_IMAGE_BASE"):
-            if value := settings_data.get(key):
-                setting_hashes.add(str(value))
-
-        hashes = submission_hashes | setting_hashes
-        return merge_repohash(sorted(hashes))
+        submission_hashes: set[str] = {
+            _submission_hash(sub, issues_arch, test_issues[issue].version)
+            for issue, subs in test_submissions.items()
+            if issue != ALL_ISSUES_KEY
+            for sub in subs
+        }
+        setting_keys = ("PUBLIC_CLOUD_IMAGE_ID", "PUBLIC_CLOUD_TOOLS_IMAGE_BASE")
+        setting_hashes: set[str] = {str(settings_data[key]) for key in setting_keys if settings_data.get(key)}
+        return merge_repohash(sorted(submission_hashes | setting_hashes))
 
     def process_arch(
         self,


okurz:
@mpagot @grisu48

asmorodskyi:
I think that current version in master is **more readable**

okurz:
> I think that current version in master is **more readable**

why, because it's using procedural code?

openqabot/types/aggregate.py
```diff
@@ -218,21 +224,14 @@ def _compute_repohash(
...
-                submission_hashes.add(f"{sub.id}:{revision}")
-
-        setting_hashes = set()
-        for key in ("PUBLIC_CLOUD_IMAGE_ID", "PUBLIC_CLOUD_TOOLS_IMAGE_BASE"):
-            if value := settings_data.get(key):
-                setting_hashes.add(str(value))
-
-        hashes = submission_hashes | setting_hashes
-        return merge_repohash(sorted(hashes))
+        submission_hashes = {
```
mpagot: Python uses `{ ... }` both for dict() and set(), this make me personal struggle reading it like this.

I also find the double nested loop not so readable.

Both are highly subjective opinions from a not Python novice.

openqabot/types/aggregate.py
```diff
@@ -30,6 +30,12 @@
 
 VALID_AGGREGATE_KEYS = frozenset({"FLAVOR", "archs", "onetime", "packages", "excluded_packages", "test_issues"})
 
+REPOHASH_SETTING_KEYS = ("PUBLIC_CLOUD_IMAGE_ID", "PUBLIC_CLOUD_TOOLS_IMAGE_BASE")
```
mpagot: What we gain by this, when reading line 233 I need to jump to line 33.

mpagot:
> > I think that current version in master is **more readable**
> 
> why, because it's using procedural code?

You mentioned "which is more code for the same result." ... what is the problem about it?

Is there anything in this PR that aim to improve the qem-bot runtime performances? Is it anything in https://github.com/openSUSE/qem-bot/commit/6458c42bf28c66dbb6a6e9c0358a75926310dbf4 that significantly degrade them? Is there any performance goal?

openqabot/types/aggregate.py
asmorodskyi: +1 to this

asmorodskyi:
> > I think that current version in master is **more readable**
> 
> why, because it's using procedural code?

in your version of code which is using less lines I need to process this code in my head to understand what exactly will happen . Code from Michele I can just **read**

okurz:
> Is there anything in this PR that aim to improve the qem-bot runtime performances? Is it anything in [6458c42](https://github.com/openSUSE/qem-bot/commit/6458c42bf28c66dbb6a6e9c0358a75926310dbf4) that significantly degrade them? Is there any performance goal?

In theory the comprehension is faster but on this level of the code the performance impact is very likely not relevant.

asmorodskyi:
> > Is there anything in this PR that aim to improve the qem-bot runtime performances? Is it anything in [6458c42](https://github.com/openSUSE/qem-bot/commit/6458c42bf28c66dbb6a6e9c0358a75926310dbf4) that significantly degrade them? Is there any performance goal?
> 
> In theory the comprehension is faster but on this level of the code the performance impact is very likely not relevant.

IMHO you are too much concentrated on performance and not realizing that by squeezing every second from the code you paying price with other things like code readability which IMHO for scripts like qem-bot where you have very complicated code flows is much much more important

okurz:
> > > Is there anything in this PR that aim to improve the qem-bot runtime performances? Is it anything in [6458c42](https://github.com/openSUSE/qem-bot/commit/6458c42bf28c66dbb6a6e9c0358a75926310dbf4) that significantly degrade them? Is there any performance goal?
> > 
> > 
> > In theory the comprehension is faster but on this level of the code the performance impact is very likely not relevant.
> 
> IMHO you are too much concentrated on performance and not realizing that by squeezing every second from the code you paying price with other things like code readability which IMHO for scripts like qem-bot where you have very complicated code flows is much much more important

I agree. But it looks like I couldn't make myself clear. I stated above that the performance impact is *not relevant* and my commit message doesn't even mention it. Why are you then stating that I am too much concentrated on performance?

asmorodskyi:
> 
> I agree. But it looks like I couldn't make myself clear. I stated above that the performance impact is _not relevant_ and my commit message doesn't even mention it. Why are you then stating that I am too much concentrated on performance?

my fault , it was wrong

openqabot/types/aggregate.py
ricardobranco777: > Python uses { ... } both for dict() and set(), this make me personal struggle reading it like this.

But this is not a reason to avoid set comprehensions. =/

The previous version was fine because it captured the intention behind one single assignment.  It was a bit unreadable because it had a string with `:` which lets us quickly differentiate between dict & set comprehensions. 

The alternative could have been:

```
        hashes = set() | {
            f"{sub.id}:{sub.revisions_with_fallback(issues_arch, test_issues[issue].version)}"
            for issue, subs in test_submissions.items()
            if issue != ALL_ISSUES_KEY
            for sub in subs
        } | {
            str(settings_data[key])
            for key in ("PUBLIC_CLOUD_IMAGE_ID", "PUBLIC_CLOUD_TOOLS_IMAGE_BASE")
            if settings_data.get(key)
        }
```

Or even better with type-hint:

```
        hashes: set[str] = {
            f"{sub.id}:{sub.revisions_with_fallback(issues_arch, test_issues[issue].version)}"
            for issue, subs in test_submissions.items()
            if issue != ALL_ISSUES_KEY
            for sub in subs
        } | {
            str(settings_data[key])
            for key in ("PUBLIC_CLOUD_IMAGE_ID", "PUBLIC_CLOUD_TOOLS_IMAGE_BASE")
            if settings_data.get(key)
        }
```

ricardobranco777:
```suggestion
        submission_hashes: set[str] = {
```

openqabot/types/aggregate.py
```diff
@@ -218,21 +224,14 @@ def _compute_repohash(
...
-
-        hashes = submission_hashes | setting_hashes
-        return merge_repohash(sorted(hashes))
+        submission_hashes = {
+            _submission_hash(sub, issues_arch, test_issues[issue].version)
+            for issue, subs in test_submissions.items()
+            if issue != ALL_ISSUES_KEY
+            for sub in subs
+        }
+        setting_hashes = {str(settings_data[key]) for key in REPOHASH_SETTING_KEYS if settings_data.get(key)}
```
ricardobranco777:
```suggestion
        setting_hashes: set[str] = {str(settings_data[key]) for key in REPOHASH_SETTING_KEYS if settings_data.get(key)}
```

ricardobranco777 (commented):
I think all of this could have been solved with proper type hints.  Set comprehensions are fine.

openqabot/types/aggregate.py
okurz: ok, I followed your suggestion using type hints.

openqabot/types/aggregate.py
okurz: inlined again

openqabot/types/aggregate.py
okurz: done

openqabot/types/aggregate.py
okurz: done

mpagot (approved):
LGTM

ricardobranco777 (approved):

asmorodskyi (approved):

okurz:
thank you guys :heart:

PR #675, opened 2026-08-19
feat(increment-approve): support cloned jobs

okurz:
Motivation:
Allow the product increment approver of qem-bot to consider additional jobs
cloned via openqa-clone-job (e.g. to handle failures).

Design Choices:
Pass "infer_groups_from_scheduled_product" to the openQA "isos/job_stats"
endpoint so the server resolves cloned jobs into the statistics. Keep it in the
shared constant CLONE_AWARE_STATS_PARAMS, opted into per call site, so the
giteatrigger presence check stays unaffected and tests assert the same value
the bot sends. This widens approval input beyond the originally scheduled job
IDs; clone provenance is still bounded by check_unique_jobid_request_pair and
by development-group filtering on each job's own group.

Benefits:
Cloned and rescheduled product increment tests are automatically considered for
approval without requiring manual intervention.

Related issue: https://progress.opensuse.org/issues/192442

### DIFF
--- a/openqabot/incrementapprover.py
+++ b/openqabot/incrementapprover.py
@@ -21,7 +21,7 @@ import osc.core
 
 from openqabot import config
 from openqabot.config import OBSOLETE_PARAMS
-from openqabot.openqa import ENRICH_KEYS, OpenQAInterface
+from openqabot.openqa import CLONE_AWARE_STATS_PARAMS, ENRICH_KEYS, OpenQAInterface
 from openqabot.pc_helper import apply_public_cloud_settings
 
 from .commenter import Commenter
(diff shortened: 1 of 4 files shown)


Martchus (changes requested):
Let's see how the change on the openQA-side looks like first.

openqabot/incrementapprover.py
```diff
@@ -134,6 +134,7 @@ def fetch_stats(p: dict[str, Any]) -> OpenQAResult:
                 "arch": p["ARCH"],
                 "build": p["BUILD"],
                 "product": p.get("PRODUCT"),
+                **CLONE_AWARE_STATS_PARAMS,
```
asmorodskyi: what about call of `get_scheduled_product_stats` in giteatrigger ? maybe better implement that as boolean flag for function ? or even just hardcode this in function does not look like something should be in caller hands , imagine that in one place of logic we querying openQA with that flag and in another without it and this two calls are part of same flow ....

asmorodskyi (changes requested):

okurz:
maybe helps to do the fix-commit separately first: https://github.com/openSUSE/qem-bot/pull/682

PR #676, opened 2026-08-21
test: show image selection is arbitrary when regex  matches multiple images

mpagot:
[[ NOT TO BE MERGED ]]

qem-bot receives PINT metadata and query parameters as input. It does not control what the user writes in the metadata configuration. It does not control the content or ordering of the PINT server JSON response either.

qem-bot runs a PINT query built from the metadata, filters the response by name regex, region, and state, then selects the most recently published image. When multiple images share the same publish date, only one can be returned.

The current code uses max() over the filtered results. When publish dates are equal, max() retains whichever element it encounters first. The element it encounters first is whichever the PINT server happened to serialize first in its JSON array.

This test demonstrates that swapping the order of two images in the input list changes which image is returned. The selected image is not the best or the expected  match, it is simply the first match.

### DIFF
--- a/tests/test_pc_helper.py
+++ b/tests/test_pc_helper.py
@@ -188,6 +188,30 @@ def test_get_recent_pint_image() -> None:
     assert ret == img3
 
 
+def test_get_recent_pint_image_broad_regex_order_dependent() -> None:
+    """Broad regex matching multiple architectures is order-dependent.
+
+    When publishedon is equal, max() keeps the first element among ties,
+    so the result depends entirely on the PINT API response order.
+    """
+    img_x86 = {
+        "name": "sle-micro-6-1-byos-v20260618-x86-64",
+        "state": "active",
+        "publishedon": "20260618",
+    }
+    img_arm = {
+        "name": "sle-micro-6-1-byos-v20260618-arm64",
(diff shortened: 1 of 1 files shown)


asmorodskyi:
I am not sure why this was created as not mergable PR , I think topic is valid but better to raise it as ticket and close this PR . in a fact suggested test currently is expected behavior so I don't mind to merge as valid case

mpagot:
> I am not sure why this was created as not mergable PR , I think topic is valid but better to raise it as ticket and close this PR . in a fact suggested test currently is expected behavior so I don't mind to merge as valid case

**in a fact suggested test currently is expected behavior** : here the key point: in my opinion it should be not. I do not see any meaningful use-case when a user specify a (too generic) regexp in metadata (like both matching arm and x86 image) and the python code of the bot randomly picks up and trigger a job with one of them. 

I prefer the bot to  loudly fails and let me know it is not able to decide based on my weak (and quite like wrong) regexp in metadata.

asmorodskyi:
> here the key point: in my opinion it should be not

I am not against that . I just said it currently expected behavior but I am open for ideas how can we improve it or make it more safe/stable .

PR #680, opened 2026-09-02
feat(increment-approve): openqa-clone-job clones via --consider-clones

Wabri:
The product increment approver discovered jobs only through openQA's scheduled-product-scoped isos/job_stats. That resolves clone chains created by UI restart/reschedule (the original's clone_id is set), but never sees a standalone job made with openqa-clone-job (e.g. cloning a failed job with CASEDIR to test a fix): the original keeps clone_id=null, so job_stats keeps reporting the stale failure and the clone is invisible.

Add an opt-in --consider-clones flag (default off) that instead discovers jobs via jobs?scope=relevant&latest=1 matching the increment's DISTRI/VERSION/FLAVOR/ARCH/BUILD. scope=relevant drops obsolete jobs and latest=1 makes a clone supersede its original, so standalone clones are finally counted. This needs no openQA-side change and works on current OSD. Kept opt-in so a failing unrelated experiment clone cannot silently block an approval that passes today.

Also fix is_in_devel_group to treat group=None like a missing group instead of crashing on `"Devel" in None` — standalone clones outside any job group have group=None.

Related issue: https://progress.opensuse.org/issues/192442


openqabot/args.py
```diff
@@ -636,6 +636,14 @@ def increment_approve(  # ruff: ignore[too-many-arguments]
             help="Always schedule a new product (even if one already exists)",
         ),
     ] = False,
+    consider_clones: Annotated[
+        bool,
+        typer.Option(
+            "--consider-clones",
```
asmorodskyi: I am not sure why it needs to be an input parameter and why we need to support both states . In my understanding the fact that we not count clones is simply wrong . we just need to **fix** broken behavior

openqabot/incrementapprover.py
```diff
@@ -122,36 +123,53 @@ def _filter_results(self, results: OpenQAResults) -> OpenQAResults:
...
         log.debug("Checking openQA job results for %s", info_str)
 
-        def fetch_stats(p: dict[str, Any]) -> OpenQAResult:
-            return self.client.get_scheduled_product_stats({
-                "distri": p["DISTRI"],
-                "version": p["VERSION"],
-                "flavor": p["FLAVOR"],
-                "arch": p["ARCH"],
-                "build": p["BUILD"],
-                "product": p.get("PRODUCT"),
```
asmorodskyi: if we would just drop this line from here and search ONE time we will get exactly same results

asmorodskyi (changes requested):

asmorodskyi:
just to clarify this one is in conflict with #675 and we should not merge both , right ?

openqabot/openqa.py
```diff
@@ -151,7 +156,7 @@ def is_in_devel_group(self, job: dict[str, Any]) -> bool:
         if config.settings.allow_development_groups:
             return False
 
-        group_name = job.get("group", "")
+        group_name = job.get("group") or ""
```
asmorodskyi: maybe you can just cherry-pick https://github.com/openSUSE/qem-bot/pull/675/changes/55ed71a51651894ca84746220f0a4e8fdef6956a ?  reasons why:

1. change is in dedicated commit and not mixed with other changes 
2. this line has a proper comment explaining why we need this

asmorodskyi: I am pretty sure @okurz  would not mind :)

openqabot/args.py
Martchus: This has not much to do with "clones" which is also ambiguous as it can also refer to restarted jobs.


```suggestion
            "--consider-standalone-jobs",
```

Martchus (changes requested):
I would also change the commit message title to `feat(increment-approve): Allow considering stand-alone jobs`. (The commit message title as currently worded sounds more like an observation of `openqa-clone-job`'s behavior.)

openqabot/args.py
asmorodskyi: the fact that you suggesting better name means that you generally think that this parameter make sense ? I can't imagine why we want not to consider them or better to say I think that having ability to switch on/off this filter may create confusion in some cases

Martchus: Making this configurable seems useful. This way we can test the change but easily go back if it doesn't work out, e.g. due to the side-effect I mentioned before.

Martchus:
Oh, looks like my comment about this having a side-effect was lost. Well, basically, with this we consider jobs from previous scheduled products that are not "shadowed" by jobs of the current scheduled product. So far we really only consider jobs that belong to the most recent matching scheduled product. I'm not sure whether that'll make a difference in practice but the current behavior seems a bit more correct in general.

openqabot/args.py
asmorodskyi: so you would like to create parameter which would stay forever just to test if change is working ??? what if we would just revert commit if it is not working ?

asmorodskyi:
> Oh, looks like my comment about this having a side-effect was lost. Well, basically, with this we consider jobs from previous scheduled products that are not "shadowed" by jobs of the current scheduled product. So far we really only consider jobs that belong to the most recent matching scheduled product. I'm not sure whether that'll make a difference in practice but the current behavior seems a bit more correct in general.

what about `{"infer_groups_from_scheduled_product": "1"}` ? I mean if we just query with this parameter we would remain in borders of scheduled product , right ?

Martchus:
I'm not sure about the semantics of the proposed parameter and what API route it would belong to - so I'm not sure whether you're right :-)

(You just made that parameter up, right? Otherwise, where does it come from?)

asmorodskyi:
> I'm not sure about the semantics of the proposed parameter and what API route it would belong to - so I'm not sure whether you're right :-)
> 
> (You just made that parameter up, right? Otherwise, where does it come from?)

I am not sure maybe @okurz made it up but I took it from #675 which trying to solve same problem in different way . well at least it is my understanding of both PRs but I might be wrong

asmorodskyi:
> Well, basically, with this we consider jobs from previous scheduled products

so @Martchus  you think that "PRODUCT" refers to Scheduling Product id from openQA ? because in a fact it contains string `SLES` and in my understanding should be just dropped because it would most probably just ignored by openQA :thinking: 

CC: @okurz

openqabot/incrementapprover.py
```diff
@@ -122,36 +123,53 @@ def _filter_results(self, results: OpenQAResults) -> OpenQAResults:
...
+
+    def _fetch_stats(self, p: dict[str, Any]) -> OpenQAResult:
+        """Fetch the {state: {result: info}} stats for a single scenario."""
+        query = {
+            "distri": p["DISTRI"],
+            "version": p["VERSION"],
+            "flavor": p["FLAVOR"],
+            "arch": p["ARCH"],
+            "build": p["BUILD"],
+        }
```
okurz: You would need to also account for `SUBMISSION_ID` here

Wabri:
> just to clarify this one is in conflict with #675 and we should not merge both , right ?

Exactly

openqabot/incrementapprover.py
Wabri: What do you mean?

okurz:
maybe helps to do the fix-commit separately first: https://github.com/openSUSE/qem-bot/pull/682

openqabot/openqa.py
okurz: please rebase on top of #682 to have that fix included.

openqabot/incrementapprover.py
okurz: See https://open.qa/docs/#separating-the-history-of-independent-submissions . The scenario you are looking for needs to account for that, otherwise qem-bot would wrongly approve submission A when there is a standalone job for submission B when all other parameters match

openqabot/args.py
```diff
@@ -636,6 +636,14 @@ def increment_approve(  # ruff: ignore[too-many-arguments]
             help="Always schedule a new product (even if one already exists)",
         ),
     ] = False,
+    consider_standalone_jobs: Annotated[
+        bool,
+        typer.Option(
+            "--consider-standalone-jobs",
```
okurz: The new default behaviour should be to account for standalone jobs so if you want to introduce the new behaviour step-wise I suggest to include the enable/disable option pair together, and then afterwards in a separate PR we switch the default.

Wabri: I've updated with both --no-consider and --consider

openqabot/incrementapprover.py
Martchus: Yes, it would make sense to take `SUBMISSION_ID` into account. This would address the concern of taking arbitrary jobs wrongly into account.

I'm not sure whether we can take  `SUBMISSION_ID` into account without a change in openQA because I'm not sure whether the `GET jobs` route actually supports filtering by that parameter.

I'm also not sure whether it would help with the side effect mentioned in https://github.com/openSUSE/qem-bot/pull/680#issuecomment-5527635048. If we re-schedule a scheduled product for a specific submission, the new jobs will have the same `SUBMISSION_ID`. However, we might be ok with this downside.

Martchus: Note that in other cases where we do this kind of lookup we also don't use a `SUBMISSION_ID` but there the build is (at least so far) always different. If that's true for product increments as well, we might not need `SUBMISSION_ID` - although it would nevertheless make sense for correctness.

